PR #298, opened 2015-07-09
Domain mnt lower

8v060htwyc:


eshryane:
this feature change is on hold, pending the outcome of the DB-WG discussion regarding route object authorisation.

PR #310, opened 2015-10-16
Delete with references bug

tpolychniatis:
- fixed delete with references bug
- removed password and cookie as parameters from the @PUT method in the ReferenceService class because they are currently ignored and it is confusing.


dacruz:
:+1:

PR #326, opened 2016-01-20
Fix no cause

tpolychniatis:
fixes the "no-cause" error message.


whois-client/src/main/java/net/ripe/db/whois/api/rest/client/RestClientTarget.java
```diff
@@ -365,21 +366,20 @@ public StreamingRestClient streamingSearch() {
             urlConnection.setReadTimeout(60 * 1000);
 
             return new StreamingRestClient(urlConnection.getInputStream());
+        } catch (ConnectException ce) {
+            LOGGER.info("search failed: {}/search?{}", baseUrl, printParams());
+            LOGGER.error("Caught exception while connecting", ce);
+            throw new RestClientException(503, ce);
         } catch (IOException e) {
-            if (urlConnection == null) {
```
eshryane: urlConnection may indeed be null, check 28b4a368f99f3384046d9c75a12bc1d78adbf5bc

PR #338, opened 2016-02-24
Use local (slave) database for NRTM queries.

eshryane:
NRTM currently uses the master datasource for all queries, as it depends on the serialDao.

This change switches to the local (slave) datasource.


hhazewinkel:
Just wondering, but why is the 'Slave Dao' just in whois-nrtm?
Wouldn't it make more sense to have it in a common place, so others can use that as well?

eshryane:
the Slave Dao is only needed by the NRTM module for querying the local (slave source) database, everywhere else the SerialDao is used for updates (master source). I think it would be confusing to offer two SerialDaos in a common location (even though I did tag the master one as the default with @Primary).

The NRTM client is used to keep a whois database up to date using a remote NRTM server: https://github.com/RIPE-NCC/whois/wiki/Setup-a-RIPE-GRS-mirror-using-Bootstrap-and-NRTM
The NRTM client needs a master datasource to write to. 

The NRTM server sends database updates to clients - it should read from the local (slave) source.

hhazewinkel:
I guess, it is personal preference, but the code change looks good to me.

PR #339, opened 2016-02-24
Batch transfer

MarcGrol:
We moved the "transfer"-function from whois-internal to whois. 
The main reason to do so, is to be able to support multiple concurrent transfer requests (that actually happen a lot in the field).
With the functionality embedded within whois, we can perform multiple searches (in tree and master-database) and updates in a transactional manner. Several concurrent transfers will be processed sequentially, where the next transfer continues with the end result of the previous transfer.

### DIFF
--- a/whois-api/src/test/java/net/ripe/db/whois/api/transfer/inetnum/InetnumOutgoingTransfersServiceIntegrationTest.java
+++ b/whois-api/src/test/java/net/ripe/db/whois/api/transfer/inetnum/InetnumOutgoingTransfersServiceIntegrationTest.java
@@ -406,10 +406,7 @@ public class InetnumOutgoingTransfersServiceIntegrationTest extends AbstractInet
             printErrorMessage(exc.getResponse().readEntity(WhoisResources.class));
             throw exc;
         } catch (WebApplicationException e) {
-            System.out.println("-------------------------");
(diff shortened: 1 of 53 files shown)


eshryane:
Waiting for Jenkins Integration build

whois-api/src/test/java/net/ripe/db/whois/api/transfer/inetnum/InetnumOutgoingTransfersServiceIntegrationTest.java
```diff
@@ -0,0 +1,422 @@
...
+                    .path(URLEncoder.encode(inetnum, "UTF-8"))
+                    .request(MediaType.APPLICATION_JSON_TYPE)
+                    .delete(WhoisResources.class);
+            printErrorMessage(resp);
+        } catch (ClientErrorException exc) {
+            printErrorMessage(exc.getResponse().readEntity(WhoisResources.class));
+            throw exc;
+        } catch (WebApplicationException e) {
+            System.out.println("-------------------------");
+            System.out.println(e.getResponse().readEntity(String.class));
```
dacruz: can we remove the sysout lines?

dacruz:
:+1:

PR #343, opened 2016-03-24
Static immutable lists in validators

hhazewinkel:
Refactor/Implement all actions allowed and types applicable as immutable static lists.
These were mostly dynamic lists that were made with a call to getTypes() or getActions(). With this patch it are returning the static immutable lists.

NOTE: The getTypes() and getActions() are moved to the end of the class as they have no essential functions beyond returning the list.


eshryane:
Looks good! Some Lists are not immutable, but this was already the case, and we can fix separately.

Please wait for the Jenkins integration check to pass before merging.

hhazewinkel:
I will check for it, but I beleive there were only 1 or 2 not immutable as the list was build up from the object template. Maybe we should wrap that.
Although, I had the idea that for instance, the LanguageValidator also was applicable to a few object types instead of all of them. So that could be build up 'dynamiccally' from the object template as well.
But that is more change than was intended.

PR #344, opened 2016-03-29
Validation errors in response

eshryane:
return more meaningful error messages from REST API, when JSON or XML validation fails.


hhazewinkel:
code change and approach look good to me.

PR #345, opened 2016-03-29
Nserver validation

eshryane:
Fix validation of nserver attribute glue record for ENUM (e164.arpa) domains.


hhazewinkel:
Code changes look fine. Just two notes:
- Code formatting in NServerValidator.java does not seem default in IntelliJ
- The SignedMessageHelper was added and deleted

eshryane:
> Code formatting in NServerValidator.java does not seem default in IntelliJ

I've put an opening brace on a new line within the switch statement, to separate the case code. It's non-standard but I think it's easier to read (especially on the second and third cases, I want to separate the brace).

> The SignedMessageHelper was added and deleted

My mistake - I was working on something else, and missed that class during the commit. Feel free to git merge --squash the branch instead, so that commit won't appear on master.

hhazewinkel:
As for the special formatting, I am not against it, but the first coming along applying default formatting, may kill it.

eshryane:
Be careful with automatic formatting, there are other places in the code where we have some "non-standard" formatting - it's hard to get right with an automated tool. If you do reformat, make a separate commit for it (ref. https://github.com/RIPE-NCC/whois/wiki/Coding-standard)

hhazewinkel:
agreed

PR #346, opened 2016-03-30
Correct the syntax description of the person attribute.

hhazewinkel:


whois-client/src/main/java/net/ripe/db/whois/common/rpsl/AttributeSyntax.java
```diff
@@ -857,6 +857,7 @@ public String getDescription(final ObjectType objectType) {
                     "Each word consists of letters, digits or the following symbols:\n" +
                     ".`'_-\n" +
                     "The first word should begin with a letter.\n" +
+                    "At least one other word should also begin with a letter.\n" +
```
dacruz: This fix came up after looking at the Ticket #1208498.

PR #348, opened 2016-03-31
fail gracefully when rpsl has double created attribute

dacruz:


hhazewinkel:
I have it and works similarly, but is a test for multiple last-modified attributes also needed?

dacruz:
The test would be redundant since both attributes are “GENERATED”.

—
Thiago da Cruz

PR #349, opened 2016-04-01
remove timestamps when rpsl has double generated attribute

dacruz:


eshryane:
Looks good to me

hhazewinkel:
From the code it is not directly clear which timestamp is removed.
I can imaging that for last-modified it is the most recent while for created it is the oldest.

Is this something to consider?

dacruz:
Both attributes are removed and and the generated value is added to the RPSL object.
The generated value is defined outside the method. Take a look at the code, I don’t know by heart how the value is defined. 

Cheers

—
Thiago da Cruz

PR #352, opened 2016-04-03
Rpsl module

tpolychniatis:
The object template, attribute syntax/documentation, and the Rpslobject[builder] classes can be very handy. 

Therefore I extracted the rpsl related classes from whois-client to their own module (whois-rpsl) and tried to leave as few dependencies as possible so that the new module can be used as a jar dependency in other projects.


hhazewinkel:
I can see the use of this, but I believe a major part of using a more unique package name for all this rpsl object code should also be applied. Without that better and more unique naming of the package, I kind of disagree with this change.

tpolychniatis:
I totally agree with the renaming of the package, but not in this pull request (PR).
I actually wrote this point, together with other points for improvement, in the email I sent you containing the PR link. The problem with the same package name (net.ripe.db.whois.common) across different modules (whois-client, whois-commons and maybe other places) is already an issue. 
I did not want to solve everything with this pull request, just the problem at hand having as few side effects as possible. Also, after these changes, it will be easier to do the refactoring.

hhazewinkel:
I personally, do not mind for more changes as the package renaming.
Although, I noticed a minor change in the RpslObject.java:
 +            if (rpslAttributes != null && !rpslAttributes.isEmpty()) { 

Not a change I cannot agree with, but is not part of this 'operation'.

tpolychniatis:
I replaced the CollectionUtils.isEmpty() from apache/commons-collections with the line you mentioned above. Just for one usage of the isEmpty method a whole dependency was pulled to the module/project. The main goal for this PR, as my initial comment says, is to create a library with as few dependencies as possible.

I can explain in more detail my choices in person, because writing comments really takes time :)

hhazewinkel:
Looks quit fine, but I see an CI error (related to the lex/yacc usage)

tpolychniatis:
The CI / Integration tests are all passing. I updated the travis script because the path was moved after the refactoring (if this is the CI error you refer to)

whois-rpsl/src/test/java/net/ripe/db/whois/common/rpsl/AttributeSyntaxTest.java
```diff
@@ -143,8 +142,6 @@ public void auth() throws Exception {
 
     @Test
     public void changed() throws Exception {
-        new ChangedAttrFeatureToggle(true);
-
```
hhazewinkel: Could have been done to master first, but I am fine with it.

hhazewinkel:
Looks fine to me, just minor comment (which we can leave as is)

PR #356, opened 2016-04-18
Catch exception if hazelcast instance is not active

eshryane:
Ignore IllegalStateException: (Hazelcast Instance is not active) - do not account for personal queries.

### DIFF
--- a/whois-query/src/main/java/net/ripe/db/whois/query/pipeline/ExceptionHandler.java
+++ b/whois-query/src/main/java/net/ripe/db/whois/query/pipeline/ExceptionHandler.java
@@ -10,6 +10,7 @@ import org.jboss.netty.handler.codec.frame.TooLongFrameException;
 import org.jboss.netty.handler.timeout.TimeoutException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.dao.DataAccessException;
 
 import java.io.IOException;
 import java.nio.channels.ClosedChannelException;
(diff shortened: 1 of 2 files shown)


hhazewinkel:
In the HazelcastPersonalObjectAccounting object, I am kind of wondering whether we should not catch any kind of exception instead of a few specific ones. Also the methods could be written like:
    public int getQueriedPersonalObjects(final InetAddress remoteAddress) {
        try {
            return counterMap.get(remoteAddress);
        } catch (OperationTimeoutException e) {
            // prevents user from seeing "internal server error"
        } catch (IllegalStateException e) {
            // hazelcast instance is not active
         } catch (Exception e) {
            // anything???
        }
        return 0;
    }

eshryane:
We could catch multiple exception types in the same block:

catch (OperationTimeoutException | IllegalStateException e)

but it's not clear why we are catching these exceptions. We currently have different catch blocks for different reasons. We can separately catch and log (Runtime)Exception too.

Also: the map can return null, we need to do a null check or we may get a NullPointerException.

whois-query/src/main/java/net/ripe/db/whois/query/pipeline/ExceptionHandler.java
```diff
@@ -43,6 +44,14 @@ public void exceptionCaught(final ChannelHandlerContext ctx, final ExceptionEven
             handleException(channel, Collections.singletonList(QueryMessages.inputTooLong()), QueryCompletionInfo.EXCEPTION);
         } else if (cause instanceof IOException) {
             handleException(channel, Collections.<Message>emptyList(), QueryCompletionInfo.EXCEPTION);
+        } else if (cause instanceof DataAccessException) {
+            LOGGER.error("Caught exception on channel id = {}, from = {} for query = {}\n{}",
+                    channel.getId(),
+                    ChannelUtil.getRemoteAddress(channel),
+                    query,
+                    cause.toString());
```
hhazewinkel: Why not adding the cause.toString() always? Are the ones of the 'else' not interesting?

NOTE: in the else is the 'cause' ignored due to less '{}' then parameters.

hhazewinkel:
You are right that the 'implicit' cast from Integer (when value is null) will throw an NPE. The compiler does int a = new Integer(2).intValue(); (adding the intValue() method)

eshryane:
I've combined the separate catch blocks, though I think it's less readable, the intent is to swallow these exceptions (default to returning zero).

whois-query/src/main/java/net/ripe/db/whois/query/pipeline/ExceptionHandler.java
eshryane: The else clause (i.e. the existing code) logs the entire stack trace. 

In case of a DataAccessException, we don't care about the stack trace - just log that it's happened. This exception can happen if a database operation is interrupted on shutdown.

hhazewinkel:
Looks fine to me

PR #359, opened 2016-04-19
Drop org.jetbrains dependency.

hhazewinkel:


eshryane:
Looks good to me!

tpolychniatis:
looks good!

PR #361, opened 2016-04-20
Package minor updates

hhazewinkel:
This updates various package with minor versions.
It also updates Jackson, but creates a slight different behavior of JSON in some cases.
In particular, 'null' is returned as a value.

I would suggest we decide whether the jackson update is valuable to keep or drop that part.


eshryane:
reviewed release notes for changed dependencies, couldn't find any significant changes (apart from Jackson where the behaviour on null elements changed, fixed by setting a flag).

PR #364, opened 2016-04-29
Pipeline Transformer

hhazewinkel:
- dummy (for testing)
    - auto number


whois-update/src/main/java/net/ripe/db/whois/update/handler/transformpipeline/TransformPipeline.java
```diff
@@ -0,0 +1,31 @@
...
+
+import net.ripe.db.whois.common.rpsl.RpslObject;
+import net.ripe.db.whois.update.domain.Action;
+import net.ripe.db.whois.update.domain.Update;
+import net.ripe.db.whois.update.domain.UpdateContext;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
+public class TransformPipeline {
```
petchells: I'm getting a bit muddled up between a PipelineTransformer and a TransformPipeline. Maybe use another name for one of them?

hhazewinkel: I could not come up with a better name, but how about this rename:
PipelineTransformer -->> Transformer; .i.e  the interface of a specific 'transformer' implementation (class)
- TransformerPipeline -->> TransformerChain;  i.e. the class (imeplementation) that loops thru all the Transformer implementations

dacruz:
👍

PR #372, opened 2016-05-09
Validation tests

8v060htwyc:


hhazewinkel:
My main comment would be that this pull request does various things.
It does not just add validation tests, but also changes validators.

PR #373, opened 2016-05-09
Nrtm pool size

eshryane:
Increase the NRTM scheduler pool size (from 10 to 32), to allow more clients to use keepalive.
Also increase the executor pool size (from 8 to 32), to handle more concurrent (non-keepalive) requests.


dacruz:
👍

PR #390, opened 2016-06-30
Remove spring related dependencies from whois-rpsl module

tpolychniatis:
Hi,
This PR which removes the spring dependencies from whois-rpsl module, so that we can keep the module as small as possible. 

The problem is that the aspects library used, pulls a lot of transitive dependencies from spring, but whois-rpsl module does not need them.
Using the "mvn dependency:tree" command you can see that after the pom changes the following dependencies are removed from the whois-rpsl module:
[INFO] +- org.springframework:spring-context:jar:4.2.5.RELEASE:compile
[INFO] |  +- org.springframework:spring-aop:jar:4.2.5.RELEASE:compile
[INFO] |  |  \- aopalliance:aopalliance:jar:1.0:compile
[INFO] |  +- org.springframework:spring-beans:jar:4.2.5.RELEASE:compile
[INFO] |  +- org.springframework:spring-core:jar:4.2.5.RELEASE:compile
[INFO] |  |  \- commons-logging:commons-logging:jar:1.2:compile
[INFO] |  \- org.springframework:spring-expression:jar:4.2.5.RELEASE:compile
[INFO] +- org.springframework:spring-tx:jar:4.2.5.RELEASE:compile

As discussed, I did not move everything in whois-rpsl in a module specific package, so that we can keep the changes to a minimum.

Please read the comments in the parent pom.xml for more info.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -791,22 +791,40 @@
                 <configuration>
                     <source>${java.version}</source>
                     <target>${java.version}</target>
-                    <!-- TODO [TP]: if spring-aspects lib is removed from whois-rpsl/pom.xml,
-                                    then we need to exclude the paths to that module here.
-
-                                    In any case, we do not have aspects in that module.
-
-                                    (It would be even nicer if we had inclusion paths
-                                    for the modules/packages we actually need aspects)
-
+                    <!-- [TP] This is a bit ugly part.
+                        aspectjweaver, spring-context, spring-tx libs were removed from whois-rpsl/pom.xml, because they
+                        pull many deps, and no aspects are/should be present in this module. By removing the libs, a lot of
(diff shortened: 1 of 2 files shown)


hhazewinkel:
As you mention... Is it not wise to perform also a package rename in order to 'fix' this?

tpolychniatis:
There is not just one package. There are a few packages which are shared with other modules too. Therefore what should be done, is create a unique package for the module and move classes and other packages as sub packages to the new one. By doing this refactoring, the affected files would be really a lot. Imagine that e.g. RpslObject appears in most of the classes of every module.

Regardless of the effort, I suggested performing the refactoring, but after speaking with the team, I was asked not to (for now), and just remove (most of) the spring dependencies, which I did.

PR #393, opened 2016-07-11
Allow notset status modification

r0bmiller:


whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/update/InetnumStatusNotSetSpec.groovy
```diff
@@ -0,0 +1,284 @@
...
+                source:     TEST
+                """
+        ]}
+
+    //TODO: ALLOCATED should NOT be allowed but currently is
+    def "change NOT-SET status to allowed statuses with ALLOCATED UNSPECIFIED parent"() {
+        given:
+        syncUpdate(getTransient("ALLOC-UNS") + "password: owner3\npassword: hm\npassword: lir")
+        syncUpdate(getTransient("NOTSET") + "override: denis,override1")
+
```
hhazewinkel: Why adding data with a specific password, if you use for the other the override?
Just using override for all could simplify code (that part - the initial data insertion - 
is not really part of the test, I would say)

whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/update/InetnumStatusNotSetSpec.groovy
```diff
@@ -0,0 +1,284 @@
...
+@org.junit.experimental.categories.Category(IntegrationTest.class)
+class InetnumStatusNotSetSpec extends BaseQueryUpdateSpec {
+
+    // in the database the only parent that exists for NOT-SET resources is ALLOCATED UNSPECIFIED
+    // and the only child resource status with a NOT-SET resource as a parent is LEGACY
+    @Override
+    Map<String, String> getTransients() {
+        [
+                "ALLOC-UNS": """\
+                inetnum:      192.168.0.0 - 192.169.255.255
```
hhazewinkel: Why using the 192.168.0.0 - 192.169.255.255 range as this is part 'private networks' and some not. I would suggest to use 192.168.0.0 - 192.168.255.255 as toplevel.

Yes, I saw it in many other places too.

whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/update/InetnumStatusNotSetSpec.groovy
r0bmiller: Agreed, is unnecessary.

PR #396, opened 2016-07-14
Switch embedded Jetty from 8 to 9 (latest)

eshryane:
Start using Jetty 9 for embedded servlet container / http server, for latest bug fixes and features.
Should be completely backwards compatible.
Only concern is different CPU or memory usage in production (need to load test in prepdev).


hhazewinkel:
Code change look fine.
I have found some references that Jetty 9.x should be 30% faster, so I guess the CPU should be fine. (although, most of the processing is the WHOIS application I presume)

PR #399, opened 2016-08-10
implemented override for ReferencesServices delete call

r0bmiller:


eshryane:
http://gph.is/2a3zNOI

PR #401, opened 2016-08-24
Mnt routes link

eshryane:


whois-api/src/main/java/net/ripe/db/whois/api/rest/mapper/FormattedServerAttributeMapper.java
```diff
@@ -56,4 +63,18 @@ private static String getAttributeValue(final Attribute attribute) {
...
     }
+
+    private static String getLinkKey(final AttributeType attributeType, final CIString value) {
+        switch (attributeType) {
+            case MNT_ROUTES:
+                try {
+                    final MntRoutes mntRoutes = MNT_ROUTES_PARSER.parse(value.toString());
+                    return mntRoutes.getMaintainer().toString();
+                } catch (AttributeParseException e) {
+                    return value.toString();
```
dacruz: If we cannot parse it, we return "value". 
But, in this case, we are going the have a broken link. Am I correct?!

eshryane: This will work if the mnt-routes: value is syntactically correct. We presume objects are well-formed. If not correct, the original value is used (same behaviour as before).

dacruz:
👍

PR #406, opened 2016-09-20
added comment to whois.properties and debug option for locally running server

warpian:


whois-api/src/test/java/net/ripe/db/whois/api/AbstractIntegrationTest.java
```diff
@@ -32,4 +34,23 @@ public void stopServer() throws Exception {
...
+        Instant start = Instant.now();
+
+        while (!Thread.currentThread().isInterrupted()) {
+            try {
+                Instant end = Instant.now();
+                Duration timeElapsed = Duration.between(start, end);
+                System.out.println(String.format("Server listening for %d minutes on port %d", timeElapsed.toMinutes(), getPort()));
+                wait(60000);
+            } catch (InterruptedException ex) {
+                Thread.currentThread().interrupt();
```
eshryane: why do we need to interrupt the thread again? Can we just re-throw?
Or, use something like Uninterruptibles.sleepUninterruptibly(10, TimeUnit.MINUTES);

petchells:
nice addition. can be merged, for my money.

whois-api/src/test/java/net/ripe/db/whois/api/AbstractIntegrationTest.java
eshryane: This line clears the interrupted state (caused by the InterruptedException), so will the isInterrupted() check ever return true?

eshryane: is the interrupt() needed if the wait threw an InterruptedException? The thread isInterrupted() should still be true, even without the interrupt() ?

eshryane: This link explains what's going on here: http://stackoverflow.com/questions/1087475/when-does-javas-thread-sleep-throw-interruptedexception

petchells: if 'isInterrupted()' is called and returns 'true' then the next invocation will return 'false' since the method resets the thread's internal flag. It's better to explicitly call interrupt() here so that anything watching this thread will get the right state.

PR #408, opened 2016-11-09
added handling for IllegalArgumentExceptions

warpian:


whois-api/src/main/java/net/ripe/db/whois/api/rest/DomainObjectService.java
```diff
@@ -106,10 +106,15 @@ public Response create(
 
         } catch (UpdateFailedException e) {
             return createResponse(e.status, e.whoisResources);
+
+        } catch (IllegalArgumentException e) {
+            return createResponse(BAD_REQUEST, e.getMessage());
```
eshryane: Does db-web-ui handle 400 Bad Request separately?

petchells:
Worth having -- it's good to respond to 'illegal args' with 'bad request'

PR #409, opened 2016-12-06
detect fails when mask doesn't cover inetnum

petchells:
compares mask against bits in the inetnum and fails if any aren't covered

### DIFF
--- a/whois-query/src/test/java/net/ripe/db/whois/query/support/QueryExecutor.java
+++ b/whois-query/src/test/java/net/ripe/db/whois/query/support/QueryExecutor.java
@@ -9,9 +9,8 @@ import net.ripe.db.whois.common.io.ByteArrayInput;
 import net.ripe.db.whois.common.rpsl.ObjectType;
 import net.ripe.db.whois.common.rpsl.RpslObject;
 import net.ripe.db.whois.common.support.ByteArrayContains;
-import net.ripe.db.whois.common.support.TelnetWhoisClient;
 import net.ripe.db.whois.common.support.QueryExecutorConfiguration;
-import net.ripe.db.whois.query.acl.AccessControlListManager;
+import net.ripe.db.whois.common.support.TelnetWhoisClient;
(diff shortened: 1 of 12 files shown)


eshryane:
Looks good to me

eshryane:
Don't merge yet - a test is failing in DomainObjectServiceTestIntegration:
create_domain_object_unparsable_json

eshryane:
There are also failing Groovy tests (*Spec.groovy in whois-endtoend).

petchells:
All tests have been fixed: http://bamboo.ripe.net/browse/DB-WHOIS119

eshryane:
Reviewing now

whois-query/src/test/java/net/ripe/db/whois/query/support/QueryExecutor.java
```diff
@@ -31,12 +30,10 @@
     private static final Set<String> ERRORS_TO_IGNORE = Sets.newHashSet("ERROR:101:", "ERROR:106:");
 
     private final QueryExecutorConfiguration configuration;
-    private final AccessControlListManager accessControlListManager;
     private final Logger logger;
 
-    public QueryExecutor(final QueryExecutorConfiguration configuration, final AccessControlListManager accessControlListManager, final Logger logger) throws UnknownHostException {
```
eshryane: Confirmed aclManager is not needed in test support class - personal object accounting is done with hardcoded object type check.

agoston:
Haha, we had the exact same issue years back. Turns out netmasks like 10.0.0.1/8 are perfectly valid as per RFC. Hence we left this in there intentionally.

agoston:
It's called combined host and network addresses, and is already used in RIPE DB. I wouldn't go and implement this without announcement, it's actually a handy feature to be able to specify a specific host inside the network.

Another issue is that it's implemented in IPV4Resource, so it pretty much applies to _every_ attribute. Note, this is a major change, even if it doesn't seem so. Announcement will definitely be necessary if you're planning to go through with this.

PR #414, opened 2017-01-19
Implement Zonemaster DNS checks (DB-175)

eshryane:
As GII I want Whois to use Zonemaster instead of DNSCheck so that we have a modern, less buggy interface


whois-update/src/main/java/net/ripe/db/whois/update/dns/zonemaster/domain/ZonemasterRequest.java
```diff
@@ -0,0 +1,62 @@
...
+public abstract class ZonemasterRequest {
+
+    @JsonProperty("jsonrpc")
+    private String jsonRpc = "2.0";
+    @JsonProperty
+    private int id;
+    @JsonProperty
+    private String method;
+
+    public void setMethod(final ZonemasterRequest.Method method) {
```
petchells: should be a protected method

whois-update/src/main/java/net/ripe/db/whois/update/dns/zonemaster/ZonemasterDnsGateway.java
```diff
@@ -0,0 +1,164 @@
...
+import static net.ripe.db.whois.common.Messages.Type.ERROR;
+
+@DeployedProfile
+@Component
+public class ZonemasterDnsGateway implements DnsGateway {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ZonemasterDnsGateway.class);
+
+    private static final int TEST_PROGRESS_SLEEP_SECONDS = 1;
+    private static final int TEST_PROGRESS_MAXIMUM_RETRIES = 60 * 5;
```
petchells: I suggest increasing the sleep time then pinging for a result fewer times. e.g. sleep 5 sec and retry 50 times (250 sec in total)

eshryane: I picked the values at random: retry every second, for a maximum of 5 minutes. I'd prefer to keep the interval short, as zonemaster usually completes within a short amount of time - a 5 second delay may be mostly wasted time, if we happen to just miss the completion. 
I'm assuming retries are practically free.
I can ask Anand how long we should retry for, maybe 5 minutes is indeed excessive.

whois-update/src/main/java/net/ripe/db/whois/update/dns/zonemaster/domain/ZonemasterRequest.java
eshryane: Will fix this, thanks

whois-update/src/main/java/net/ripe/db/whois/update/dns/zonemaster/ZonemasterDnsGateway.java
eshryane: I've increased the polling interval to 5 seconds, upside is less network traffic and load on loadmaster, downside is that user may wait for longer than necessary.
I left the number of retries at 60, so poll for 5 minutes maximum, as Anand suggested a zonemaster check may take this long.

whois-update/src/main/java/net/ripe/db/whois/update/dns/zonemaster/domain/ZonemasterRequest.java
eshryane: fixed

PR #417, opened 2017-01-31
Db 96 notify pgp

dacruz:


whois-update/src/main/java/net/ripe/db/whois/update/domain/Update.java
```diff
@@ -93,11 +94,22 @@ public String toString() {
         return builder.toString();
     }
 
-    public void setEffectiveCredential(final String effectiveCredential) {
+
+    public void setEffectiveCredential(final String effectiveCredential, EffectiveCredentialType effectiveCredentialType) {
```
eshryane: Update should not be mutable, we should find a better way to track the effective credential

whois-update/src/main/java/net/ripe/db/whois/update/domain/Update.java
```diff
@@ -16,6 +16,7 @@
     private final List<String> deleteReasons;
     private final RpslObject submittedObject;
     private String effectiveCredential;
+    private EffectiveCredentialType effectiveCredentialType;
```
eshryane: mutable field

whois-update/src/main/java/net/ripe/db/whois/update/domain/Update.java
petchells: missing 'final'

whois-update/src/test/java/net/ripe/db/whois/update/authentication/credential/PgpCredentialValidatorTest.java
```diff
@@ -322,7 +377,14 @@ public void knownCredentialIsInvalid() {
         final RpslObject keycertObject = RpslObject.parse("key-cert: PGPKEY-5763950D");
         when(rpslObjectDao.getByKey(ObjectType.KEY_CERT, keycertObject.getKey().toString())).thenReturn(keycertObject);
 
-        assertThat(subject.hasValidCredential(update, updateContext, Sets.newHashSet(offeredCredential), knownCredential), is(false));
+        assertThat(subject.hasValidCredential(preparedUpdate, updateContext, Sets.newHashSet(offeredCredential), knownCredential), is(false));
         verify(loggerContext).logString(any(Update.class), anyString(), anyString());
     }
+
+    private Update getUpdate() {
```
eshryane: change method name to createUpdate() ? it's confusing

whois-update/src/test/java/net/ripe/db/whois/update/authentication/credential/PgpCredentialValidatorTest.java
```diff
@@ -35,13 +45,16 @@
...
-    @Mock private PreparedUpdate update;
+    @Mock private PreparedUpdate preparedUpdate;
+
     @Mock private UpdateContext updateContext;
     @Mock private RpslObjectDao rpslObjectDao;
     @Mock private DateTimeProvider dateTimeProvider;
     @Mock private LoggerContext loggerContext;
     @InjectMocks private PgpCredentialValidator subject;
 
+    private final Update update = getUpdate();
```
eshryane: final Update instance is shared between tests, but contains mutable state. Better to remove final and create in Before method instead

whois-update/src/main/resources/templates/notification.vm
```diff
@@ -93,10 +93,15 @@ $update.updatedObject
 #end
 $!update.reason
 
-#if ($update.preparedUpdate.update.effectiveCredential)
+#if ($update.preparedUpdate.update.effectiveCredential && $update.preparedUpdate.update.effectiveCredentialType ==  'SSO')
```
eshryane: can we move logic outside the template?

petchells: don't need the first part of the test -- it's implied by the second part

eshryane:
The MailGateway interface was extended with a method:
{{{
    void sendEmail(String to, ResponseMessage responseMessage, String replyTo);
}}}
However, the ResponseMessage already contains a replyTo field, so the String replyTo argument is redundant.

Can you refactor this method and calls to it?

PR #418, opened 2017-02-02
DB-322 Notification text changed followed with integration test changes

isvonja:


whois-update/src/test/java/net/ripe/db/whois/update/handler/response/ResponseFactoryTest.java
```diff
@@ -621,4 +629,16 @@ private void assertVersion(final String response) {
                 "Handled mail update \\(TEST, 1970-01-01 00:00:00\\)\n" +
                 ".*"));
     }
+
+    private void assertVersionWithoutContact(final String response) {
```
eshryane: Is there any need for this method, if assertNotification already checks the content?

whois-update/src/main/resources/templates/notification.vm
```diff
@@ -134,5 +133,11 @@ $!update.reason
 #end
 #end
 
-#parse("/templates/include/terms-and-conditions.vm")
+>>>>>
```
eshryane: can you include the terms-and-conditions.vm template instead?

PR #423, opened 2017-03-30
Handle "missing crc check" in PGP signed update message.

eshryane:


eshryane:
Integration and e2e tests passing

PR #426, opened 2017-04-26
Refactored Transfer Code

eshryane:
- use final everywhere
- use isEmpty() rather than size() == 0
- simplified code
- removed verbose logging


whois-api/src/main/java/net/ripe/db/whois/api/transfer/logic/asn/stages/CreateNewFollowingBlockStage.java
```diff
@@ -35,8 +36,8 @@ protected String getName() {
         begin = end = transfer.getResource().asBigInteger().longValue();
 
         final RpslObject newAsBlock = createAsBlock(begin, end, blockTemplate);
-        requests.add(new ActionRequest(newAsBlock, Action.CREATE));
 
+        requests.add(new ActionRequest(newAsBlock, Action.CREATE));
```
lolepezy: Why not Collections.singletonList ?

eshryane: Fixed

PR #428, opened 2017-05-01
Refactored hashcode, equals, toString methods

eshryane:


whois-update/src/main/java/net/ripe/db/whois/update/keycert/X509CertificateWrapper.java
```diff
@@ -75,21 +76,17 @@ private String convertFromRfc2253ToCompatFormat(String name) {
...
+        if (o == null || getClass() != o.getClass()) return false;
 
-        if (o == null || getClass() != o.getClass()) {
-            return false;
-        }
+        final X509CertificateWrapper that = (X509CertificateWrapper) o;
 
-        return ((X509CertificateWrapper) o).getCertificate().equals(
-                this.getCertificate());
+        return Objects.equals(certificate, that.certificate);
```
dacruz: should we use **this.**certificate?!

eshryane: I dropped this, to be consistent with the other equals() methods. I think it's clear from the context what it is.

PR #429, opened 2017-05-01
Update pom.xml

avengers444:


eshryane:
Thanks for your pull request. However, only the http scheme in the pom.xml has changed, is this deliberate? I'm closing for now.

PR #436, opened 2017-06-27
Change org name

eshryane:
Make validation of changes to the org-name: attribute case-sensitive.


eshryane:
CI build and Integration tests have passed.

PR #438, opened 2017-06-29
Remove generated autnum remarks

eshryane:
Don't generate a remarks attribute for legacy resources in aut-num objects any more.


eshryane:

PR #441, opened 2017-07-24
Improved Search API

eshryane:
Added support for additional parameters to the Whois Search API:

* managed-attributes
  * Annotate whether an attribute is managed by the RIPE NCC (as a boolean flag)
  * Adds managed="true" at the attribute level, there's nothing at the object level
* resource-holder
  * Adds the responsible organisation name and org-id for a resource (inetnum, inet6num, aut-num only)
* abuse-contact
  * Adds the abuse contact for a resource or organisation.
* limit
  * limit the number of objects returned (e.g. paging)
* offset
  * starting point in results to return from (e.g. paging).


eshryane:
All integration tests passed

eshryane:
All integration tests passed (commit b9e68b7)

PR #446, opened 2017-08-15
Remove dnscheck data source (obsolete after we moved to zonemaster)

eshryane:


eshryane:
All integration tests have passed

PR #449, opened 2017-08-31
Db 910 abuse finder consider abuse c on resources

sanderbuskens:


whois-query/src/main/java/net/ripe/db/whois/query/planner/AbuseCFinder.java
```diff
@@ -76,18 +76,32 @@ public RpslObject getAbuseContactRole(final RpslObject object) {
     @CheckForNull
     @Nullable
     private RpslObject getAbuseContactRoleInternal(final RpslObject object) {
+        RpslObject abuseContact;
```
eshryane: Can this code be simplified? It's quite hard to follow

PR #450, opened 2017-08-31
Use local database for referenced type lookups

eshryane:
Use local database for reference type lookups from REST and Search APIs.


petchells:
looks good to me

PR #452, opened 2017-09-12
Db 917 inverse pn

petchells:
https://jira.ripe.net/browse/DB-917

### DIFF
--- a/whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/query/AbuseQuerySpec.groovy
+++ b/whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/query/AbuseQuerySpec.groovy
@@ -952,16 +952,6 @@ class AbuseQuerySpec extends BaseQueryUpdateSpec {
         queryObject("-i pn AH2-TEST", "aut-num", "AS200")
     }
 
-    def "query for abuse-c organisation"() {
-        given:
-        databaseHelper.addObject(getTransient("ROLE"))
-        databaseHelper.addObject(getTransient("ORG-W-ABUSE_C"))
-        databaseHelper.addObject(getTransient("AUTNUM"))
-        expect:
-        queryObject("-rBG -T organisation ORG-FO1-TEST", "abuse-c", "AB-TEST")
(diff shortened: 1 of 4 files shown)


whois-query/src/test/java/net/ripe/db/whois/query/integration/AbuseCTestIntegration.java
```diff
@@ -45,10 +45,21 @@
...
+            "abuse-mailbox: abuseperson@ripe.net\n" +
+            "mnt-by:        TEST-MNT\n" +
+            "source:        TEST",
+
             "organisation:  ORG-TEST-1\n" +
             "abuse-c:       ABU-TEST\n" +
             "source:        TEST",
 
+            "organisation:  ORG-TEST-ABUSEC-PERSON\n" +
+            "abuse-c:       ABUSEC-PERSON-TEST\n" +
```
eshryane: This can't happen, only a role can be used as an abuse-c reference, not a person. Can you change the test data to use a role instead?

whois-query/src/test/java/net/ripe/db/whois/query/query/QueryTest.java
```diff
@@ -500,14 +500,14 @@ public void multiple_proxies() {
     public void attributes_person() {
         final Query query = Query.parse("-i person name");
```
eshryane: Also test the short format ('pn') of person

whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/query/AbuseQuerySpec.groovy
```diff
@@ -907,4 +907,30 @@ class AbuseQuerySpec extends BaseQueryUpdateSpec {
         queryObject("--abuse-contact AS200", "abuse-mailbox", "more_abuse@test.net")
     }
 
+    def "inverse query for aut-num using person"() {
```
eshryane: Can you also add a test case for the usual reference from an organisation to a role using an abuse-c reference?

PR #460, opened 2017-09-27
abuse-mailbox code cleanup

sanderbuskens:


whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/integration/MaintainerIntegrationSpec.groovy
```diff
@@ -304,38 +304,7 @@ class MaintainerIntegrationSpec extends BaseWhoisSourceSpec {
...
-            source: TEST
-            password: update
-            """.stripIndent())
-
-        when:
-        def response = syncUpdate update
-
-        then:
-        response =~ /Modify SUCCEEDED: \[mntner] DEV-MNT/
+        response =~ /"abuse-mailbox" is not valid for this object type/
```
eshryane: Why was this error message changed? The abuse-mailbox can only be added to role objects. Is this generic error message good enough?

whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/update/IrtSpec.groovy
```diff
@@ -459,9 +459,7 @@ class IrtSpec extends BaseQueryUpdateSpec {
         ack.countErrorWarnInfo(1, 0, 0)
         ack.errors.any { it.operation == "Create" && it.key == "[irt] irt-test" }
         ack.errorMessagesFor("Create", "[irt] irt-test") ==
-                ["\"abuse-mailbox:\" can only be added to ROLE objects intended to be " +
-                 "referenced through the \"abuse-c:\" attribute in ORGANISATION, " +
-                 "INET(6)NUM and AUT-NUM objects."]
+                ["\"abuse-mailbox\" is not valid for this object type"]
```
eshryane: More specific error message removed?

whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/integration/MaintainerIntegrationSpec.groovy
sanderbuskens: The outcome of this test is now that it fails because abuse-mailbox is no longer valid for the mntner object. The test tests for creation for a maintainer with an abuse-mailbox. I think it's a good thing to keep the test because it proves that this is no longer possible.

whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/update/IrtSpec.groovy
sanderbuskens: The abuse-mailbox attribute is now no longer allowed on IRT objects, so the error message is correct in my opinion..

eshryane: Ok! It's the generic error message for any invalid attribute.

whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/integration/MaintainerIntegrationSpec.groovy
eshryane: Ok

whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/update/IrtSpec.groovy
sanderbuskens: Yes, we could say that maybe the test is no longer necessary as there's no point in testing all the attributes that are not allowed on the object, but seen as abuse-mailbox was allowed before but will no longer be allowed I think it's good to have captured this change in behaviour for regression test purposes..

PR #471, opened 2018-01-31
Separate validation for LIR organisation attributes

eshryane:
Separate validation for LIR organisation attributes that can be managed by the user in the LIR Portal.


eshryane:
Integration tests passing

PR #474, opened 2018-02-08
Do not allow any specified email addresses to be used in any attribute with email syntax.

eshryane:
Do not allow any specified email addresses to be used in any attribute with email syntax.


eshryane:

PR #475, opened 2018-02-14
Disallow inconsistent formatting in the Organisation org-name attribute value

eshryane:
Disallow multiple spaces, any tab characters, or newlines within the org-name attribute, so that the org name is consistent for making comparisons.

Discussion on db-wg: https://www.ripe.net/ripe/mail/archives/db-wg/2017-November/005737.html


eshryane:
Investigating failed integration tests

eshryane:
All integration tests now pass (fixed in 4e3679e968d213b026132517ff9b5a3e86ac0db1

PR #476, opened 2018-02-15
Nrtm keepalive end of stream

eshryane:
Add an end of stream indicator comment at the end of each NRTM update in keepalive mode.

Ref. https://www.ripe.net/ripe/mail/archives/db-wg/2017-November/005725.html


eshryane:
integration tests have passed (commit id 42e6d34)

PR #487, opened 2018-07-17
Out of region validations

sbusk:


whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/BaseQueryUpdateSpec.groovy
```diff
@@ -48,6 +57,20 @@ abstract class BaseQueryUpdateSpec extends BaseEndToEndSpec {
         getDatabaseHelper().addObjects(rpslObjects)
     }
 
+    private void allAuthoritativeResources() {
+        if (!getAuthoritativeResources().isEmpty()) {
+            getAuthoritativeResourceDao().delete("test", "0.0.0.0/0")
```
eshryane: Add comment why this is done (e.g. all address space is in-region by default, remove these defaults if out-of-region objects added).

sbusk: Done.

PR #494, opened 2018-07-25
Remove mnt-routes from autnum as it's not longer needed.

eshryane:
Also need to filter out mnt-routes with warning on update.


eshryane:
Waiting for integration tests

PR #499, opened 2018-08-07
Fixed rtr-set members and mp-members syntax

eshryane:
Allow single IPv4 addresses (and not IPv4 prefixes) in rtr-set members and mp-members attributes.


eshryane:
One object affected by this change, and needs to be updated: rtr-set:        rtrs-stanbicbank-uganda
members:        196.8.207.0/24
members:        196.8.208.0/24

eshryane:
@job This fix was deployed to production in Whois 1.92.4 yesterday

PR #510, opened 2018-09-11
DB-1783: Cleaning up pending code.

gabriel-a:
Cleaning up NWI-5 Pending code.
- Removed the code related to Pending


eshryane:
Need to remove any remaining "pending" references.

PR #511, opened 2018-09-18
Case Sensitive Source Bug

eshryane:
Make comparison of source attribute value case-insensitive. Fixes reported bug where an object has "source: ripe" in the database, and a REST API lookup didn't find it.

### DIFF
--- a/whois-api/src/main/java/net/ripe/db/whois/api/rest/SourceResolver.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/rest/SourceResolver.java
@@ -7,47 +7,44 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 
-import static net.ripe.db.whois.common.rpsl.ObjectType.AUT_NUM;
-
 @Component
 public class SourceResolver {
 
     private final AuthoritativeResourceData authoritativeResourceData;
-    private final String mainSourceNameString;
-    private final String nonAuthSourceName;
+    private final CIString mainSource;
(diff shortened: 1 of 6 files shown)


eshryane:
Integration and E2E tests passed

whois-api/src/main/java/net/ripe/db/whois/api/rest/SourceResolver.java
```diff
@@ -7,47 +7,44 @@
...
     public SourceResolver(final AuthoritativeResourceData authoritativeResourceData,
-                          @Value("${whois.source}") final String mainSourceNameString,
-                          // TODO [SB]: we need to either use the property or the SourceContext for the nonauth source
-                          @Value("${whois.nonauth.source}") final String nonAuthSourceName) {
+                          @Value("${whois.source}") final String mainSource,
+                          @Value("${whois.nonauth.source}") final String nonAuthSource) {
         this.authoritativeResourceData = authoritativeResourceData;
-        this.mainSourceNameString = mainSourceNameString.toLowerCase();
-        this.nonAuthSourceName = nonAuthSourceName.toLowerCase();
+        this.mainSource = CIString.ciString(mainSource);
```
sbusk: Maybe use static import for ciString

eshryane: I prefer including the class name, so it's clear where it's calling

eshryane: I don't see a clear precedent in the code - it's approximately equal between CIString.ciString() and static import ciString().

PR #518, opened 2019-01-15
Filter out comment in Autocomplete response

eshryane:
~~Fixed by filtering out all comments when objects are added to the Lucene fulltext index (previously attribute value and any comment(s) were copied straight into the same index field). Downside is that comments are no longer searchable in Fulltext Search.~~

Considered separating comment into a separate, searchable field - the Lucene field only holds a key/value pair, no room to separate comment.

The fulltext search strips newlines out of the attribute value, so we can't parse the value properly in the Autocomplete service - we just strip out everything past the first comment.

### DIFF
--- a/whois-api/src/main/java/net/ripe/db/whois/api/autocomplete/AutocompleteSearch.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/autocomplete/AutocompleteSearch.java
@@ -28,11 +28,13 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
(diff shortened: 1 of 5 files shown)


eshryane:

eshryane:
TODO: need to account for multiple values in some attributes, e.g.
* holes:
* member-of:
* members:
* mnt-by:
* mnt-ref:
* mp-members:

May need to filter out comment separately.

Also may want to allow comments in the full text index, so users can search in them.

eshryane:
Lucene field is key/value, cannot add comment separately to a field.

eshryane:
Integration tests have passed (commit 06d77a)

eshryane:
Integration tests passed commit b96ba3f

PR #519, opened 2019-01-16
Added bogon validator.

eshryane:


eshryane:
Integration tests passed (commit id 0x440b7c3)

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/route/BogonValidator.java
```diff
@@ -0,0 +1,111 @@
...
+
+    @Override
+    public void validate(final PreparedUpdate update, final UpdateContext updateContext) {
+        final RpslObject updatedObject = update.getUpdatedObject();
+        if (updatedObject == null) {
+            return;
+        }
+
+        final Optional<CIString> prefix = getPrefix(updatedObject);
+        if (prefix.isPresent() && isBogon(prefix.get().toString())) {
```
sbusk: you could use: prefix.ifPresent(prefix -> if (isBogon(prefix.toString())) ...

PR #520, opened 2019-01-22
Warn the user if the resource abuse-c is identical to the organisation's abuse-c

eshryane:


eshryane:
Integration tests passed (commit 796932c)

PR #521, opened 2019-01-25
Remove transfers logic (used for placeholders)

eshryane:


eshryane:
Integration tests passed 7095b0e

PR #522, opened 2019-02-05
Do not Allow Expired Messages, Revoked or Expired Keys

eshryane:
* Do not allow a PGP or X509 signed Whois Update after 1 hour (was 1 week).
* Do not allow expired PGP or X509 keys to be used to sign updates (used to only warn).
* Do not allow revoked PGP keys to be used to sign updates.


eshryane:
Integration tests passed 1f184f7

PR #523, opened 2019-02-06
NRTM Client End of Stream

eshryane:
Allow NRTM client to skip the end of stream comment in keepalive mode.


eshryane:
Integration tests have passed cd1e145

PR #527, opened 2019-02-28
Xss fix

sbusk:


sbusk:
A test needs to be added still (for normal search return values and for highlighted search return values).

sbusk:
Closing pull request, we'll only be fixing this in the frontend, not in the whois response.

PR #532, opened 2019-04-12
Do not allow local-only addresses, i.e. missing @domain part

eshryane:


eshryane:
Integration tests passed

PR #536, opened 2019-05-08
Character Substitutions

eshryane:
* Convert Strings into Latin-1
* Substitute control characters with '?'.
* Substitute regular space for non-break space.


eshryane:
Integration tests now passing

eshryane:
As discussed, the warning message (that characters were substituted) doesn't always appear, we need to investigate further.

eshryane:
Integration tests have now passed

PR #542, opened 2019-05-28
send as-block incase autnum is not found

maggarwal13:
404 Not Found is returned for reserved AS numbers.

For example: curl -v https://rdap.db.ripe.net/autnum/65535

Instead, return the AS64496 - AS131071 as-block.

Example from ARIN:
https://rdap.arin.net/registry/autnum/65535


whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -5,29 +5,14 @@
...
-import net.ripe.db.whois.api.rdap.domain.Entity;
-import net.ripe.db.whois.api.rdap.domain.Event;
-import net.ripe.db.whois.api.rdap.domain.Ip;
-import net.ripe.db.whois.api.rdap.domain.Link;
-import net.ripe.db.whois.api.rdap.domain.Nameserver;
-import net.ripe.db.whois.api.rdap.domain.RdapObject;
-import net.ripe.db.whois.api.rdap.domain.Remark;
-import net.ripe.db.whois.api.rdap.domain.Role;
-import net.ripe.db.whois.api.rdap.domain.SearchResult;
+import net.ripe.db.whois.api.rdap.domain.*;
```
eshryane: We have a convention to define each import separately, can you update your settings in IntelliJ?

* In Preferences: Editor -> Code Style -> Java
** Enable "Use single class import"
** Set Class count to use import with '*' to 999

In this case, we want to see every import individually, it makes it easier to read them all, and especially to detect changes.

In general, please follow any existing code style.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -143,6 +111,9 @@ private RdapObject getRdapObject(final String requestUrl, final RpslObject rpslO
             case AUT_NUM:
                 rdapResponse = createAutnumResponse(rpslObject);
                 break;
+            case AS_BLOCK:
```
eshryane: Make sure that it's not possible to query for an as-block, only for the autnum.

See RFC7482 for the full list of allowed types: https://datatracker.ietf.org/doc/rfc7482/?include_text=1

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -321,6 +289,22 @@ private static Autnum createAutnumResponse(final RpslObject rpslObject) {
...
     }
 
+    private static Autnum createAsBlockResponse(final RpslObject rpslObject) {
+        final Autnum autnum = new Autnum();
+
+        String key = rpslObject.getValueForAttribute(AttributeType.AS_BLOCK).toString();
+        String[] startAndEndKey = key.split("-");
+
+        autnum.setHandle(startAndEndKey[0]);
+        //TODO :check what should be the name
```
eshryane: Check the RFC's for the definition, or for an example response that includes this.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -321,6 +289,22 @@ private static Autnum createAutnumResponse(final RpslObject rpslObject) {
         return autnum;
     }
 
+    private static Autnum createAsBlockResponse(final RpslObject rpslObject) {
+        final Autnum autnum = new Autnum();
+
+        String key = rpslObject.getValueForAttribute(AttributeType.AS_BLOCK).toString();
```
eshryane: Use final keyword where possible, so variables are never re-defined.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -321,6 +289,22 @@ private static Autnum createAutnumResponse(final RpslObject rpslObject) {
...
+        String[] startAndEndKey = key.split("-");
```
eshryane: Try using AsBlockRange class instead of parsing yourself (and make sure to catch and handle any exception).

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -321,6 +289,22 @@ private static Autnum createAutnumResponse(final RpslObject rpslObject) {
...
+
+        autnum.setHandle(startAndEndKey[0]);
+        //TODO :check what should be the name
+        autnum.setName(key.replace(" ", ""));
+        autnum.setStartAutnum(Long.parseLong(startAndEndKey[0].trim().substring(2)));
+        autnum.setEndAutnum(Long.parseLong(startAndEndKey[1].trim().substring(2)));
+        autnum.setType("DIRECT ALLOCATION");
```
eshryane: Where did you find 'DIRECT ALLOCATION' defined? Also, define as a static field.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapQueryHandler.java
```diff
@@ -0,0 +1,107 @@
...
+import javax.ws.rs.core.Response;
+import java.net.InetAddress;
+import java.util.List;
+import static net.ripe.db.whois.common.rpsl.ObjectType.*;
+
+@Component
+public class RdapQueryHandler {
+
+    private static final int STATUS_TOO_MANY_REQUESTS = 429;
+    private static final Logger LOGGER = LoggerFactory.getLogger(RdapQueryHandler.class);
```
eshryane: Move the LOGGER field up so it's first (it's a generic field in a lot of classes, and by convention we separate it at the top).

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapQueryHandler.java
```diff
@@ -0,0 +1,107 @@
...
+    private final QueryHandler queryHandler;
+
+    @Autowired
+    public RdapQueryHandler(final QueryHandler queryHandler) {
+        this.queryHandler = queryHandler;
+    }
+
+    public List<RpslObject> handleQuery(final Query query, final HttpServletRequest request) {
+
+        final int contextId = System.identityHashCode(Thread.currentThread());
```
eshryane: What's this? Is it necessary? Why not use 0?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapQueryHandler.java
```diff
@@ -0,0 +1,107 @@
...
+                }
+            });
+
+            return result;
+        } catch (final QueryException e) {
+            if (e.getCompletionInfo() == QueryCompletionInfo.BLOCKED) {
+                throw tooManyRequests();
+            } else {
+                LOGGER.error(e.getMessage(), e);
+                throw new IllegalStateException("query error");
```
eshryane: Does this result in a 500 Internal Server Error? We should test for this scenario.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapQueryHandler.java
```diff
@@ -0,0 +1,107 @@
...
+        final List<RpslObject> resultAsBlock = Lists.newArrayList();
+
+        try {
+            queryHandler.streamResults(query, remoteAddress, contextId, new ApiResponseHandler() {
+                @Override
+                public void handle(final ResponseObject responseObject) {
+                    if (responseObject instanceof RpslObject && ((RpslObject) responseObject).getType() == AUT_NUM) {
+                        resultAutNum.add((RpslObject) responseObject);
+
+                        } else if(responseObject instanceof RpslObject && ((RpslObject) responseObject).getType() == AS_BLOCK) {
```
eshryane: Fix formatting - always put the else inside it's own block - it's easier to read, and less error-prone.

```
if () {
} else {
    if () {
    }
}
```

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRequestValidator.java
```diff
@@ -0,0 +1,69 @@
...
+import net.ripe.db.whois.query.acl.AccessControlListManager;
+import net.ripe.db.whois.query.planner.AbuseCFinder;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+import static net.ripe.db.whois.common.rpsl.ObjectType.ORGANISATION;
+
+@Component
+public class RdapRequestValidator {
```
eshryane: You could make this a "Mapper" class instead, and return the parsed value if you need it afterwards.

Also include as block here.

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapObjectMapperTest.java
```diff
@@ -187,6 +182,57 @@ public void autnum() {
         assertThat(result.getPort43(), is("whois.ripe.net"));
     }
 
+    @Test
+    public void asBlock() {
+        final Autnum result = (Autnum) map((RpslObject.parse("" +
+                "as-block:       AS100-AS200\n" +
```
eshryane: as-block objects in the database always include spaces, e.g. "AS100 - AS200", can you fix the test data?

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapRequestValidatorTest.java
```diff
@@ -0,0 +1,92 @@
...
+    private RdapExceptionMapper rdapExceptionMapper;
+
+    private RdapRequestValidator validator;
+
+    @Rule
+    public ExpectedException expectedEx = ExpectedException.none();
+
+    @Before
+    public void setup() {
+        this.validator = new RdapRequestValidator(rdapExceptionMapper);
```
eshryane: You can use @InjectMocks for the validator field, this will inject the rdapExceptionMapper automatically. No need for a setup() method.

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapRequestValidatorTest.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void shouldThrowExceptionForInvalidEntity() {
+        validator.validateEntity("invalidEntity");
+    }
+
+    @Test
+    public void shouldThrowExceptionForInvalidOrganisation() throws Exception{
```
eshryane: Do you need throws Exception? Is there a checked exception?

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapRequestValidatorTest.java
```diff
@@ -0,0 +1,92 @@
...
+        expectedEx.expect(IllegalArgumentException.class);
+        expectedEx.expectMessage("Invalid syntax");
+
+        validator.validateEntity("ORG-Test");
+    }
+
+    @Test(expected = Test.None.class)
```
eshryane: No need for expected

whois-api/src/test/java/net/ripe/db/whois/api/rdap/WhoisRdapServiceTestIntegration.java
```diff
@@ -728,6 +728,33 @@ public void lookup_single_autnum() {
         assertThat(remarks.get(0).getDescription().get(0), is("A single ASN"));
     }
 
+    @Test
+    public void lookup_as_block_when_no_autnum_found() {
+        final Autnum autnum = createResource("autnum/103")
+                .request(MediaType.APPLICATION_JSON_TYPE)
+                .get(Autnum.class);
+
+        assertThat(autnum.getHandle(), equalTo("AS100 "));
```
eshryane: Why is there a trailing space in the handle?

whois-api/src/test/java/net/ripe/db/whois/api/rdap/WhoisRdapServiceTestIntegration.java
```diff
@@ -728,6 +728,33 @@ public void lookup_single_autnum() {
...
+
+    @Test
+    public void lookup_asblock_with_rdap_json_content_type() {
+        final Response response = createResource("autnum/103")
+                .request("application/rdap+json")
+                .get();
+
+        assertThat(response.getMediaType(), is(new MediaType("application", "rdap+json")));
+        final String entity = response.readEntity(String.class);
+        assertThat(entity, containsString("\"handle\" : \"AS100 \""));
```
eshryane: Check trailing space

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
maggarwal13: We already send the hardcoded value  'DIRECT ALLOCATION' for autnum response, I used the same value (check function createAutnumResponse). Although specification says "type -- a string containing an RIR-specific classification of the autnum". not sure if it correct, we can discuss on this.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
maggarwal13: If user tries to query specifically for AS_BLOCK, controller will throw Bad Request Exception.
Query to AS_Block if there is no autnum entry is found. In that case we will return AS_Block

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapQueryHandler.java
maggarwal13: we use the IdentityHashcode for logging purpose only. Not sure if it is helpful in any case

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
eshryane: Great, thanks. Can you add a test that a query for an as-block returns a bad request?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
eshryane: Ok! I see it elsewhere in the code. We can leave as-is for now.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapQueryHandler.java
eshryane: In the query code (WhoisLog.java), the Netty channel id is used for this field. It's not really used by anything, you can use zero instead.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRequestValidator.java
maggarwal13: This class is for only validating the key supplied by user.  If supplied key maps to object type it is valid. we are not using parsed key as this code ultimately calls the same code that query search api calls.
regarding including AS_Block, user cannot request for  as_block as its not allowed.
User can only pass autnum not as block. so validation can happen only on autnum.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -44,31 +51,21 @@
...
-import static net.ripe.db.whois.common.rpsl.AttributeType.GEOLOC;
-import static net.ripe.db.whois.common.rpsl.AttributeType.MNT_BY;
-import static net.ripe.db.whois.common.rpsl.AttributeType.ORG;
-import static net.ripe.db.whois.common.rpsl.AttributeType.ORG_NAME;
-import static net.ripe.db.whois.common.rpsl.AttributeType.PERSON;
-import static net.ripe.db.whois.common.rpsl.AttributeType.PHONE;
-import static net.ripe.db.whois.common.rpsl.AttributeType.ROLE;
-import static net.ripe.db.whois.common.rpsl.AttributeType.TECH_C;
-import static net.ripe.db.whois.common.rpsl.AttributeType.ZONE_C;
+import static net.ripe.db.whois.common.rpsl.AttributeType.*;
```
eshryane: use explicit imports (not star)

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -332,7 +329,24 @@ private static Autnum createAutnumResponse(final RpslObject rpslObject) {
...
+        autnum.setType(DIRECT_ALLOCATION);
+        autnum.getEntitySearchResults().addAll(createContactEntities(rpslObject));
+        return autnum;
+    }
+
+    private static Autnum createAsBlockResponse(final RpslObject rpslObject) {
+        final Autnum autnum = new Autnum();
+
+        final String key = rpslObject.getValueForAttribute(AttributeType.AS_BLOCK).toString();
+        final String handle = key.split("-")[0].replace(" ", "");
```
eshryane: Can you read the handle from the blockRange below? i.e. blockRange.getBegin()

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -332,7 +329,24 @@ private static Autnum createAutnumResponse(final RpslObject rpslObject) {
...
+        final AsBlockRange blockRange = getAsBlockRange(key);
+
+        autnum.setHandle(handle);
+        //TODO :check what should be the name
+        autnum.setName(key.replace(" ", ""));
```
eshryane: Use trim() instead of replace.. ?

eshryane: Or are there internal spaces you also want to replace?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapQueryHandler.java
```diff
@@ -0,0 +1,103 @@
...
+        final List<RpslObject> resultAutNum = Lists.newArrayList();
+        final List<RpslObject> resultAsBlock = Lists.newArrayList();
+
+        try {
+            queryHandler.streamResults(query, remoteAddress, 0, new ApiResponseHandler() {
+                @Override
+                public void handle(final ResponseObject responseObject) {
+                    if (responseObject instanceof RpslObject) {
+
+                        if (((RpslObject) responseObject).getType() == AUT_NUM) {
```
eshryane: re-write as a switch statement? Maybe more readable. Also add a default: with a throw unexpected exception

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
maggarwal13: Handle property has prefix AS. AsRange only holds numeric value

eshryane: There's no need to also parse the key, use the AsRange and convert to a String. It's less logic and easier to understand (imo).
```
final String handle = Long.toString(blockRange.getBegin());
```

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
maggarwal13: oops.. seems like when i use shortcut to remove unused imports Intellij replaces the imports with *, even after setting prefrences :(

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
maggarwal13: Internal spaces, as CreateAutnumResponse also removes internal spaces.
I believe its because in database we always store the block with space (AS0 - AS6) an while returning JSON we remove the spaces

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
eshryane: Found the as-block test that returns 400 bad request, thanks

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapQueryHandler.java
eshryane: It's  not straightforward to test this scenario, and at least it's logged (it doesn't ever happen in production).

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRequestValidator.java
eshryane: Ok!

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/route/OriginValidator.java
```diff
@@ -69,7 +69,7 @@ public void validate(PreparedUpdate update, UpdateContext updateContext) {
         }
     }
 
-    private boolean isReservedAsNumber(Long asn) {
+    public boolean isReservedAsNumber(Long asn) {
```
eshryane: Better not to expose private methods inside validators, but to move the logic out to a separate class.

Can you create something like ReservedAutnum in the same package as net.ripe.db.whois.update.domain.LegacyAutnum? Then move isReservedAsNumber to it.

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapObjectMapperTest.java
```diff
@@ -26,12 +26,7 @@
 import java.util.Collections;
 import java.util.List;
 
-import static org.hamcrest.Matchers.contains;
-import static org.hamcrest.Matchers.containsInAnyOrder;
-import static org.hamcrest.Matchers.emptyIterable;
-import static org.hamcrest.Matchers.hasSize;
-import static org.hamcrest.Matchers.is;
-import static org.hamcrest.Matchers.nullValue;
+import static org.hamcrest.Matchers.*;
```
eshryane: Update imports here too

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
eshryane: Better not to re-format the key here, but also use String.format("...") with the parsed blockRange values instead. 

So you leave the parsing to AsBlockRange, and you specify exactly how to format the handle.

whois-rpsl/src/main/java/net/ripe/db/whois/common/rpsl/attrs/AsBlockRange.java
```diff
@@ -60,4 +61,17 @@ public long getBegin() {
...
+        return String.format(AS_PREFIX_FORMAT, end);
+    }
+
+    public String getBeginWithPrefix() {
+        return String.format(AS_PREFIX_FORMAT, begin);
+    }
+
+    @Override
+    public String toString() {
+        return String.join("-", getBeginWithPrefix(), getEndWithPrefix());
```
eshryane: This is inconsistent with the original input (which has a space before and after the dash)

whois-update/src/main/java/net/ripe/db/whois/update/domain/ReservedAutnum.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.util.List;
+
+@Component
+public class ReservedAutnum {
+
+    private final List<LongRange> reservedAsnumbers;
+
+    @Autowired
+    public ReservedAutnum(@Value("${whois.reserved.as.numbers}") final String reservedAsNumbers) {
```
eshryane: As we discussed, this autowired value could become a static field.

whois-update/src/main/java/net/ripe/db/whois/update/domain/ReservedAutnum.java
```diff
@@ -0,0 +1,46 @@
...
+                return true;
+            }
+            if (asn < range.getMinimumLong()) {
+                break;
+            }
+        }
+        return false;
+    }
+
+    private List<LongRange> parseReservedAsNumbers(final String reservedAsNumbers) {
```
eshryane: This code could disappear if reserved numbers are turned into a field

PR #544, opened 2019-06-03
adding notice and test cases for multiple language and country

maggarwal13:
As multi language support was already implemented, added integration test and junit fix for the same


whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -74,6 +76,9 @@
 
     private static final List<String> RDAP_CONFORMANCE_LEVEL = Lists.newArrayList("rdap_level_0");
 
+    public static final String NOICE_MULTIPLE_VALUE_DESC = "There are multiple %s %s in %s, but only the first %s %s was returned.";
+    public static final String NOTICE_MULTIPLE_VALUE_TITLE = "Multiple %s found";
```
eshryane: please add the word 'attributes' so it's more readable (i.e. multiple %s attributes found).

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -74,6 +76,9 @@
 
     private static final List<String> RDAP_CONFORMANCE_LEVEL = Lists.newArrayList("rdap_level_0");
 
+    public static final String NOICE_MULTIPLE_VALUE_DESC = "There are multiple %s %s in %s, but only the first %s %s was returned.";
```
eshryane: Add the word 'attributes' so 'There are multiple %s %s attributes in...'

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
eshryane: No need for a static field if the value is only used once? It's more readable if the text is where it's used

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -463,4 +465,38 @@ private static VCard createVCard(final RpslObject rpslObject) {
 
         return builder.build();
     }
+
+    private static void handleLanguageAttribute(RpslObject rpslObject, RdapObject rdapObject) {
```
eshryane: Please use final in method parameters and variables, so they can't be re-assigned.

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapObjectMapperTest.java
```diff
@@ -397,6 +402,38 @@ public void organisation() {
...
+                "abuse-c:        ABU-TEST\n" +
+                "mnt-by:         FRED-MNT\n" +
+                "source:         TEST"));
+
+        assertThat(result.getLang(), is("DK"));
+
+        final List<Notice> notices = result.getNotices();
+
+        assertThat(notices, hasSize(1));
+        assertNull(notices.get(0));
```
eshryane: Why is there a null notice?

maggarwal13: mapCommons method adds an empty notice apparently.
rdapResponse.getNotices().add(noticeFactory.generateTnC(requestUrl));
this line adds a null notice.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
sbusk: Also the constants should be private right?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -463,4 +462,40 @@ private static VCard createVCard(final RpslObject rpslObject) {
 
         return builder.build();
     }
+
+    private static void handleLanguageAttribute(final RpslObject rpslObject, final RdapObject rdapObject) {
```
sbusk: handleLanguageAttribute and handleCountryAttribute look very similar, can they be generalised?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -463,4 +462,40 @@ private static VCard createVCard(final RpslObject rpslObject) {
...
+        ip.setCountry(countries.get(0).getCleanValue().toString());
+        addNoticeForMultipleValues(ip, AttributeType.COUNTRY, countries, ip.getHandle());
+    }
+
+    private static void addNoticeForMultipleValues(final RdapObject rdapObject, final AttributeType type, final List<RpslAttribute> values, final String key) {
+        if(values.isEmpty() || values.size() == 1) {
+            return;
+        }
+
+        final String commaSeperatedValues = values.stream().map( x -> x.getCleanValue()).collect(Collectors.joining(","));
```
sbusk: Maybe use a somewhat more descriptive variable name rather than 'x' :)

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapObjectMapperTest.java
```diff
@@ -397,6 +402,38 @@ public void organisation() {
         assertThat(result.getPort43(), is("whois.ripe.net"));
 
         assertThat(result.getLang(), is("DK"));
+
+        final List<Notice> notices = result.getNotices();
+
+        MatcherAssert.assertThat(notices.get(0).getTitle(), is("Multiple language attributes found"));
```
sbusk: Use static imports, like in organisation_should_not_have_notice_for_language?

whois-api/src/test/java/net/ripe/db/whois/api/rdap/WhoisRdapServiceTestIntegration.java
```diff
@@ -315,6 +315,90 @@ public void lookup_inetnum_multiple_country_codes() {
         assertThat(ip.getName(), is("TEST-NET-NAME"));
         assertThat(ip.getLang(), is(nullValue()));
         assertThat(ip.getParentHandle(), is("IANA-BLK"));
+
+        final List<Notice> notices = ip.getNotices();
+        assertThat(notices, hasSize(4));
+        Collections.sort(notices);
+        assertThat(notices.get(0).getTitle(), is("Filtered"));
+        assertThat(notices.get(1).getTitle(), is("Multiple country attributes found"));
+        assertThat(notices.get(1).getDescription().get(0), is("There are multiple country attributes NL,DE in 192.0.0.0 - 192.255.255.255, but only the first country NL was returned."));
```
sbusk: Can I be a bit nitpicky and suggest we put spaces after a comma (NL, DE -> NL,DE)?

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapObjectMapperTest.java
eshryane: Can that be fixed? Strange that generateTnC returns null?

maggarwal13: In reality generateTnc will never return null, integrationTest always have TnC, Its Junit that we didn't mock the call to method generateTnc to return a notice. I will change the unit test case for it. I don't think mapCommons needs to have a null check for notice?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
eshryane: Looks good, thanks!

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapObjectMapperTest.java
eshryane: Ok thanks

PR #545, opened 2019-06-05
Leave the specification of an actual logging-impl to the user.

hanswesterbeek:
From what I see, this should just work. I got caught by this because most of our apps use logback, and then a log4j-dependency gets in the way.


eshryane:
@hanswesterbeek Will dependant projects have runtime problems (ClassNotFoundException) if they don't make sure to include these dependencies? Can we only depend on slf4j-api in whois-rpsl?

eshryane:
This change is included in the 1.94.1 release

PR #546, opened 2019-06-05
Refactored the latin1 conversion into seperate conversion and unsuppo…

sbusk:
…rted character substitution.


sbusk:
Won't be using this.

PR #557, opened 2019-06-25
Rdap return org contact 

maggarwal13:
response to include org handle with role as registrant. IP queries to include contact entities.

### DIFF
--- a/whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
@@ -98,6 +98,7 @@ class RdapObjectMapper {
         CONTACT_ATTRIBUTE_TO_ROLE_NAME.put(TECH_C, Role.TECHNICAL);
         CONTACT_ATTRIBUTE_TO_ROLE_NAME.put(MNT_BY, Role.REGISTRANT);
         CONTACT_ATTRIBUTE_TO_ROLE_NAME.put(ZONE_C, Role.ZONE);
+        CONTACT_ATTRIBUTE_TO_ROLE_NAME.put(ORG, Role.REGISTRANT); // TODO: [MA] both mnt_by and org have same role
     }
 
     private final NoticeFactory noticeFactory;
@@ -226,6 +227,8 @@ class RdapObjectMapper {
 
         handleLanguageAttribute(rpslObject, ip);
         handleCountryAttribute(rpslObject, ip);
+
+        ip.getEntitySearchResults().addAll(createContactEntities(rpslObject));
 
         return ip;
     }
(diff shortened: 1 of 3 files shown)


whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -98,6 +98,7 @@
         CONTACT_ATTRIBUTE_TO_ROLE_NAME.put(TECH_C, Role.TECHNICAL);
         CONTACT_ATTRIBUTE_TO_ROLE_NAME.put(MNT_BY, Role.REGISTRANT);
         CONTACT_ATTRIBUTE_TO_ROLE_NAME.put(ZONE_C, Role.ZONE);
+        CONTACT_ATTRIBUTE_TO_ROLE_NAME.put(ORG, Role.REGISTRANT); // TODO: [MA] both mnt_by and org have same role
```
sbusk: Should this TODO be resolved before we merge the PR?

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapObjectMapperTest.java
```diff
@@ -223,13 +238,16 @@ public void asBlock() {
...
-        assertThat(entities, hasSize(2));
+        assertThat(entities, hasSize(3));
+
         assertThat(entities.get(0).getHandle(), is("AP1-TEST"));
         assertThat(entities.get(0).getRoles(), containsInAnyOrder(Role.TECHNICAL, Role.ADMINISTRATIVE));
         assertThat(entities.get(0).getVCardArray(), is(nullValue()));
-        assertThat(entities.get(1).getHandle(), is("UPD-MNT"));
-        assertThat(entities.get(1).getRoles().get(0), is(Role.REGISTRANT));
-        assertThat(entities.get(1).getVCardArray(), is(nullValue()));
+        assertThat(entities.get(1).getHandle(), is("ORG-NCC1-RIPE"));
```
sbusk: Are the entities always in a deterministic order?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
eshryane: No, this is a known issue, documented in README.RDAP.md, we'll resolve it later.

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapObjectMapperTest.java
maggarwal13: Yes, We are using treemap RdapObjectMapper - line 307, although RFC does not enforce the natural order for entities

sbusk: I was thinking more about the tests running in a deterministic way as we've had some tests here and there that sometimes fail because of indexes being shuffled but in this case great! :)

PR #561, opened 2019-07-16
Delay authoritative resources import

sbusk:


sbusk:
No longer necessary with upcoming scheduler change.

PR #566, opened 2019-08-19
Renamed getNanoTime to getElapsedTime

eshryane:
To be clear it's not an absolute value, just an offset.

### DIFF
--- a/whois-commons/src/main/java/net/ripe/db/whois/common/DateTimeProvider.java
+++ b/whois-commons/src/main/java/net/ripe/db/whois/common/DateTimeProvider.java
@@ -13,8 +13,8 @@ public interface DateTimeProvider {
     LocalDateTime getCurrentDateTime();
     ZonedDateTime getCurrentDateTimeUtc();
 
-    /** returns System.nanoTime(), the high-res timer that counts 0 from JVM startup */
-    long getNanoTime();
+    /** returns elapsed time. N.B. not related to system time */
+    long getElapsedTime();
 
     /** specify the local timezone when creating a LocalDateTime from a timestamp */
     static LocalDateTime fromEpochMilli(final long timestamp) {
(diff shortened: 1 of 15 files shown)


eshryane:
Waiting for integration tests to finish

whois-commons/src/test/java/net/ripe/db/whois/common/TestDateTimeProvider.java
```diff
@@ -34,15 +34,11 @@ public ZonedDateTime getCurrentDateTimeUtc() {
         return getCurrentDateTime().atZone(ZoneOffset.UTC);
     }
 
-    public long getNanoTime() {
-        return DateTimeProvider.toEpochMilli(localDateTime);
+    public long getElapsedTime() {
+        return 100L;    // return a predicatable non-zero value that's testable
```
sbusk: predicatable -> predictable

whois-commons/src/main/java/net/ripe/db/whois/common/DateTimeProvider.java
```diff
@@ -13,8 +13,8 @@
     LocalDateTime getCurrentDateTime();
     ZonedDateTime getCurrentDateTimeUtc();
 
-    /** returns System.nanoTime(), the high-res timer that counts 0 from JVM startup */
-    long getNanoTime();
+    /** returns elapsed time. N.B. not related to system time */
+    long getElapsedTime();
```
sbusk: Maybe call it getElapsedTimeNanos() so it's clear that the unit is nanons and not milliseconds..?

eshryane: I'd prefer not to include the unit precision in the method name, it makes it very verbose. I updated the comment to be more specific.

eshryane:
Completed changes, waiting for integration tests to finish

eshryane:
Integration tests have all passed

eshryane:
If everyone is happy, I'd like to merge this change to include in Whois 1.95

PR #567, opened 2019-08-20
Don't log stack trace for certain "expected" exceptions.

eshryane:


eshryane:
I'd also like to merge this change to include in Whois 1.95

PR #569, opened 2019-08-28
Use a Timestamp class

eshryane:
To encapsulate a database timestamp, to avoid timestamp calculations around the code.
Also removes static "helper" methods.


eshryane:
Need to update VersionQueryExecutorTest to use UTC, as the Travis build runs in a different timezone

eshryane:
Integration tests have passed

whois-update/src/main/java/net/ripe/db/whois/update/keycert/PgpSignedMessage.java
```diff
@@ -176,7 +177,7 @@ public boolean verify(final PGPPublicKey publicKey) {
 
     // The signing time must be within an hour of the current time.
     public boolean verifySigningTime(final DateTimeProvider dateTimeProvider) {
-        final LocalDateTime signingTime = DateTimeProvider.fromDate(getPgpSignature().getCreationTime());
+        final LocalDateTime signingTime = DateUtil.fromDate(getPgpSignature().getCreationTime());
```
eshryane: Tested ok (sent signed message via syncupdates and mailupdates, updated object ok)

PR #570, opened 2019-09-18
WebApplicationException is expected in RDAP exception handler

eshryane:
Avoid logging WebApplicationException as "unexpected", as the RDAP code generates it in certain cases, e.g. if the ACL Limit is reached for the client (returns 429 status response).


eshryane:

eshryane:
Integration tests pass

eshryane:
Now return an RDAP-specific JSON object, rather than a generic HTML response body on 429 Too Many Requests.
Also include the query error message (i.e. the ACL blocked message).

PR #573, opened 2019-09-20
Abuse-c validation comment must refer to the LIR organisation

eshryane:
Abuse-c validation comment must refer to the LIR organisation, if the address is suspect.

### DIFF
--- a/whois-query/src/main/java/net/ripe/db/whois/query/dao/jdbc/JdbcAbuseValidationStatusDao.java
+++ b/whois-query/src/main/java/net/ripe/db/whois/query/dao/jdbc/JdbcAbuseValidationStatusDao.java
@@ -9,6 +9,7 @@ import org.springframework.dao.RecoverableDataAccessException;
 import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.stereotype.Repository;
 
+import javax.annotation.Nullable;
 import javax.sql.DataSource;
 
 @Repository
@@ -23,10 +24,10 @@ public class JdbcAbuseValidationStatusDao implements AbuseValidationStatusDao {
     }
 
     @Override
-    public boolean isSuspect(CIString email) {
-        return internalsTemplate.queryForObject(
-                "select count(*) from abuse_email where address = ? and status = ?",
-                new String[] { email.toLowerCase(), "SUSPECT" },
+    public boolean isSuspect(@Nullable final CIString address) {
+        return (address != null) && internalsTemplate.queryForObject(
+                "SELECT count(*) FROM abuse_email WHERE address = ? AND status = ?",
+                new Object[] { address.toString(), "SUSPECT" },
                 Integer.class) > 0;
     }
 }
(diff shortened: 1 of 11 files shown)


eshryane:

whois-query/src/main/java/net/ripe/db/whois/query/dao/jdbc/JdbcAbuseValidationStatusDao.java
```diff
@@ -23,10 +24,10 @@ public JdbcAbuseValidationStatusDao(@Qualifier("internalsDataSource") final Data
     }
 
     @Override
-    public boolean isSuspect(CIString email) {
-        return internalsTemplate.queryForObject(
-                "select count(*) from abuse_email where address = ? and status = ?",
-                new String[] { email.toLowerCase(), "SUSPECT" },
+    public boolean isSuspect(@Nullable final CIString address) {
+        return (address != null) && internalsTemplate.queryForObject(
```
sbusk: why the null check?

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/organisation/LirMntByAttributeCountValidator.java
```diff
@@ -37,7 +36,7 @@ public LirMntByAttributeCountValidator(final Maintainers maintainers) {
     @Override
     public void validate(final PreparedUpdate update, final UpdateContext updateContext) {
         final RpslObject originalObject = update.getReferenceObject();
-        if (!LIR.equals(originalObject.getValueForAttribute(AttributeType.ORG_TYPE))) {
+        if (! OrgType.LIR.getName().equals(originalObject.getValueForAttribute(AttributeType.ORG_TYPE))) {
```
sbusk: LIR == OrgType.getFor(originalObject.getValueForAttribute(AttributeType.ORG_TYPE)) is a bit cleaner..?

whois-query/src/main/java/net/ripe/db/whois/query/dao/jdbc/JdbcAbuseValidationStatusDao.java
eshryane: I added more input checks, but it's not necessary in this case, I removed it again

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/organisation/LirMntByAttributeCountValidator.java
eshryane: Agreed, refactored in all validators to be consistent.

eshryane:
Integration tests have all passed

whois-query/src/main/java/net/ripe/db/whois/query/planner/AbuseCFinder.java
```diff
@@ -46,61 +47,97 @@ public AbuseCFinder(final RpslObjectDao objectDao,
...
         }
 
-        final RpslObject responsibleOrg = findResponsibleOrgObject(object);
-        return responsibleOrg != null? responsibleOrg.getValueForAttribute(AttributeType.ORG) : null;
+        return findResponsibleOrgReference(rpslObject);
     }
 
     @Nullable
-    private RpslObject findResponsibleOrgObject(final RpslObject rpslObject) {
+    private CIString findResponsibleOrgReference(final RpslObject rpslObject) {
```
sbusk: findResponsibleOrgReference() and findResponsibleLirOrgReference() are practically identical apart from findResponsibleLirOrgReference() checking the org is an lir, maybe have just one method which also takes on the org type we're looking for as an argument so we don't duplicate the logic..

eshryane: Agreed, I refactored the methods

eshryane:
Integration tests have all passed

PR #581, opened 2019-10-18
Use slave datasource for query

maggarwal13:


whois-api/src/main/java/net/ripe/db/whois/api/rdap/WhoisRdapService.java
```diff
@@ -39,6 +39,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
```
eshryane: Replace ```import javax.ws.rs.*;``` with individual imports, also commented import

PR #584, opened 2019-10-23
Update bouncycastle dependency to latest

eshryane:
Also replaced deprecated X509CertParser.

(Reverted changes affecting "owner" attribute) ~~Be aware the "owner" attribute in the replacement X509Certificate has a different format. This implies the generated owner attribute in X509 key-cert objects will change when the object is updated. Need to investigate whether this is really an issue (as most X509 key-cert objects in the DB are expired, they cannot be used for authentication).~~


eshryane:
Failing integration tests:
* AuthSpec
* SignedMessageIntegrationSpec

eshryane:
Ran script to verify *all* signed messages in Whois updates for this year so far, with the old bouncycastle dependency (1.60) and this branch (1.64), and there were no differences.

eshryane:
Related: https://github.com/bcgit/bc-java/issues/636

eshryane:
Bouncy Castle 1.65 is now available, but not yet available in Maven.

PR #598, opened 2020-01-21
Include build timestamp and commit id in the Whois lookup and search responses.

eshryane:


eshryane:
Waiting for integration tests to pass.

eshryane:
Integration tests are passing. Added some tests, ready to merge.

eshryane:
Whois search and lookup responses will get a new "version" element, which looks like in XML:
```
<version version="1.97-SNAPSHOT" timestamp="2020-01-22T18:22:13Z" commit-id="0e58632"/>
```
and in JSON:
```
"version" : {
  "version" : "1.97-SNAPSHOT",
  "timestamp" : "2020-01-22T18:22:13Z",
  "commit-id" : "0e58632"
}
```

The version is also added to the full text search response, but it must be compatible with the Solr client format:

```
 {
    "lst" : {
      "name" : "version",
      "strs" : [ {
        "str" : {
          "name" : "version",
          "value" : "1.97-SNAPSHOT"
        }
      }, {
        "str" : {
          "name" : "timestamp",
          "value" : "2020-01-22T18:22:13Z"
        }
      }, {
        "str" : {
          "name" : "commit_id",
          "value" : "0e58632"
        }
      } ]
    }
```




N.B. This version information is for the *application* release and not the *API* (which isn't versioned at all), hopefully this is not confusing for consumers.

PR #599, opened 2020-01-28
abusec contact  only for ripe sources

maggarwal13:


whois-query/src/main/java/net/ripe/db/whois/query/planner/AbuseCFinder.java
```diff
@@ -1,5 +1,6 @@
 package net.ripe.db.whois.query.planner;
 
+import com.google.common.collect.ImmutableSet;
```
eshryane: I suggest using the built-in wrapper java.util.Collections.unmodifiableSet(<Set>) instead of guava

whois-query/src/main/java/net/ripe/db/whois/query/planner/AbuseCFinder.java
```diff
@@ -46,6 +54,26 @@ public AbuseCFinder(@Qualifier("jdbcRpslObjectSlaveDao") final RpslObjectDao obj
         this.ipv6Tree = ipv6Tree;
         this.maintainers = maintainers;
         this.abuseValidationStatusDao = abuseValidationStatusDao;
+        this.ripeSources = getRipeSources(mainSource, nonAuthSource, grsSource);
+    }
+
+    private ImmutableSet<CIString> getRipeSources(final String mainSource, final String nonAuthSource, final String grsSource) {
```
eshryane: Return generic Set<> in method signature (the immutable property is a detail)

whois-query/src/main/java/net/ripe/db/whois/query/planner/AbuseCFinder.java
```diff
@@ -46,6 +54,26 @@ public AbuseCFinder(@Qualifier("jdbcRpslObjectSlaveDao") final RpslObjectDao obj
...
+        ImmutableSet.Builder<CIString> sourceBuilder  = new ImmutableSet.Builder<>();
+
+        if(StringUtils.isNotEmpty(mainSource)) {
```
eshryane: whois.source is mandatory (although it could be empty it's unlikely!), no need for a check.

whois-query/src/main/java/net/ripe/db/whois/query/planner/AbuseCFinder.java
```diff
@@ -46,6 +54,26 @@ public AbuseCFinder(@Qualifier("jdbcRpslObjectSlaveDao") final RpslObjectDao obj
...
+            sourceBuilder.add(CIString.ciString(mainSource));
+        }
+
+        if(StringUtils.isNotEmpty(nonAuthSource)) {
+            sourceBuilder.add(CIString.ciString(nonAuthSource));
+        }
+
+        final String matchingGrs = mainSource+ "-GRS";
```
eshryane: I'd just automatically add a GRS source as mainsource + "GRS", no need to check if one is configured

whois-query/src/main/java/net/ripe/db/whois/query/planner/AbuseCFinder.java
```diff
@@ -34,9 +38,13 @@
     private final Ipv6Tree ipv6Tree;
     private final Maintainers maintainers;
     private final AbuseValidationStatusDao abuseValidationStatusDao;
+    private final ImmutableSet<CIString> ripeSources;
```
eshryane: I'd call this property "mainSources" as it's not necessarily called "ripe"

whois-query/src/main/java/net/ripe/db/whois/query/planner/AbuseCFinder.java
eshryane: Also rename method to "getMainSources..."

whois-query/src/main/java/net/ripe/db/whois/query/planner/AbuseCFinder.java
```diff
@@ -106,6 +134,10 @@ private boolean isLir(@Nullable final RpslObject rpslObject) {
     @CheckForNull
     @Nullable
     public RpslObject getAbuseContactRole(final RpslObject rpslObject) {
+        if(!ripeSources.contains(rpslObject.getValueForAttribute(AttributeType.SOURCE))) {
```
eshryane: Good pattern (return early)!

whois-query/src/test/java/net/ripe/db/whois/query/integration/AbuseCTestIntegration.java
```diff
@@ -239,4 +240,66 @@ public void brief_query_shows_abusemailbox_twice() {
         final String briefResponse = TelnetWhoisClient.queryLocalhost(QueryServer.port, "-b 193.0.0.0");
         assertThat(briefResponse, not(containsString("notshown@abuse.net")));
     }
+    @Test
+    public void should_not_lookup_abuseC_if_rpsl_source_do_not_match() {
+
+        databaseHelper.updateObject(RpslObject.parse( "inetnum:       173.0.0.0 - 173.255.255.255\n" +
```
eshryane: Don't make reference to "RIPE" as the source, but something like 'NON-TEST' or 'INVALID'

whois-query/src/test/java/net/ripe/db/whois/query/planner/AbuseCFinderTest.java
```diff
@@ -1,6 +1,7 @@
 package net.ripe.db.whois.query.planner;
 
 
+import com.google.common.collect.ImmutableSet;
```
eshryane: Use Set

whois-query/src/main/java/net/ripe/db/whois/query/planner/AbuseCFinder.java
maggarwal13: Collections.unmodifiableSet() still can be changed as its a view that's why i usually prefer immutable set

eshryane: A view is fine, if you don't have any other references to the actual set (i.e. wrap creating the Set in a method, and return the view). We shouldn't have multiple ways to do the same thing in the code. But here it's not a big deal.

whois-query/src/test/java/net/ripe/db/whois/query/planner/AbuseCFinderTest.java
eshryane: As above, not a big deal.

PR #601, opened 2020-02-03
Restrict maximum size of object.

sbusk:


whois-api/src/main/java/net/ripe/db/whois/api/UpdatesParser.java
```diff
@@ -179,7 +182,12 @@ private String extractOverride(final Set<Credential> credentials, final String p
             }
 
             try {
-                updates.add(createUpdate(paragraph, operation, deleteReasons, content, updateContext));
+                if (content.length() > MAXIMUM_OBJECT_SIZE) {
```
eshryane: minor suggestion to move length check outside unrelated try-catch, and put in a continue statement instead.

PR #602, opened 2020-02-10
One JVM Fork per Core to improve test speed

eshryane:
Conservative improvement to improve the speed of the integration test build: switch from hard-coded "3" forks, to one fork per core. This takes advantage of more cores if available, but also won't over-commit if there are less.

Tested this change produces ~10% improvement (5 minutes less) in build times on my own laptop and on CI server.

Don't vary thread count (also 1 per core), or parallel setting, run tests one per fork to avoid threading issues (failing tests).


eshryane:
Merge gitlab branch to master first

PR #607, opened 2020-03-04
Do not allow inconsistent org-name: formatting on create.

eshryane:
Previously fixed for MODIFY in #475.

Now also checked during CREATE, and added tests.


eshryane:
> I would rather add a test case in WhoisRestServiceTestIntegration for creating, I usually do not add more tests in groovy file. Not sure if that's the right way to do or not



> I would rather add a test case in WhoisRestServiceTestIntegration for creating, I usually do not add more tests in groovy file. Not sure if that's the right way to do or not

For me, the Groovy tests are for functional (business) testing (for 'if/when/then' scenarios), across the whole application, and the Java Integration tests are usually for a single service.

In this case, I don't have a single service to test against, so the existing Groovy test seemed a better fit for me.

PR #609, opened 2020-03-12
Added rate limiting filter.

sbusk:


whois-api/src/main/java/net/ripe/db/whois/api/httpserver/JettyBootstrap.java
```diff
@@ -94,12 +109,29 @@ private Server createAndStartServer(final int port) {
         }
     }
 
+    private FilterHolder createDosFilter() throws JmxException, JMException {
```
eshryane: Maybe add a comment linking to https://www.eclipse.org/jetty/documentation/current/dos-filter.html that explains the options?

sbusk: Good suggestion, will add.

PR #596, opened 2020-01-08
Nrtm access wg

maggarwal13:

PR #612, opened 2020-03-16
Nrtm acl limit

maggarwal13:
ACL Limit in NRTM Service

### DIFF
--- a/whois-nrtm/src/main/java/net/ripe/db/whois/nrtm/NrtmConnectionPerIpLimitHandler.java
+++ b/whois-nrtm/src/main/java/net/ripe/db/whois/nrtm/NrtmConnectionPerIpLimitHandler.java
@@ -2,14 +2,14 @@ package net.ripe.db.whois.nrtm;
 
 import net.ripe.db.whois.common.pipeline.ChannelUtil;
 import net.ripe.db.whois.common.pipeline.ConnectionCounter;
+import net.ripe.db.whois.query.QueryMessages;
+import net.ripe.db.whois.query.acl.AccessControlListManager;
 import org.jboss.netty.channel.Channel;
 import org.jboss.netty.channel.ChannelFutureListener;
 import org.jboss.netty.channel.ChannelHandler;
 import org.jboss.netty.channel.ChannelHandlerContext;
 import org.jboss.netty.channel.ChannelStateEvent;
 import org.jboss.netty.channel.SimpleChannelUpstreamHandler;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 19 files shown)


whois-nrtm/src/test/java/net/ripe/db/whois/nrtm/NrtmConnectionPerIpLimitHandlerTest.java
```diff
@@ -11,14 +13,14 @@
...
+import org.mockito.Matchers;
 import org.mockito.Mock;
 import org.mockito.runners.MockitoJUnitRunner;
 
 import java.net.Inet4Address;
 import java.net.InetSocketAddress;
 
-import static org.mockito.Matchers.anyObject;
-import static org.mockito.Matchers.argThat;
+import static org.mockito.Matchers.*;
```
eshryane: Please fix your IDE settings not to use "star" for imports, it's harder to diff changes

whois-nrtm/src/main/java/net/ripe/db/whois/nrtm/NrtmConnectionPerIpLimitHandler.java
```diff
@@ -70,4 +72,23 @@ private boolean connectionsExceeded(final InetAddress remoteAddresss) {
         final Integer count = connectionCounter.increment(remoteAddresss);
         return (count != null && count > maxConnectionsPerIp);
     }
+
+    private boolean canOpenConnection(final Channel channel, final InetAddress remoteAddress) {
```
eshryane: This method is misleading, the name suggests there are no side-effects, but then it can close the connection. The method does multiple things, and it's only called once - I think it's clearer to inline it into the channelOpen method()?

whois-nrtm/src/main/java/net/ripe/db/whois/nrtm/NrtmConnectionPerIpLimitHandler.java
```diff
@@ -23,29 +23,31 @@
 @ChannelHandler.Sharable
 public class NrtmConnectionPerIpLimitHandler extends SimpleChannelUpstreamHandler {
```
eshryane: Can you move the ACL logic into a separate class? Look at the #596 PR for an example (there's a separate class for AccessControlList).  It's confusing that this Handler now also includes the ACL logic, it should be separate.

maggarwal13: Earlier I thought of moving it to a different class but this class name is connection limit perIp handler and acl also works on per Ip limit so I put the logic in same class, but I not a big deal to put it in a seperate class if it makes it more readable

eshryane: Great, thanks!

whois-nrtm/src/test/java/net/ripe/db/whois/nrtm/integration/NrtmClientACLLimitTestIntegration.java
```diff
@@ -0,0 +1,103 @@
...
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.test.annotation.DirtiesContext;
+
+import java.net.InetAddress;
+
+@Category(IntegrationTest.class)
```
eshryane: Can you add one "success" test, where the client *isn't* blocked or denied by ACL?

maggarwal13: Have added a success case in junit. 

Actually this integration test does not make sense here. As we are just waiting for 5 seconds and checking the nrtm client thread count > 0, (copied from NrtmClientInvalidHostTestIntegration). We should be deleting this integration test and same for NrtmClientInvalidHostTestIntegration or find a different way to check if channel is open or not.  

Infact NrtmClientTestIntegration also does not do much or  am I missing something?

maggarwal13: for NrtmClientTestIntegration  we inject different jdbctemplate that is fine then

eshryane: Thanks for fixing the tests

PR #616, opened 2020-03-23
Disable DoS filter by default, conditionally (de)register JMX bean.

sbusk:


whois-api/src/main/java/net/ripe/db/whois/api/httpserver/JettyBootstrap.java
```diff
@@ -47,16 +47,20 @@
 
     private final String trustedIpRanges;
 
+    private boolean dosFilterEnabled;
```
eshryane: Use final? Or will it be possible to enable later via JMX?

whois-api/src/test/java/net/ripe/db/whois/api/httpserver/JettyBootstrapTest.java
```diff
@@ -2,16 +2,12 @@
...
-
 import static org.junit.Assert.assertTrue;
 
 public class JettyBootstrapTest {
 
     @Test
-    public void testWhitelist() throws MalformedObjectNameException {
-        JettyBootstrap jbs = new JettyBootstrap(null, null, null, "127.0.0.1,::1,193.0.0.0 - 193.0.23.255,2001:67c:2e8::/48,10.0.0.0 - 10.255.255.255");
-
+    public void testWhitelist() {
```
eshryane: "test" is redundant in the method name?

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/JettyBootstrap.java
sbusk: Hmm yes it is but this is the value at startup.
Will change it to final.

PR #618, opened 2020-03-25
add display name in crowd dummy

maggarwal13:


whois-api/src/test/java/net/ripe/db/whois/api/CrowdServerDummy.java
```diff
@@ -128,13 +130,15 @@ private String getUuid(final String username) {
             return String.format("<attributes><attribute name=\"uuid\"><values><value>%s</value></values></attribute></attributes>", username);
         }
 
-        private String getUsername(final String uuid) {
+        private String getUsername(final String uuid, final String displayName) {
```
eshryane: Maybe rename to serialiseUser ? Odd that it's a "getter" and named "username".

maggarwal13: then I have to rename the getUuuid as well to serialize Uuid?

eshryane: Yes, and also getUserSessionResponse(). Those methods are serialising some data into XML, it's not a getter.

PR #630, opened 2020-04-30
Update log4j slf4j

sbusk:


whois-api/src/test/java/net/ripe/db/whois/api/log/JettyRequestLogTestIntegration.java
```diff
@@ -43,7 +49,7 @@
             "mnt-by:        OWNER-MNT\n" +
             "source:        TEST");
 
-    @Value("var${jvmId:}/log/jetty")
+    @Value("target/log/jetty")
```
eshryane, line 52: The jvmId is important for separating logging for tests running in parallel, can you re-add it?

PR #638, opened 2020-06-11
ignore case in query

maggarwal13:


whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapQueryHandler.java
```diff
@@ -63,8 +64,7 @@ public void handle(final ResponseObject responseObject) {
                 }
 
                 private void addIfPrimaryObject(final RpslObject responseObject) {
-                    final String primaryKey = responseObject.getKey().toString();
-                    if(primaryKey.equals(query.getSearchValue())) {
+                    if(responseObject.getKey().equals(ciString(query.getSearchValue()))) {
```
eshryane, line 67: You can use responseObject.getKey().equals(query.getSearchValue()) (drop the static ciString)

maggarwal13: I have to add equals ignore case then, as cistring as our own implementation of equals, wont it be best if we use that?

eshryane: responseObject.getKey() is already a CIString, and it overrides equals() to perform an insensitive match on a String. So it's unnecessary to also convert the searchValue to a CIString.

maggarwal13: Ah, I missed the point, Thanks :)

PR #640, opened 2020-06-16
Full Text Search thread pool and timeout

eshryane:
Limit concurrent Full Text searches using an executor pool and replace maxConcurrentSearches.


maggarwal13:
I do not see the benefit, also we are moving to ES there is already a pull request for that. So we can analyse the performance and refactor accordingly there.

PR #641, opened 2020-06-23
Reload legacy autnums every 15 minutes.

sbusk:


whois-scheduler/src/main/java/net/ripe/db/whois/scheduler/task/autnum/LegacyAutnumReloadTask.java
```diff
@@ -21,7 +21,7 @@ public LegacyAutnumReloadTask(final LegacyAutnum legacyAutnum) {
     }
 
     @Override
-    @Scheduled(cron = "0 0 0 * * *")
+    @Scheduled(cron = "0 1/15 * * * *")
     @SchedulerLock(name = "ReloadLegacyAutnums")
```
eshryane, line 25: Is this a global lock? Do we still need it?

sbusk: Yes it's the global lock and you're right we don't need it. I've removed it as the task needs to be executed on all nodes every 15 minutes.

PR #642, opened 2020-07-01
Disallow weak key lengths and hash algorithms in PGP and X509 keycerts

eshryane:


whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/keycert/PgpKeycertValidator.java
```diff
@@ -0,0 +1,95 @@
...
+                }
+                break;
+            case PublicKeyAlgorithmTags.RSA_GENERAL:
+                if (wrapper.getPublicKey().getBitStrength() < MINIMUM_KEY_LENGTH_RSA) {
+                    updateContext.addMessage(update, UpdateMessages.publicKeyLengthIsWeak("RSA", MINIMUM_KEY_LENGTH_RSA, wrapper.getPublicKey().getBitStrength()));
+                }
+                break;
+            default:
+                // skip key length check until we are sure about an appropriate minimum length for that algorithm
+                updateContext.log(new Message(Messages.Type.INFO, "Skipping public key length check for algorithm %d", wrapper.getPublicKey().getAlgorithm()));
```
sbusk, line 77: Wouldn't this be a bit confusing? You perform an update and get an info message saying 'skipping length check'? This would happen for EC keys for example. 
Better to not add any info messages (we don't at the moment) imo.

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/keycert/X509KeycertValidator.java
```diff
@@ -0,0 +1,101 @@
...
+
+@Component
+public class X509KeycertValidator implements BusinessRuleValidator {
+
+    private static final ImmutableList<Action> ACTIONS = ImmutableList.of(Action.CREATE);
+    private static final ImmutableList<ObjectType> TYPES = ImmutableList.of(ObjectType.KEY_CERT);
+
+    private static final CIString METHOD_X509 = CIString.ciString("X509");
+
+    private static final Set<String> WEAK_HASH_ALGORITHMS = Sets.newHashSet("SHA1withRSA", "SHA1withDSA", "MD5withRSA", "MD5withDSA");
```
sbusk, line 35: So md2WithRSAEncryption (see https://www.ietf.org/rfc/rfc3279.txt) would be ok?
I think it's better to whitelist here or even check just for hash algorithm presence (so should be SHA256, SHA384, SHA512) and make those configurable.

eshryane:
52 failing integration tests on this branch (we use a lot of weak keys in our test data). This will take some time to resolve.

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/keycert/PgpKeycertValidator.java
eshryane: These messages go into the audit log, not to the user. I wanted to make sure we log any exceptions or unhandled ("default") cases, without spamming the console log, so we have metrics to make improvements later. For now, I'm testing with the existing key-cert objects from the split file.

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/keycert/X509KeycertValidator.java
eshryane: I prefer a blocklist for known weak hash algorithms, rather than blocking new (more secure) algorithms by default.

I generated this list using the known keys in the key-cert split file, I can always add more.

We could indeed extract this to a property (with a default value), so we can keep it updated without needing a release, but the list changes very rarely.

I'm doing a string comparison because java.security.cert.X509Certificate seems to only return a string (either getSigAlgName() or getSigAlgOID()). There doesn't appear to be an integer value for the X509 hash algorithm, like in org.bouncycastle.bcpg.PublicKeyAlgorithmTags which I can use for PGP keys.

PR #643, opened 2020-07-07
NWI-11 Punycode Internationalized Domain Names in email addresses.

sbusk:
This is the NWI-11 Implementation, which automatically converts international (i.e. IDN or non-ASCII) domain names in email addresses into Punycode (since the RIPE database uses Latin-1).

Problem Definition:
https://www.ripe.net/ripe/mail/archives/db-wg/2020-June/006524.html

Solution Definition:
https://www.ripe.net/ripe/mail/archives/db-wg/2020-June/006525.html

Implementation Plan:
https://www.ripe.net/ripe/mail/archives/db-wg/2020-July/006595.html

### DIFF
--- a/whois-client/src/main/java/net/ripe/db/whois/api/syncupdate/SyncUpdateBuilder.java
+++ b/whois-client/src/main/java/net/ripe/db/whois/api/syncupdate/SyncUpdateBuilder.java
@@ -32,6 +32,7 @@ public class SyncUpdateBuilder {
     private boolean diff;
     private boolean aNew;
     private boolean redirect;
+    private String charset;
 
     public SyncUpdateBuilder setHeaders(final MultivaluedMap<String, String> headers) {
         this.headers = headers;
@@ -83,23 +84,20 @@ public class SyncUpdateBuilder {
         return this;
     }
 
+    public SyncUpdateBuilder setCharset(final String charset) {
+        this.charset = charset;
+        return this;
+    }
+
     public Client build() {
-        if (url != null) {
-            return new Client(url, headers, data, help, diff, aNew, redirect);
-        }
-
-        if (host != null && port != null && source != null) {
-            return new Client(host, port, source, headers, data, help, diff, aNew, redirect);
-        }
-
-        throw new IllegalStateException("Either (host, port, source) or (url) should not be null");
+        return new Client(url, host, port, source, headers, data, help, diff, aNew, redirect, charset);
     }
 
     public static class Client {
 
         private static final Joiner.MapJoiner PARAM_JOINER = Joiner.on('&').withKeyValueSeparator("=");
         private static final Pattern CHARSET_PATTERN = Pattern.compile(".*;charset=(.*)");
-        private static final String CHARSET = "ISO-8859-1";
+        private static final String DEFAULT_CHARSET = "ISO-8859-1";
 
         private final URL url;
         private final MultivaluedMap<String, String> headers;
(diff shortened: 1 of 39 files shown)


whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/update/RoleSpec.groovy
```diff
@@ -671,6 +671,43 @@ class RoleSpec extends BaseQueryUpdateSpec {
         query_object_not_matches("-rBT role FR1-TEST", "role", "First Role", "tech-c:")
     }
 
+    def "punycode abuse-mailbox with IDN"() {
```
eshryane, line 674: Need an additional test to confirm local-part with a non-ASCII value is *not* converted to Punycode but is rejected with a syntax error.

eshryane: Confirmed that IDNEmailAddressConverter only converts the domain part, not the local part 👍

eshryane:
Non-latin1 characters are substituted and not converted to Punycode properly (added unit test)

eshryane:
Updated PR to handle non-Latin-1 domain names

sbusk:
Changes look pretty clean. I would add a test for a syntactically incorrect rpsl object with a idn email attribute though (should the punycode conversion then have taken place or only the syntax error reported or both?).

whois-client/src/main/java/net/ripe/db/whois/api/syncupdate/SyncUpdateBuilder.java
```diff
@@ -83,23 +84,20 @@ public SyncUpdateBuilder setRedirect(final boolean redirect) {
...
+    public Client build() {
+        return new Client(url, host, port, source, headers, data, help, diff, aNew, redirect, charset);
     }
 
     public static class Client {
 
         private static final Joiner.MapJoiner PARAM_JOINER = Joiner.on('&').withKeyValueSeparator("=");
         private static final Pattern CHARSET_PATTERN = Pattern.compile(".*;charset=(.*)");
-        private static final String CHARSET = "ISO-8859-1";
+        private static final String DEFAULT_CHARSET = "ISO-8859-1";
```
dadepo, line 100: Minor: Maybe consider using `StandardCharsets.ISO_8859_1` instead of hardcoded string

dadepo (approved):

whois-client/src/main/java/net/ripe/db/whois/api/syncupdate/SyncUpdateBuilder.java
eshryane: Done!

eshryane:
> object with a idn email attribute though (should the punycode conversion then have taken place or only the syntax error repor

You added a test in RoleSpec that tests an umlaut in the local part and domain part of an RPSL object, I think that works pretty well. The domain part is converted into Punycode, but the local part is not, and then that produces a syntax error.

eshryane:

eshryane:
> > object with a idn email attribute though (should the punycode conversion then have taken place or only the syntax error repor
> 
> You added a test in RoleSpec that tests an umlaut in the local part and domain part of an RPSL object, I think that works pretty well. The domain part is converted into Punycode, but the local part is not, and then that produces a syntax error.

Added test for "punycode abuse-mailbox but object is syntactically incorrect"

PR #644, opened 2020-07-07
Handle object deleted before full text search index updated

eshryane:
An object may be deleted from the RIPE database and a full text search matches it before its been removed from the index.

Handle this by catching the DB exception, and decrementing the result size (i.e. skip the non-existant match).


eshryane:
Integration tests passed

PR #650, opened 2020-08-05
rest service allow options

maggarwal13:


eshryane:
Do we need to allow the OPTIONS method? It's just used for the preflight check itself, and it's not supported by the REST API.

maggarwal13:
You are right. I will delete this branch, its not needed

PR #652, opened 2020-08-11
Prevent race condition when updating authoritative resources

eshryane:
Wrap authoritative_resource update in a single transaction.

Although the ResourceDataDao class has a Transactional annotation, in the store() method the DELETE statement was written to disk before the INSERT statement executed (confirmed with manual testing).

Wrapping the two statements in a transactionTemplate causes the DELETE and INSERT to be executed together, preventing the race condition (again, confirmed with manual testing, by adding a sleep between the two and querying the database).


eshryane:
Suggestions from team:
* Review the difference between the Transactional annotation and transactiontemplate
* If refresh doesn't find anything in the table, throw an exception
* Log a diff on import (so we can check changes)
* Multiple DAOs on authoritative_resource table

PR #658, opened 2020-09-23
Query and Update log parsers

eshryane:


whois-commons/src/test/java/net/ripe/db/whois/common/support/QueryLogEntryTest.java
```diff
@@ -0,0 +1,44 @@
...
+import org.junit.Test;
+
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+
+public class QueryLogEntryTest {
+
+    @Test
+    public void negative_channel_id() {
+        QueryLogEntry.parse("20180401 00:14:37 -215685797 PW-QRY-INFO <0+0+0>  3ms [127.0.0.1] --  WWW.COM.EU");
```
dadepo, line 12: This test does not have any assertion...is this intended or an omission?

dadepo (approved):
Looks 👍🏼

whois-commons/src/test/java/net/ripe/db/whois/common/support/QueryLogEntryTest.java
eshryane: It's intentional, the parser shouldn't throw an exception on a negative channel id (I annotated the method accordingly). This field is skipped so I can't check the value after parsing.

Issue #307, opened 2015-08-31
RDAP missing CORS header


codebje:
The RDAP server running at rdap.db.apnic.net is not adding Access-Control-Allow-Origin headers, against the recommendation of RFC 7480 s5.6.  This means it can't be queried by scripts running in web pages.

eshryane:
Hey Byron,

we recently added a CrossOriginFilter class, which currently only accepts .ripe.net origins:

https://github.com/RIPE-NCC/whois/blob/master/whois-api/src/main/java/net/ripe/db/whois/api/rest/CrossOriginFilter.java

This could be modified, and added to:

https://github.com/RIPE-NCC/whois/blob/rdap/whois-api/src/main/java/net/ripe/db/whois/api/whois/rdap/WhoisRdapServletDeployer.java

codebje:
Hey Ed,

I'm doing ASN lookups from a web page, and currently it works for ASNs in all regions except RIPE, because RIPE's service doesn't have an appropriate CORS header.  I'm not sure you'd call it a bug with the code, so much as an issue with the deployment or operation of RIPE's RDAP service - so perhaps github isn't the best place for this?

eshryane:
Thanks, I've committed a TODO on the rdap branch (dc1a864f55cba4ef6c2f3a93b922a405746b8400). We plan to work further on our rdap implementation soon, I think this issue can remain open until we address it.

eshryane:
Created pull request with a fix #412

eshryane:
Merged into rdap branch, and deployed to https://rdap.db.ripe.net.

eshryane:
Thanks Byron, I've re-opened the issue.

eshryane:
Until this is fixed in Whois, I made a temporary proxy change to include "Access-Control-Allow-Origin: *" in RDAP responses (if Origin is not specified in the request).

eshryane:
Confirmed fixed.

PR #661, opened 2020-10-06
Allow *any* CORS origin on RDAP GET requests by default

eshryane:
As recommended by RFC7480:

   "When responding to queries, it is RECOMMENDED that servers use the
    Access-Control-Allow-Origin header field, as specified by
    [W3C.REC-cors-20140116].  A value of "*" is suitable when RDAP is
    used for public resources."

Ref. https://tools.ietf.org/html/rfc7480#section-5.6

### DIFF
--- a/whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapCrossOriginFilter.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapCrossOriginFilter.java
@@ -1,0 +1,40 @@
+package net.ripe.db.whois.api.rdap;
+
+import com.google.common.net.HttpHeaders;
+import org.eclipse.jetty.servlets.CrossOriginFilter;
+
+import javax.servlet.FilterChain;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.ws.rs.HttpMethod;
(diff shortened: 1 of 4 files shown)


whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapCrossOriginFilter.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Override
+    public void doFilter(final ServletRequest request, final ServletResponse response, final FilterChain chain) throws IOException, ServletException
+    {
+        handle((HttpServletRequest)request, (HttpServletResponse)response);
+        super.doFilter(request, response, chain);
+    }
+
+    private void handle(final HttpServletRequest request, final HttpServletResponse response) {
+        if ((request.getHeader(HttpHeaders.ORIGIN) == null) && request.getMethod().equals(HttpMethod.GET)) {
```
dadepo, line 36: question: why is the requirement for the ORIGIN to be null placed? I do not see anything that cann point to this in the rfc...mind clarifying?

eshryane: The RFC recommends that the CORS header *always* appears in the GET response. If there is an Origin header, then the response is handled by the existing CrossOrigin filter. This PR adds a CORS header if there *isn't* an Origin header in the request.

eshryane:
Resolves #307 (we've implemented a workaround in the HTTP proxy but this code fix will make that redundant)

dadepo (approved):

PR #662, opened 2020-10-14
Whois fix "--diff-versions" parsing

dadepo:


whois-query/src/main/java/net/ripe/db/whois/query/query/Query.java
```diff
@@ -225,15 +225,19 @@ public int getObjectVersion() {
...
-            }
-            if (secondValue == firstValue) {
-                throw new QueryException(QueryCompletionInfo.PARAMETER_ERROR, QueryMessages.malformedQuery("diff versions are the same"));
+            try {
+                final int firstValue = Integer.parseInt(values[0]);
+                final int secondValue = Integer.parseInt(values[1]);
+                if (firstValue < 1 || secondValue < 1) {
+                    throw new QueryException(QueryCompletionInfo.PARAMETER_ERROR, QueryMessages.malformedQuery("diff version number must be greater than 0"));
+                }
+                if (secondValue == firstValue) {
```
eshryane, line 234: Can you also test that the second value cannot be smaller (or equal to) the first value? It doesn't make sense to allow that.

dadepo: I thought about that but I held back because:

1. Judging from the tests, this seems to be the intended behaviour (at least as at 2013 when it was implemented). If you look at:
```def "--diff-versions (forward range)"()``` and ```def "--diff-versions (inverse range)"()``` they test that both 2:1 and 1:2 works.  

2. Making such change would also be a breaking change. Since one cant really consider the current behaviour a _bug_, I hesitate to turn it into a breaking change

Do let me know if this changes your thoughts or not

eshryane: Fair enough, it's been working as-is for long enough not to introduce a breaking change now. I'm fine to leave it out

sbusk (approved):

eshryane (approved):

PR #664, opened 2020-10-26
filter sensitive information in response body on exception

dadepo:


whois-api/src/main/java/net/ripe/db/whois/api/rest/RestServiceHelper.java
```diff
@@ -65,14 +67,18 @@ private static String filter(final String queryString) {
         final StringBuilder builder = new StringBuilder();
         char separator = '?';
 
-        for (String next : AMPERSAND_SPLITTER.split(queryString)) {
-            final Iterator<String> iterator = EQUALS_SPLITTER.split(next).iterator();
-            if (iterator.hasNext() && iterator.next().equalsIgnoreCase("password")) {
-                continue;
-            }
+        if (queryString.contains(OVERRIDE_STRING)) {
```
dadepo, line 70: Thought this morning: consider moving  this logic into `PasswordFilter.filterPasswordsInUrl`. Before merging see if that won't be a better way to have it

eshryane: 👍

eshryane:
Try using`javax.ws.rs.core.UriBuilder` to filter the URL and remove the custom parsing code altogether.

```
    @Test
    public void test() {
        final UriBuilder builder;
        try {
            builder = UriBuilder.fromUri("https://rest.db.ripe.net/ripe/mntner/SHRYANE-MNT?password=test&override=test,password,reason&param=value&password=another");
        } catch (IllegalArgumentException e) {
            // invalid URI, skip filtering
            return;
        }

        builder.replaceQueryParam("password", "FILTERED");
        builder.replaceQueryParam("override", "FILTERED");

        assertThat(builder.toString(), is("https://rest.db.ripe.net/ripe/mntner/SHRYANE-MNT?password=FILTERED&param=value&override=FILTERED"));
    }
```

dadepo:
Unfortunately `javax.ws.rs.core.UriBuilder` does not work for all the use cases (well unless there is some extra code to handle the use cases)

The proposed solution does not take into consideration the following cases:

1. When a reason is given in the string. the solution replaces everything (including the reason)  with "FILTERED". i.e. 

```
http://test.net?unformatted=true&override=rsng,TEST-DBM-MNT,my reason -- reason given
results into
http://test.net/?password=FILTERED&override=FILTERED&unformatted=true -- reason stripped off
but should be 
http://test.net?unformatted=true&override=rsng,FILTERED,the reason -- reason should be in the string
```

2. When the URL is returned in the response body, the whole password query is removed. In audit, the value is replaced with FILTERED. The `replaceQueryParam` method does not take care of this as it just replaces values

3.  When override is used, the suggested method will still append `password=FILTERED`. But this should not be the case.
```
http://test.net?unformatted=true&override=rsng,TEST-DBM-MNT,my reason -- no password query params
results into
http://test.net/?password=FILTERED&override=FILTERED&unformatted=true -- password query params is added
but should be 
http://test.net?unformatted=true&override=rsng,FILTERED,the reason -- password query params should not be added
```

To make use of the `javax.ws.rs.core.UriBuilder` will need to ensure all the use cases are taken care of and I doubt if it will lead to a cleaner implementation. Hence I will rather leave the current implementation as is.

whois-api/src/main/java/net/ripe/db/whois/api/rest/RestServiceHelper.java
dadepo: This will involve passing in a `flag` parameter to `PasswordFilter.filterPasswordsInUrl` method, to tell it whether the url should be filtered for use in response body (then password query and value is removed) or whether the query should be filtered for audit logs (then value of password is replaced with `FILTERED`).    

This is because there is nothing in the log for the method to decide this.

Hence I will leave as is, as I agree with Fowler, [Flag arguments should generally be avoided](https://martinfowler.com/bliki/FlagArgument.html)

eshryane:
Makes sense, no UriBuilder then (I think we should use the same filtering code for the response and audit log), and there's no iterator to walk through query parameters either.

eshryane (approved):

PR #667, opened 2020-11-02
require an -mnt suffix in maintainer name

dadepo:


eshryane:
The syntax check is going to prevent modifying and deleting existing mntners. The extra validation should *only* be done on create. I suggest using a BusinessRuleValidator (i.e. have a look at MaintainerNameValidator as an example).

dadepo:
> The syntax check is going to prevent modifying and deleting existing mntners. The extra validation should _only_ be done on create. I suggest using a BusinessRuleValidator (i.e. have a look at MaintainerNameValidator as an example).

Updated as suggested

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/maintainer/MaintainerNameValidator.java
```diff
@@ -32,6 +33,9 @@
     @Override
     public void validate(final PreparedUpdate update, final UpdateContext updateContext) {
         final RpslObject updatedObject = update.getUpdatedObject();
+        if (Action.CREATE.equals(update.getAction()) && !updatedObject.getKey().toUpperCase().endsWith(MNT_NAME_SUFFIX)) {
```
eshryane, line 36: No need for toUpperCase(), CIString implements endsWith()

whois-update/src/main/java/net/ripe/db/whois/update/domain/UpdateMessages.java
```diff
@@ -176,6 +176,10 @@ public static Message invalidMaintainerForOrganisationType(final CharSequence or
         return new Message(Type.ERROR, "Value '%s' can only be set by the RIPE NCC for this organisation.", orgType);
     }
 
+    public static Message invalidMaintainerName() {
+        return new Message(Type.ERROR, "Maintainer name must end with a -MNT suffix");
```
eshryane, line 180: Update the message to clarify that the error is on CREATE only:
`When creating a MNTNER the name must end with an -MNT suffix`

whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/update/UpdateMaintainerSpec.groovy
```diff
@@ -304,6 +323,40 @@ class UpdateMaintainerSpec extends BaseQueryUpdateSpec {
         queryObject("-rGBT mntner CRE-MNT", "mntner", "CRE-MNT")
     }
 
+    def "create maintainer object with invalid  (without -mnt suffix) maintainer name"() {
```
eshryane, line 326: Is there a successful test to create a maintainer with an -mnt suffix?

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/maintainer/MaintainerNameValidator.java
dadepo: Removed

whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/update/UpdateMaintainerSpec.groovy
dadepo: Ys, there are existing test that checks the creation of a maintainer and they used -mnt suffix, for example `create self maintained maintainer object check pw hash filtering` and `create maintainer object maintained by other mntner`

whois-update/src/main/java/net/ripe/db/whois/update/domain/UpdateMessages.java
dadepo: done

eshryane (approved):

PR #670, opened 2020-11-10
Updated bouncy castle to latest (1.67)

eshryane:
Updated bouncy castle to latest (1.67): https://www.bouncycastle.org/latest_releases.html

Read release notes, no new features affecting us.

CI build passes, no new deprecated warnings.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,7 +32,7 @@
         <aspectj.version>1.8.8</aspectj.version>
         <awaitility.version>1.3.5</awaitility.version>
 
-        <bouncycastle.version>1.66</bouncycastle.version>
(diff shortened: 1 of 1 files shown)


dadepo (approved):
...Might be an idea to use something like https://dependabot.com/ to automate some of this?

eshryane:
> ...Might be an idea to use something like https://dependabot.com/ to automate some of this?

I'm not sure if that'll save much effort, as we also need to manually check that integration tests don't fail. We'd have to review each dependabot PR manually in any case.

We do have a number of out-of-date dependencies, maybe we should make some time to review those first?

eshryane:
Integration tests passed

PR #682, opened 2020-11-13
added ability to deploy to aws for prepdev

dadepo:


sbusk (approved):
We need to further generify the aws deployment and come up with better aws credentials handling but for now this works.

PR #684, opened 2020-11-16
Fix integration test due to mnt prefix

dadepo:


eshryane (approved):
Looks good (and confirmed integration tests are passing)

PR #690, opened 2020-11-18
Sonar coverage

maggarwal13:


pom.xml
```diff
@@ -685,21 +685,72 @@
                             <goal>prepare-agent</goal>
                         </goals>
                         <configuration>
-                            <append>true</append>
+                            <destFile>${project.build.directory}/jacoco-output/jacoco-unit-tests.exec</destFile>
```
eshryane, line 688: "append" was removed, was this deliberate? What's the effect of this?

eshryane:
Do the configuration changes also need to be applied to whois-internal ?

pom.xml
maggarwal13: Yes, Now we have a merged report with both unit test and integration just like whois-internal.

maggarwal13:
> Do the configuration changes also need to be applied to whois-internal ?

yes it already been merged in master. But for whois-internal sonar will always run on master branch in gitlab and whois is only scheduled as it takes lot longer time

pom.xml
eshryane: Ah, it was already done in whois-internal, thanks

PR #699, opened 2020-11-20
managed attribute country code for org

maggarwal13:


whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceTestIntegration.java
```diff
@@ -1563,6 +1563,79 @@ public void lookup_inetnum_managed_attributes_resource_holder_abuse_contact() {
...
+                .get(WhoisResources.class);
+
+        assertThat(searchinetnum.getWhoisObjects(), hasSize(1));
+        assertThat(searchinetnum.getWhoisObjects().get(0).getPrimaryKey().get(0).getValue(), is("10.0.0.0 - 10.0.0.255"));
+        assertThat(searchinetnum.getWhoisObjects().get(0).getAttributes().get(3).getName(), is("country"));
+        assertThat(searchinetnum.getWhoisObjects().get(0).getAttributes().get(3).getManaged(), is(nullValue()));
+    }
+
+    @Test
+    public void update_orgnasation_country_code_fails() {
```
sbusk, line 1610: typo: orgnasation -> organisation

This test is for testing the validation of country code updating only being allowed by RIPE NCC, not for the attribute being flagged as managed?

sbusk: I see the test case tests specifically that country code value can only be changed by RIPE NCC.
Can you add a test also where you do authenticate as RIPE and change the attribute value?

maggarwal13: There is separate test case to test that country code is shown as managed for organisation and not as managed attribute for other object type.

maggarwal13: I will add a test case that country code can be updated using overrides

maggarwal13: Done

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/organisation/LirRipeMaintainedAttributesValidator.java
```diff
@@ -30,7 +30,8 @@
     private static final List<AttributeType> RIPE_NCC_MANAGED_ATTRIBUTES = ImmutableList.of(
             AttributeType.MNT_BY,
             AttributeType.ORG,
-            AttributeType.ORG_TYPE);
+            AttributeType.ORG_TYPE,
+            AttributeType.COUNTRY);
```
eshryane, line 34: This validation is already done in CountryRipeMaintainedValidator.java. If you add it here also, there may be duplicate error messages in the response. 

I don't mind if you want to move it here for consistency, but then please delete the other validator.

eshryane (approved):

PR #707, opened 2020-12-02
use nexus to deploy

maggarwal13:


pom.xml
```diff
@@ -103,11 +105,11 @@
     <distributionManagement>
```
eshryane, line 105: Do we still need this section in the POM?

maggarwal13: I think we do, We are not using Ripe parent pom.xml. Also we are using maven command to upload and download artifacts from nexus.  where else do you think this information should come from?

eshryane: OK! Just wanted to confirm that we need the info in the POM.

PR #715, opened 2020-12-09
Handle rfc3986 IPv6 address provided by Jetty by getRemoteAddress() s…

sbusk:
…ince v9.4.32.


eshryane (approved):

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/RemoteAddressFilter.java
```diff
@@ -58,7 +58,11 @@ public String getRemoteAddr() {
         private static String getRemoteAddress(final HttpServletRequest request) {
             final String forwardedAddress = getForwardedAddress(request);
             if (forwardedAddress == null) {
-                return request.getRemoteAddr();
+                final String address = request.getRemoteAddr();
```
dadepo, line 61: Maybe use `InetAddresses.forUriString` instead? 

Inspired from this fix [here](https://github.com/paulmillar/dcache/commit/db66d579453ed1b7d340f9d3229be65638b3b731)

dadepo (approved):

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/RemoteAddressFilter.java
sbusk: I had yet to reply, I actually started with littering that everywhere where we use InetAddresses.forString() and run into the IPv6 parsing issue but realised the methods are not compatible, the forUriString() genuinely expects IPv6 addresses to always be contained within brackets so it would then result in a try catch with assuming it to be IPv4 in the catch clause.. not great and also a somewhat expensive operation.
The bracket check that has been implemented now was inspired by the forUriString() implementation though :)

PR #723, opened 2020-12-16
Delete object bug

maggarwal13:


whois-update/src/main/java/net/ripe/db/whois/update/handler/SingleUpdateHandler.java
```diff
@@ -102,7 +102,7 @@ public void handle(final Origin origin, final Keyword keyword, final Update upda
         }
 
         if (action == Action.DELETE && originalObject == null) {
-            updateContext.addMessage(update, UpdateMessages.objectNotFound(updatedObject.getFormattedKey()));
+            updateContext.addMessage(update, UpdateMessages.objectNotFound(update.getSubmittedObject().getFormattedKey()));
```
eshryane, line 105: So the error message will contain the key as the user submitted it?

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceTestIntegration.java
```diff
@@ -4707,6 +4707,33 @@ public void delete_route_out_of_region_redirect() {
...
+        databaseHelper.addObject(
+                "domain:        193.in-addr.arpa4\n" +
+                        "nserver:         test.ns.\n" +
+                        "mnt-by:          OWNER-MNT\n" +
+                        "source:          TEST\n");
+
+        final WhoisResources whoisResources = RestTest.target(getPort(), "whois/test/domain/193.in-addr.arpa4")
+                .request()
+                .get(WhoisResources.class);
+        assertThat(whoisResources.getWhoisObjects().get(0).getPrimaryKey().get(0).getValue(), is("193.in-addr.arpa4"));
```
eshryane, line 4721: Do you need to test the primary key in the test? It seems superfluous

whois-update/src/main/java/net/ripe/db/whois/update/handler/SingleUpdateHandler.java
maggarwal13: Yes, like it was before

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceTestIntegration.java
maggarwal13: Its just a way to test that before delete key exists and after it does not. I can remove that if you prefer that

eshryane: I don't think it's needed, and can be safely removed

whois-update/src/main/java/net/ripe/db/whois/update/handler/SingleUpdateHandler.java
eshryane: Great, thanks

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceTestIntegration.java
maggarwal13: Done

eshryane (approved):

PR #727, opened 2020-12-30
Reduce logging on Authoritative resource import exception

eshryane:
Don't log the entire stack trace if the authoritative resource import fails, it clutters up the log and we don't need to know.


eshryane:
Integration tests passed

PR #737, opened 2021-01-11
[DB-3298] turn join limit reached illegalArgumentException to bad req…

dadepo:
…uest response

### DIFF
--- a/whois-api/src/main/java/net/ripe/db/whois/api/rest/RestServiceHelper.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/rest/RestServiceHelper.java
@@ -142,7 +142,9 @@ public class RestServiceHelper {
         return createWebApplicationException(exception, request, Lists.newArrayList());
     }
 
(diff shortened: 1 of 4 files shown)


whois-api/src/main/java/net/ripe/db/whois/api/rest/RestServiceHelper.java
```diff
@@ -154,6 +156,9 @@ public static WebApplicationException createWebApplicationException(final Runtim
             }
             messages.addAll(queryException.getMessages());
 
+        } else if (exception instanceof IllegalArgumentException) {
+            responseBuilder = Response.status(Response.Status.BAD_REQUEST);
+            messages.add(QueryMessages.tooManyArguments());
```
eshryane, line 161: Can't there be other reasons why an IllegalArgumentException is thrown?

eshryane: Can you catch the exception closer to where it happens and add a global error there instead?

dadepo: updated

eshryane (approved):

PR #738, opened 2021-01-12
Modified date check generated attributes

maggarwal13:


whois-api/src/test/java/net/ripe/db/whois/api/rest/SyncUpdatesServiceTestIntegration.java
```diff
@@ -453,6 +455,47 @@ public void create_person_with_changed_attributes() {
...
+                "admin-c:        TP1-TEST\n" +
+                "tech-c:         TP1-TEST\n" +
+                "mnt-by:         mntner-mnt\n" +
+                "source:         TEST-NONAUTH\n" +
+                "override:       agoston,zoh\n"))
+                .request()
+                .get(String.class);
+
+        assertThat(response, containsString("Modify SUCCEEDED: [aut-num] AS104"));
+        assertThat(databaseHelper.lookupObject(ObjectType.AUT_NUM, "AS104").getValueForAttribute(AttributeType.LAST_MODIFIED), is(not(orginialModifiedDate)));
```
eshryane, line 494: If tests are using 'testDateTimeProvider' then time doesn't move forwards between tests, so last-modified will be the same, always?

maggarwal13: Actually when creating a object, I specifically added last-modified to  " 2001-02-04T17:00:00Z". So while adding it is not using the testDateTimeProvider. 

Also assertion is done on comparing the last-modified before and after the update and it should be not equal.
But yes for readability I will move testDateTimeProvider time between creating the aut-num and updating the aut-num

maggarwal13: Done

eshryane: Thanks Mahesh!

eshryane (approved):

PR #742, opened 2021-01-18
allow reading of whois properties from classpath for aws

maggarwal13:


maggarwal13:
whois.properties are loaded last instead of parameter store properties

PR #746, opened 2021-01-22
Removed unused class dependency

dadepo:


dadepo:
Browsing through the code and noticed some unused dependency. Adding only you @sbusk  to the PR because it is a small change and you are currently the runner :)

eshryane:
The ```loadbalancer.nrtm.timeout``` property was added in 39e4e847 but was never used, it's safe to remove

eshryane:
IpRanges was present since the initial import 8227230b and was used to support a Syncupdates REDIRECT flag to disable notifications, also safe to remove. 

We should remove that REDIRECT flag altogether, as it doesn't do anything.

eshryane (approved):

dadepo:
> The `loadbalancer.nrtm.timeout` property was added in [39e4e84](https://github.com/RIPE-NCC/whois/commit/39e4e847589af4c66dee88d2e291a23dbe84e85f) but was never used, it's safe to remove

I just checked in the property files and this seems to have been removed already

dadepo:
> IpRanges was present since the initial import [8227230](https://github.com/RIPE-NCC/whois/commit/8227230b2eeca4638fcf154569ad6716b7233666) and was used to support a Syncupdates REDIRECT flag to disable notifications, also safe to remove.
> 
> We should remove that REDIRECT flag altogether, as it doesn't do anything.

by REDIRECT flag I guess you are referring to `Command.REDIRECT` as seen [here](https://github.com/RIPE-NCC/whois/blob/fe3aef285bcd7ea380e53bc11c1bd2d4e6bba002/whois-api/src/main/java/net/ripe/db/whois/api/rest/SyncUpdatesService.java#L463:29). And removal of this field together with any part of the system that makes use of this?

If you can confirm...then I'll add an issue in our backlog to work on this

PR #749, opened 2021-01-27
Fix aut-num managed flag in Whois Search Response

eshryane:
Aut-num objects with status ASSIGNED or LEGACY and maintained by a RIPE NCC maintainer, should appear as "managed" in the DB web query response (using a managed flag at the object and attribute level in the Whois REST API lookup and search responses).


sbusk (approved):

eshryane:
Integration tests passed

PR #753, opened 2021-02-04
Use sequential instance names.

sbusk:


whois-commons/src/main/java/net/ripe/db/whois/common/InstanceLock.java
```diff
@@ -0,0 +1,68 @@
...
+                final String instanceName = INSTANCE_NAME_PREFIX + i;
+                final Path instanceDir = base.resolve(Path.of(instanceName));
+
+                if (!Files.isDirectory(instanceDir)) {
+                    Files.createDirectory(instanceDir);
+                }
+
+                lockFile = instanceDir.resolve(LOCK_FILE_NAME);
+
+                if (!Files.exists(lockFile)) {
```
eshryane, line 42: else lockFile = null?

eshryane: Maybe not an issue, if the exception stops the application anyway

whois-commons/src/main/java/net/ripe/db/whois/common/InstanceLock.java
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.commons.lang.StringUtils;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+public class InstanceLock {
+
+    private final static int MAX_INSTANCES = 16;
+    final static String INSTANCE_NAME_PREFIX = "whois-aws-";
```
eshryane, line 12: Follow `private static final` order convention

sbusk: The constant is package protected so it can be used by the test.

whois-commons/src/main/java/net/ripe/db/whois/common/InstanceLock.java
sbusk: Correct, if a lock cannot be obtained the application cannot start.

eshryane (approved):

PR #772, opened 2021-02-17
Updated to Jetty 10

eshryane:
* https://webtide.com/jetty-10-and-11-have-arrived/


eshryane:
Integration tests passed

PR #776, opened 2021-02-23
Hazelcast java config

maggarwal13:


whois-query/src/main/java/net/ripe/db/whois/query/hazelcast/HazelcastInstanceManager.java
```diff
@@ -14,42 +14,67 @@
...
 @DeployedProfile
 public class HazelcastInstanceManager {
     private static final Logger LOGGER = LoggerFactory.getLogger(HazelcastInstanceManager.class);
 
     final String interfaces;
+    final int port;
 
     @Autowired
-    public HazelcastInstanceManager( @Value("${hazelcast.config.interfaces}") final String interfaces) {
+    public HazelcastInstanceManager(@Value("${hazelcast.config.interfaces}") final String interfaces, @Value("${hazelcast.port:5701}") final int port) {
```
eshryane, line 28: It's not part of this PR (hazelcast.config.interfaces), but you can use Spring to split a comma-separated value for you, and inject into a list. For example, `@Value("#{'${email.excluded:}'.split(',')}")` in `ExcludedEmailValidator.java`

whois-query/src/main/java/net/ripe/db/whois/query/hazelcast/HazelcastInstanceManager.java
```diff
@@ -14,42 +14,67 @@
...
     @Bean
     @Profile(WhoisProfile.AWS_DEPLOYED)
     @Primary
     public HazelcastInstance hazelcastAwsInstance() {
-        LOGGER.info("Creating hazelcast instance with AWS deployed profile, interfaces {}", interfaces);
+        LOGGER.info("Creating hazelcast instance with AWS deployed profile");
 
-        final Config config = new Config();
-        config.getNetworkConfig().getJoin().getMulticastConfig().setEnabled(false);
+        final Config config = getGenericConfig();
```
eshryane, line 55: Is Multicast config disabled by default?

maggarwal13: its enabled by default. We need to disable if using any other means

whois-query/src/main/java/net/ripe/db/whois/query/hazelcast/HazelcastInstanceManager.java
maggarwal13: I did not used it specifically as it is also used in aws one and there it won't be a list. As addInterface does not take list which we use in aws but for tcp config it has that function to add list of members

eshryane: OK! Thanks

eshryane (approved):

PR #782, opened 2021-03-12
Handle trailing spaces after email address when converting to Punycode

eshryane:
Separate email address and leading/trailing space matching into separate patterns.

There was a bug where trailing space was included in the Punycode conversion, this was breaking if there were a lot of spaces. This is now fixed.


eshryane:
Integration tests have passed

sbusk (approved):

eshryane:
I don't trim spaces before and after the email address, because I don't want to change formatting. Also it's important to preserve spaces if there's a comment.

PR #783, opened 2021-03-15
Enable remote jmx connection from within private network

dadepo:


whois-db/pom.xml
```diff
@@ -132,6 +132,12 @@
                                     <jvmFlag>-Dlog4j.configurationFile=file:/export/services/whois/log4j2.xml</jvmFlag>
                                     <jvmFlag>-Dbase.dir=/export/services/whois</jvmFlag>
                                     <jvmFlag>-Dinstance.name.prefix=whois-aws-</jvmFlag>
+                                    <jvmFlag>-Dcom.sun.management.jmxremote=true</jvmFlag>
+                                    <jvmFlag>-Dcom.sun.management.jmxremote.port=6001</jvmFlag>
+                                    <jvmFlag>-Dcom.sun.management.jmxremote.rmi.port=6001</jvmFlag>
+                                    <jvmFlag>-Dcom.sun.management.jmxremote.authenticate=false</jvmFlag>
```
dadepo, line 138: The relative lax authentication is used since this will only be accessed via the internal aws networks. I can look into setting up authentication and/or with https if there is a strong argument for those

sbusk (approved):

PR #794, opened 2021-04-13
Print properties

maggarwal13:


whois-db/src/main/java/net/ripe/db/whois/db/WhoisServer.java
```diff
@@ -50,6 +54,15 @@ public void run() {
 
         whoisServer.start();
 
+        final MutablePropertySources sources = applicationContext.getEnvironment().getPropertySources();
+        StreamSupport.stream(sources.spliterator(), false)
+                .filter(ps -> ps instanceof EnumerablePropertySource)
+                .map(ps -> ((EnumerablePropertySource) ps).getPropertyNames())
+                .flatMap(Arrays::stream)
+                .distinct()
+                .filter(prop -> !(prop.contains("credentials") || prop.contains("password")))
```
eshryane, line 63: Can you log the property, but put in "***" or something for the value? As it is, nothing is logged, and so you can't tell whether the property is defined or not.

maggarwal13: done

eshryane (approved):

eshryane:
An implication of reading the property keys from `ApplicationContext.getEnvironment().getPropertySources()` is that a property may be defined multiple times (in different property sources), but the value is whatever Spring resolves it to (whichever property comes last?). In this case, we'll see the property logged multiple times, but with the final value each time?

maggarwal13:
> An implication of reading the property keys from `ApplicationContext.getEnvironment().getPropertySources()` is that a property may be defined multiple times (in different property sources), but the value is whatever Spring resolves it to (whichever property comes last?). In this case, we'll see the property logged multiple times, but with the final value each time?

No, I use distinct for that matter. So, property won't be printed twice.
Honestly in our case with aws parameter store there cannot be multiple property sources  and with inhouse just one property files.So It's safe anyhow

PR #801, opened 2021-04-14
Rewrites

sbusk:


whois-api/src/main/java/net/ripe/db/whois/api/httpserver/FixedResponseRule.java
```diff
@@ -0,0 +1,36 @@
...
+
+    private final int responseCode;
+
+    public FixedResponseRule(final int responseCode) {
+        this.responseCode = responseCode;
+    }
+
+    @Override
+    public String matchAndApply(String target, HttpServletRequest request, HttpServletResponse response) throws IOException {
+        if (responseCode >= 400) {
```
eshryane, line 19: You could avoid the hard-coded status code with (Jetty) `HttpStatus.isClientError(code) || HttpStatus.isServerError(code)`

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/HttpTransportRule.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Override
+    public String matchAndApply(String target, HttpServletRequest request, HttpServletResponse response) throws IOException {
+        return hasTransport(request)? delegate.matchAndApply(target, request, response) : null;
+    }
+
+
+    private boolean hasTransport(final HttpServletRequest request) {
+        final Enumeration<String> header = request.getHeaders("X-Forwarded-Proto");
```
eshryane, line 27: This header constant is defined in (Jetty) HttpHeader.X_FORWARDED_PROTO (we have a mix of Jakarta and Jetty constants in Whois)

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/RewriteEngine.java
```diff
@@ -74,4 +76,57 @@ public RewriteHandler getRewriteHandler() {
...
 
+    private void restRedirectRules(final VirtualHostRuleContainer virtualHost) {
+        virtualHost.addRule(new CaseInsensitiveRewriteRegexRule(
+    "^/(search|geolocation|metadata|abuse-contact|references|autocomplete|domain)/?(.*)$",
+    "/whois/$1/$2"
+        ));
+
+        virtualHost.addRule(
+            new HttpTransportRule("https",
+                new HttpMethodRule(Set.of("post", "put", "delete"), new CaseInsensitiveRewriteRegexRule(
```
eshryane, line 87: Use Jakarta or Jetty HttpMethod for constant values?

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/RewriteEngine.java
```diff
@@ -74,4 +76,57 @@ public RewriteHandler getRewriteHandler() {
...
+                String.format("^/(%s|%s)/(.*)$", source, nonAuthSource),
+                "/whois/$1/$2"
+        ))));
+
+        // Don't allow passwords over plain HTTP
+        virtualHost.addRule(
+            new HttpTransportRule("http",
```
eshryane, line 94: Use Jetty HttpScheme for "HTTP" and "HTTPS" constants?

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/FixedResponseRule.java
sbusk: Done.

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/RewriteEngine.java
sbusk: Jetty :)

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/HttpTransportRule.java
sbusk: I've gone the Jetty way.

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/RewriteEngine.java
sbusk: Done.

eshryane:
Is it necessary to have a `rewrite.engine.enabled` property? Does it ever need to be turned off?

sbusk:
> Is it necessary to have a `rewrite.engine.enabled` property? Does it ever need to be turned off?

It's necessary when we don't have Apache in front to rewrite. 
When Apache is in front it should not be enabled as it will then do double rewriting..

eshryane:
> > Is it necessary to have a `rewrite.engine.enabled` property? Does it ever need to be turned off?
> 
> It's necessary when we don't have Apache in front to rewrite.
> When Apache is in front it should not be enabled as it will then do double rewriting..

Can that happen? Once the URL is rewritten it shouldn't match again?

sbusk:
> > > Is it necessary to have a `rewrite.engine.enabled` property? Does it ever need to be turned off?
> > 
> > 
> > It's necessary when we don't have Apache in front to rewrite.
> > When Apache is in front it should not be enabled as it will then do double rewriting..
> 
> Can that happen? Once the URL is rewritten it shouldn't match again?

It will match the default case (resulting in a 400):

Consider the example
/ripe/inetnum/10.0.0.0

Will be rewritten by Apache to /whois/ripe/inetnum/10.0.0.0.

And will then match only the default case in Jetty so it will result in a 400.

eshryane (approved):

PR #802, opened 2021-04-15
override with non trusted ip

maggarwal13:


whois-api/src/test/java/net/ripe/db/whois/api/rest/OverrideWithNonTrustedIpTestIntegration.java
```diff
@@ -0,0 +1,211 @@
...
+        databaseHelper.insertUser(User.createWithPlainTextPassword("db_e2e_1", "zoh", ObjectType.PERSON));
+        databaseHelper.insertUser(User.createWithPlainTextPassword("person", "zoh", ObjectType.PERSON));
+
+        databaseHelper.addObject(OWNER_MNT);
+        databaseHelper.updateObject(TEST_PERSON);
+        testDateTimeProvider.setTime(LocalDateTime.parse("2001-02-04T17:00:00"));
+    }
+
+    @BeforeClass
+    public static void beforeClass() {
```
sbusk, line 86: Reset the ipranges.trusted value to whatever it was before in a @AfterClass method to prevent test failures in other tests.

whois-api/src/test/java/net/ripe/db/whois/api/rest/OverrideWithNonTrustedIpTestIntegration.java
```diff
@@ -0,0 +1,211 @@
...
+                .post(Entity.entity(map(PAULETH_PALTHEN), MediaType.APPLICATION_XML), WhoisResources.class);
+
+        final WhoisObject object = whoisResources.getWhoisObjects().get(0);
+        assertThat(whoisResources.getErrorMessages().get(0).getText(), is("Authorisation override used"));
+        assertThat(object.getType(), is("person"));
+        assertThat(object.getPrimaryKey(), contains(new Attribute("nic-hdl", "PP1-TEST")));
+    }
+
+    @Test
+    public void delete_person_with_override_with_sso_succeeds() {
```
sbusk, line 158: This tests the same as create (ip not trusted, valid sso + override) so this test is a bit redundant?

whois-api/src/test/java/net/ripe/db/whois/api/rest/OverrideWithNonTrustedIpTestIntegration.java
maggarwal13: Every test case reloads the test properties file  and it is defined there. So should not be an issue
But still I will clear the property in afterclass

sbusk (commented):
Maybe also add tests for scenario where ip is trusted, override is valid, sso is valid so that execution path is covered and there are no adverse effects.

whois-api/src/test/java/net/ripe/db/whois/api/rest/OverrideWithNonTrustedIpTestIntegration.java
maggarwal13: Agree, I was just including create, update and delete. If there is no need I can remove these two.

whois-api/src/test/java/net/ripe/db/whois/api/rest/OverrideWithNonTrustedIpTestIntegration.java
```diff
@@ -0,0 +1,211 @@
...
+import javax.ws.rs.core.MediaType;
+import java.time.LocalDateTime;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.hasSize;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.fail;
+
+// FIXME: make this into a suite that runs twice: once with XML, once with JSON
```
sbusk, line 37: Either do what this suggestion does or remove it but don't copy paste the FIXME ;)

maggarwal13:
> Maybe also add tests for scenario where ip is trusted, override is valid, sso is valid so that execution path is covered and there are no adverse effects.

There are lots of existing test cases for this scenario. WhoisrestServiceTestIntegration is full of them.  By default ip is trusted

sbusk:
> > Maybe also add tests for scenario where ip is trusted, override is valid, sso is valid so that execution path is covered and there are no adverse effects.
> 
> There are lots of existing test cases for this scenario. WhoisrestServiceTestIntegration is full of them. By default ip is trusted

Not the particular case that ip is trusted, override is valid, sso is valid though, that execution path is new.

maggarwal13:
> > > Maybe also add tests for scenario where ip is trusted, override is valid, sso is valid so that execution path is covered and there are no adverse effects.
> > 
> > 
> > There are lots of existing test cases for this scenario. WhoisrestServiceTestIntegration is full of them. By default ip is trusted
> 
> Not the particular case that ip is trusted, override is valid, sso is valid though, that execution path is new.

Ahh, I will add one with existing ones

maggarwal13:
> > > > Maybe also add tests for scenario where ip is trusted, override is valid, sso is valid so that execution path is covered and there are no adverse effects.
> > > 
> > > 
> > > There are lots of existing test cases for this scenario. WhoisrestServiceTestIntegration is full of them. By default ip is trusted
> > 
> > 
> > Not the particular case that ip is trusted, override is valid, sso is valid though, that execution path is new.
> 
> Ahh, I will add one with existing ones

done

PR #804, opened 2021-04-16
Jvm dns ttl

maggarwal13:


whois-commons/src/main/java/net/ripe/db/whois/common/JvmSecurityManager.java
```diff
@@ -0,0 +1,19 @@
...
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@DeployedProfile
+public class JvmSecurityManager {
+    private static final Logger LOGGER = LoggerFactory.getLogger(JvmSecurityManager.class);
+
+    @Autowired
+    public JvmSecurityManager() {
+        java.security.Security.setProperty("networkaddress.cache.ttl", "30");
```
eshryane, line 16: Can we set this property at runtime? Will it make a difference?

The OpenJDK code suggests this property is only read once (it's in a static block):
https://github.com/AdoptOpenJDK/openjdk-jdk/blob/jdk-11+28/src/java.base/share/classes/sun/net/InetAddressCachePolicy.java#L120

eshryane: Stack overflow to the rescue:

`This code must execute before any other code in the JVM attempts to perform networking operations.`

Is it enough to also put these properties in a static block?

https://github.com/AdoptOpenJDK/openjdk-jdk/blob/jdk-11+28/src/java.base/share/classes/sun/net/InetAddressCachePolicy.java#L120

eshryane: A workaround to not setting the security properties is to set the *system* properties `sun.net.inetaddr.ttl` and `sun.net.inetaddr.negative.ttl`. They are deprecated but work for now.

https://stackoverflow.com/questions/41708885/best-way-to-set-networkaddress-cache-ttl

maggarwal13: There is another way to provide security file but let's have a discussion about it, we need to think about implications of it.
https://dzone.com/articles/how-override-java-security

maggarwal13: Lets try not to use the deprecated methods first if we can find better way to do this

PR #806, opened 2021-04-21
Java security jvm ttl

maggarwal13:


tools/jvm.java.security
```diff
@@ -0,0 +1,2 @@
+networkaddress.cache.ttl=60
```
eshryane, line 1: Use an application-specific name for the file like `whois.java.security` so it's clear it's separate from the JVM

whois-db/src/main/java/net/ripe/db/whois/db/WhoisServer.java
```diff
@@ -102,4 +104,18 @@ private void stopService(final ApplicationService applicationService, boolean fo
...
     }
+
+    private static void printJvmSecurityProperties() {
+        if(!Boolean.valueOf(Security.getProperty("security.overridePropertiesFile"))) {
+            LOGGER.warn("security.overridePropertiesFile is false, cannot override security values");
+        }
+
+        final String networdkAddrCacheTtl = Security.getProperty("networkaddress.cache.ttl");
+        LOGGER.info("networkaddress.cache.ttl: {}", networdkAddrCacheTtl);
+        LOGGER.info("networkaddress.cache.negative.ttl: {}", Security.getProperty("networkaddress.cache.negative.ttl"));
```
eshryane, line 115: Also create a variable for the negative TTL value and log separately

whois-db/src/main/java/net/ripe/db/whois/db/WhoisServer.java
```diff
@@ -102,4 +104,18 @@ private void stopService(final ApplicationService applicationService, boolean fo
...
+
+        if(networdkAddrCacheTtl == null || networdkAddrCacheTtl.equals("-1")) {
+            LOGGER.warn("Networkaddress.cache.ttl is not sert properly");
```
eshryane, line 118: Typo: "sert" -> "set"

whois-db/src/main/java/net/ripe/db/whois/db/WhoisServer.java
maggarwal13: I am logging the value for negative ttl, there is no need to create a seperate variable for it as its default is 10s  and I am not logging it as a warning

eshryane: The default may change (it's JVM specific), I think both should be logged in the same way

maggarwal13: ok, I will put warning for negative one too

eshryane (approved):

PR #807, opened 2021-04-21
Switching to utc and have cron jobs run in AMS

dadepo:


dadepo:
Major code changes done. I still need to perform extensive testing.

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestApiTest.java
```diff
@@ -41,14 +41,14 @@ public void getName() {
...
 
         assertThat(subject.getResponseHeader(), is(" - From-Host: 127.0.0.1\n - Date/Time: Sun Mar 3 12:55:00 2013\n"));
     }
 
     @Test
     public void notification_header_for_any_request() {
-        when(dateTimeProvider.getCurrentDateTime()).thenReturn(LocalDateTime.of(2013, 3, 3, 12, 55));
+        when(dateTimeProvider.getLocalDateTimeUtc()).thenReturn(LocalDateTime.of(2013, 3, 3, 12, 55));
 
         assertThat(subject.getNotificationHeader(), is(" - From-Host: 127.0.0.1\n - Date/Time: Sun Mar 3 12:55:00 2013\n"));
```
sbusk, line 53: Not part of your changes but shouldn't we add the TZ to the date/time reported by whois

whois-commons/src/main/java/net/ripe/db/whois/common/DateTimeProvider.java
```diff
@@ -7,14 +7,14 @@
 
 public interface DateTimeProvider {
 
-    /** Return the current date */
-    LocalDate getCurrentDate();
+    /** Return the current date time in Amsterdam  */
+    LocalDateTime getDateTimeAms();
```
sbusk, line 11: Maybe use Cest to make it 100% crystal clear? Often people say CET/Amsterdam interchangably but it really isn't..

whois-commons/src/main/java/net/ripe/db/whois/common/scheduler/DailyScheduledTask.java
```diff
@@ -1,5 +1,6 @@
 package net.ripe.db.whois.common.scheduler;
 
 public interface DailyScheduledTask extends Runnable {
+    String runTimezone = "Europe/Amsterdam";
```
sbusk, line 4: It's a constant in an interface so should be RUN_TIME_ZONE

whois-scheduler/src/main/java/net/ripe/db/whois/scheduler/SchedulerConfig.java
```diff
@@ -0,0 +1,62 @@
...
+        scheduler.setPoolSize(2);
+        return scheduler;
+    }
+
+    @Bean
+    public LockProvider lockProvider(@Qualifier("internalsDataSource") DataSource dataSource) {
+        return new JdbcTemplateLockProvider(
+                JdbcTemplateLockProvider.Configuration.builder()
+                        .withJdbcTemplate(new JdbcTemplate(dataSource))
+                        .withTimeZone(TimeZone.getTimeZone("Europe/Amsterdam"))
```
sbusk, line 56: I understand the jobs run in CEST but should we still be persisting UTC so dates are consistent across the db

pom.xml
```diff
@@ -87,7 +87,7 @@
         <solrj.version>1.4.1</solrj.version>
         <stax-utils.version>20070216</stax-utils.version>
         <spock.version>1.3-groovy-2.5</spock.version>
-        <shedlock.version>2.5.0</shedlock.version>
+        <shedlock.version>4.23.0</shedlock.version>
```
eshryane, line 90: Are there any breaking changes in this version change? Maybe update shedlock itself in a separate PR?

whois-scheduler/src/main/java/net/ripe/db/whois/scheduler/SchedulerConfig.java
eshryane: Also, do we need to run *all* scheduled jobs using the Amsterdam timezone?

We could assume the default (local, UTC) timezone where it doesn't matter, and where it does matter use the `zone` argument.

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestApiTest.java
dadepo: We should. I'll update to include this

eshryane: We need the timezone in all timestamps (UTC = 'Z')

pom.xml
dadepo: The configuration is easier and easier to find documentation for, hence why I decided to upgrade as part of setting the job to run in AMS time. As per breaking changes, extensive testing would help answer that with surety.

whois-commons/src/main/java/net/ripe/db/whois/common/DateTimeProvider.java
dadepo: But should it be CET or CEST? Could not decide which it should be hence I went with Amsterdam Time. Not sure what the general convention is...

sbusk: Please disregard my previous comment.. reading up on it using Ams definitely is more clear than Cet/Cest as that depends on the time of year ;)

whois-scheduler/src/main/java/net/ripe/db/whois/scheduler/SchedulerConfig.java
eshryane: Do we need to default to Amsterdam time? I think we should default to UTC and only make exceptions where necessary (using the zone argument on the `@Scheduled` annotation)

pom.xml
dadepo: Reading the release notes here https://github.com/lukas-krecan/ShedLock#release-notes nothing jumps out. Via testing we know for sure

whois-scheduler/src/main/java/net/ripe/db/whois/scheduler/SchedulerConfig.java
dadepo: That is indeed the intention. Default to UTC but run the midnight jobs at Amsterdam time. I have updated the configuration and confirmed the daly midnight jobs now kicks in at midnight Amsterdam time:

```

2021-04-29T00:00:00,021 INFO  [ScheduledTaskAspect] Starting scheduled task: net.ripe.db.whois.scheduler.task.acl.ResetPersonalObjectAccounting@289fff3c
2021-04-29T00:00:00,079 INFO  [ScheduledTaskAspect] Scheduled task: net.ripe.db.whois.scheduler.task.acl.ResetPersonalObjectAccounting@289fff3c took 57.34 ms
2021-04-29T00:00:00,120 INFO  [ScheduledTaskAspect] Starting scheduled task: net.ripe.db.whois.scheduler.task.acl.AutomaticPermanentBlocks@1e6a8d01
2021-04-29T00:00:00,129 INFO  [ScheduledTaskAspect] Scheduled task: net.ripe.db.whois.scheduler.task.acl.AutomaticPermanentBlocks@1e6a8d01 took 8.803 ms
```

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestApiTest.java
dadepo: Updated to include the 'Z'

whois-scheduler/src/main/java/net/ripe/db/whois/scheduler/SchedulerConfig.java
dadepo: Also:
```
2021-04-30T00:00:00,132 INFO  [RpslObjectsExporter] Database export started
```

whois-commons/src/main/java/net/ripe/db/whois/common/DateTimeProvider.java
```diff
@@ -7,14 +7,11 @@
 
 public interface DateTimeProvider {
 
-    /** Return the current date */
-    LocalDate getCurrentDate();
```
eshryane, line 11: Do you need to rename the methods? It causes a huge diff in the PR (also in whois-internal)

dadepo: I noticed the huge diff also :/
My preference is to rename, as this makes the timezone return by these methods explicit...something which I think is handy especially given the fact we could be talking about two different timezones...

Maybe a compromise is to rename the class?  `UTCDateTimeProvider`. That should reduce the PR diff. Is that a good middle ground?

eshryane: Good idea, I'd be happy with that! Thanks

pom.xml
eshryane: Thanks

whois-scheduler/src/main/java/net/ripe/db/whois/scheduler/SchedulerConfig.java
eshryane: OK thanks

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestApiTest.java
eshryane: Thank you

whois-commons/src/main/java/net/ripe/db/whois/common/DateTimeProvider.java
dadepo: ok, so renaming the class will also lead to a huge diff. so decided to just update the comments instead mentioning the usage of UTC

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageParser.java
```diff
@@ -69,7 +70,7 @@ public MailMessage parse(final MimeMessage message, final UpdateContext updateCo
         if (deliveryDate != null && deliveryDate.length > 0 && deliveryDate[0].length() > 0) {
             messageBuilder.date(deliveryDate[0]);
```
dadepo, line 71: when no delivery date is found in `MimeMessage` message, `now` in `utc` is set for the date. Confirm in the case where there is a delivery date in  `MimeMessage` message that the timezone is aligned with when `now` is used

whois-commons/src/main/java/net/ripe/db/whois/common/DateTimeProvider.java
sbusk: Wouldn't it be better to have a UTCDateTimeProvider as implementation class rather than renaming the interface?
So we have a DateTimeProvider for which we currently have 2 implementations: ClockDateTimeProvider (which is based on system timezone) and TestDateTimeProvider, and then adding a fixed (UTC) timezone DateTimeProvider impl

tools/log4j2.xml
```diff
@@ -36,7 +36,7 @@
                      ignoreExceptions="false">
             <ThresholdFilter level="debug"/>
             <PatternLayout>
-                <Pattern>%d{yyyyMMdd HH:mm:ss} %m%n</Pattern>
+                <Pattern>%d{yyyy-MM-dd'T'HH:mm:ss'Z'} %m%n</Pattern>
```
eshryane, line 39: Don't change the date/time format in the whois logs, as it breaks any parsing code.

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageParser.java
eshryane: Do we need an integration test for this scenario?

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageParser.java
```diff
@@ -69,7 +70,7 @@ public MailMessage parse(final MimeMessage message, final UpdateContext updateCo
...
         } else {
-            messageBuilder.date(DATE_FORMAT.format(ZonedDateTime.now()));
+            messageBuilder.date(DATE_FORMAT.format(ZonedDateTime.now(ZoneOffset.UTC)));
```
eshryane, line 73: Could ```dateTimeProvider.getCurrentDateTimeUtc()``` be used instead?

whois-api/src/main/java/net/ripe/db/whois/api/rest/SyncUpdate.java
```diff
@@ -50,7 +48,7 @@ private String getHeader() {
                 " - From-Host: %s\n" +
                 " - Date/Time: %s\n",
                 remoteAddress,
-                DATE_FORMAT.format(dateTimeProvider.getCurrentDateTime()));
+                FormatHelper.dayDateTimeToUtcString(dateTimeProvider.getCurrentDateTime()));
```
eshryane, line 51: Can we get rid of ```dateTimeProvider.getCurrentDateTimeUtc()``` ?

whois-commons/src/main/java/net/ripe/db/whois/common/ClockDateTimeProvider.java
```diff
@@ -13,18 +13,19 @@
...
 
     @Override
     public LocalDateTime getCurrentDateTime() {
-        return LocalDateTime.now();
+        return getCurrentZonedDateTime().toLocalDateTime();
     }
 
     @Override
-    public ZonedDateTime getCurrentDateTimeUtc() {
+    public ZonedDateTime getCurrentZonedDateTime() {
```
eshryane, line 28: There shouldn't be any difference between LocalDateTime and ZonedDateTime, is ZonedDateTime used for the 'Z' timezone?

whois-db/src/main/resources/applicationContext-whois-base.xml
```diff
@@ -10,7 +10,9 @@
     <context:mbean-server/>
     <context:mbean-export/>
 
+    <context:annotation-config />
+    <bean class="net.ripe.db.whois.scheduler.SchedulerConfig" />
```
eshryane, line 14: Why no longer include ```applicationContext-scheduler.xml``` ?

whois-scheduler/src/main/java/net/ripe/db/whois/scheduler/task/acl/ResetPersonalObjectAccounting.java
```diff
@@ -17,7 +17,7 @@ public ResetPersonalObjectAccounting(final PersonalObjectAccounting personalObje
     }
 
     @Override
-    @Scheduled(cron = "0 0 0 * * *")
+    @Scheduled(cron = "0 0 0 * * *", zone = EUROPE_AMSTERDAM)
```
eshryane, line 20: Does this job need to run on Amsterdam time? the ACL data will be in UTC

whois-scheduler/src/main/java/net/ripe/db/whois/scheduler/task/acl/AutomaticPermanentBlocksCleanup.java
```diff
@@ -26,7 +26,7 @@ public AutomaticPermanentBlocksCleanup(final DateTimeProvider dateTimeProvider,
     }
 
     @Override
-    @Scheduled(cron = "0 0 0 * * *")
+    @Scheduled(cron = "0 0 0 * * *", zone = EUROPE_AMSTERDAM)
     @SchedulerLock(name = "AutomaticPermanentBlocksCleanup")
```
eshryane, line 30: Does this job need to run on Amsterdam time? the ACL data will be in UTC

whois-scheduler/src/main/java/net/ripe/db/whois/scheduler/task/acl/AutomaticPermanentBlocks.java
```diff
@@ -36,7 +36,7 @@ public AutomaticPermanentBlocks(final DateTimeProvider dateTimeProvider,
     }
 
     @Override
-    @Scheduled(cron = "0 0 0 * * *")
+    @Scheduled(cron = "0 0 0 * * *", zone = EUROPE_AMSTERDAM)
```
eshryane, line 39: Does this job need to run on Amsterdam time? the ACL data will be in UTC

whois-scheduler/src/main/java/net/ripe/db/whois/scheduler/task/autnum/LegacyAutnumReloadTask.java
```diff
@@ -20,7 +20,7 @@ public LegacyAutnumReloadTask(final LegacyAutnum legacyAutnum) {
     }
 
     @Override
-    @Scheduled(cron = "0 1/15 * * * *")
+    @Scheduled(cron = "0 1/15 * * * *", zone = EUROPE_AMSTERDAM)
```
eshryane, line 23: This job should run every 15 minutes, regardless of the timezone

whois-scheduler/src/main/java/net/ripe/db/whois/scheduler/task/export/DatabaseTextExport.java
```diff
@@ -16,7 +16,7 @@ public DatabaseTextExport(final RpslObjectsExporter rpslObjectsExporter) {
     }
 
     @Override
-    @Scheduled(cron = "0 0 0 * * *")
+    @Scheduled(cron = "0 0 0 * * *", zone = EUROPE_AMSTERDAM)
```
eshryane, line 19: Perhaps comment that *this* job should run in Amsterdam time as the generated file on the FTP site will then appear at a consistent time

whois-commons/src/main/java/net/ripe/db/whois/common/ClockDateTimeProvider.java
dadepo: > is ZonedDateTime used for the 'Z' timezone?

Yes. It is the current time that includes the time zone related info

eshryane: OK (assuming we still need both)

tools/log4j2.xml
dadepo: reverted

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageParser.java
dadepo: The code has been updated to reflect the latest version.

But having said that not sure how we can have this covered by integration test as it depends on checking that whatever timestamp the smtp server uses, is in UTC. And this can only be confirmed by checking directly on the servers.

eshryane: Can't we set ```Delivery-date:``` to whatever we want in a test email for an integration test?

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageParser.java
dadepo: Updated to use `dateTimeProvider.getCurrentZonedDateTime()`

whois-api/src/main/java/net/ripe/db/whois/api/rest/SyncUpdate.java
dadepo: In theory yes. but that would mean the places where we are using `dateTimeProvider.getCurrentDateTime()` to just get the current time without timezone information would have to updated to something like `dateTimeProvider.getCurrentZonedDateTime().toLocalDateTime()`...this could blow up the diff further.

If you think it is still worth the effort perhaps it can be done as a separate task after the current changes are pushed out of the door?

eshryane: Thanks for checking, no it's OK for now, we can refactor at some other time

whois-scheduler/src/main/java/net/ripe/db/whois/scheduler/task/acl/ResetPersonalObjectAccounting.java
dadepo: probably best to have it run in UTC. Updated to this effect

whois-scheduler/src/main/java/net/ripe/db/whois/scheduler/task/acl/AutomaticPermanentBlocksCleanup.java
dadepo: probably best to have it run in UTC. Updated to this effect

whois-scheduler/src/main/java/net/ripe/db/whois/scheduler/task/acl/AutomaticPermanentBlocks.java
dadepo: probably best to have it run in UTC. Updated to this effect

whois-scheduler/src/main/java/net/ripe/db/whois/scheduler/task/export/DatabaseTextExport.java
dadepo: done

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageParser.java
dadepo: Although not an integration test, I believe this [here](https://github.com/RIPE-NCC/whois/blob/325e925b393584725073828409b4aa8fb79f346b/whois-api/src/test/java/net/ripe/db/whois/api/mail/dequeue/MessageParserTest.java#L96) covers that part of the code

eshryane: Perfect

eshryane (approved):

PR #812, opened 2021-04-23
Display an error on NRTM if client is blocked by ACL limit

eshryane:
* Do not write QueryMessage into Netty channel, there is no handler to turn it into a String.
* Replace generic IllegalArgumentException with application-specific NrtmException


eshryane:
Integration tests passed

PR #813, opened 2021-04-23
Switching to utc audit logs

dadepo:


dadepo:
Note I will only be applying the changes to files we have on file after this PR is approved.

dadepo:
Merging this to `switching-to-utc` to keep PR review and testing simpler.

PR #815, opened 2021-04-26
Custom healthcheck.

sbusk:


dadepo (approved):
LGTM. Only comment is maybe have some logging so it is obvious which check fails/passes

eshryane (approved):

sbusk:
> LGTM. Only comment is maybe have some logging so it is obvious which check fails/passes

Yep, good suggestion.

PR #817, opened 2021-05-07
Switching to utc version history

dadepo:


sbusk (approved):

pom.xml
```diff
@@ -4,7 +4,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>net.ripe.db</groupId>
     <artifactId>whois</artifactId>
-    <version>1.0-SNAPSHOT</version>
+    <version>utc-switch-version-2</version>
```
eshryane, line 7: Please revert this change before merging

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageParser.java
```diff
@@ -44,6 +44,8 @@
 import java.util.Enumeration;
 import java.util.List;
 
+import static java.time.format.DateTimeFormatter.RFC_1123_DATE_TIME;
```
eshryane, line 47: Can you update the Formatter in the `MessageParser.DATE_FORMAT` field instead? The format of RFC_1123_DATE_TIME is different in more than the timezone.

eshryane: Or use the standard 'UTC' formatter we're using elsewhere

dadepo: This seems to be the format the smtp server uses for the `Delivery-date` time. And it is used when parsing from the `Delivery-date` string representation to time in Java

pom.xml
dadepo: Noted. This branch has been merged to `switching-to-utc` which should contain all changes now. And yes, it will be reverted to the right version before merging to master

PR #830, opened 2021-06-10
use local database for Resource Holder Search

eshryane:


eshryane:
Integration tests passed

PR #845, opened 2021-06-15
sort property names when printed at startup

eshryane:


eshryane:
Tested using a oneshot script, and deployed internally.

PR #857, opened 2021-07-02
only start if timezone is utc

dadepo:


eshryane (approved):

whois-db/src/main/java/net/ripe/db/whois/db/WhoisServer.java
```diff
@@ -40,6 +41,11 @@ public WhoisServer(
     }
 
     public static void main(final String[] args) {
+
+        if (!ZoneId.systemDefault().equals(ZoneId.of("UTC"))) {
+            throw new RuntimeException(String.format("Illegal timezone: %s. Application timezone should be UTC", ZoneId.systemDefault()));
```
eshryane, line 46: Throw new IllegalStateExcepion

eshryane (commented):
Commented

whois-db/src/main/java/net/ripe/db/whois/db/WhoisServer.java
dadepo: done

PR #858, opened 2021-07-06
Update dependencies.

sbusk:


eshryane:
Is the AWS SDK dependency OK with the updated Netty? Do they need to be kept in sync?

eshryane:
> Is the AWS SDK dependency OK with the updated Netty? Do they need to be kept in sync?

Sander confirmed: "I'm not sure where the comment came from in the pom but we did have a problem at 1 point where the transitive netty dependency from the aws sdk and the netty version we specified were conflicting and causing problems, but we only have 1 version now so it's no longer an issue (i deployed to aws yesterday and verified it still works)"

eshryane (approved):

PR #869, opened 2021-07-26
making it possible to enable/disable instance from lb via jmx

dadepo:


tools/whois.init
```diff
@@ -167,12 +167,20 @@ case "$1" in
                         exit 1
                 fi
                 ;;
+        disable)
+          PID=$(pgrep -f "$COMMAND_REGEXP")
+          echo -e "bean net.ripe.db.whois:name=Loadbalancer && run setLoadbalancerEnabled false" | java -jar $JMXTERMPATH --url $PID
+          ;;
+        enable)
```
eshryane, line 174: Isn't ```enable``` automatic at startup? Or should the healthcheck page wait for this JMX command before reporting OK?

tools/whois.init
```diff
@@ -167,12 +167,20 @@ case "$1" in
                         exit 1
                 fi
                 ;;
+        disable)
```
eshryane, line 170: There's a naming collision with the existing ```disable``` in the internal init script. If the intention is to replace that altogether then that's OK.

whois-api/src/main/java/net/ripe/db/whois/api/rest/HealthCheckService.java
```diff
@@ -1,31 +1,49 @@
...
+ * in jmxterm, run with:
+ *      bean net.ripe.db.whois:name=Loadbalancer
+ *      run setLoadbalancerEnabled true|false
+ *
+ */
+
+
 @Path("/healthcheck")
-public class HealthCheckService {
+public class HealthCheckService extends JmxBase {
```
eshryane, line 46: This class is getting increasingly complex, is it time to spearate each ```HealthCheck``` using an interface?

tools/whois.init
dadepo: This script is only used for when running whois locally. I am not sure I know the internal init script you refer to, but I do not think there should be any collision or confusion...?

eshryane: I mean the DB team's ```/export/init.d/whois``` script, will these changes be synced there?

whois-api/src/main/java/net/ripe/db/whois/api/rest/HealthCheckService.java
dadepo: Wanted to keep the changes as minimal as possible. That is, just make the controller a managed bean. But I agree the individual health checks should probably move to their own classes. I'll update

tools/whois.init
dadepo: Nope. It won't. This script is so different from the ones in `/export/init.d/whois` I can only imagined it diverged a long time ago and it is only used when running locally. 

I mainly updated it just to make sure that using the jmxconsole via bash script works...I would still have to update the `/export/init.d/whois` to incorporate the new ability to disable/enable via jmx call. That would probably be a different PR on db-operational first, before changing it on staging server

whois-api/src/main/java/net/ripe/db/whois/api/rest/HealthCheckService.java
eshryane: I'm OK with leaving as-is for now also 👍

tools/whois.init
eshryane: OK! Thanks for explaining.

tools/whois.init
dadepo: I think not marking as enable automatically at start up could come in handy. Eg during manual testing scenarios, where one need to check stuff, before signifying to F5 to include the node into the pool. 

The deployment script would be updated to include a call that enables after start up.

eshryane: OK

eshryane (approved):

whois-api/src/main/java/net/ripe/db/whois/api/rest/HealthCheckService.java
dadepo: On another thought, I will add another class, which will keep the state whether to enable or disable from the LB and make this managed by JMX instead of the controller. This way it can also be usable from whois-internal

PR #888, opened 2021-09-01
Support IDN's by punycoding outgoing email to and reply-to addresses

sbusk:
Probably would be better to also rewrite the existing tests in MailGatewaySmtpTest to work with MailSenderStub?

### DIFF
--- a/whois-update/src/main/java/net/ripe/db/whois/update/mail/MailGatewaySmtp.java
+++ b/whois-update/src/main/java/net/ripe/db/whois/update/mail/MailGatewaySmtp.java
@@ -80,11 +80,11 @@ public class MailGatewaySmtp implements MailGateway {
     @RetryFor(value = MailSendException.class, attempts = 20, intervalMs = 10000)
     private void sendEmailAttempt(final String to, final String replyTo, final String subject, final String text) {
         try {
-            final String punyCodedTo = PunycodeConversion.toAscii(to);
-            final String puncyCodedReplyTo = !StringUtils.isEmpty(replyTo)?
-                    PunycodeConversion.toAscii(replyTo) : "";
-
             mailSender.send(mimeMessage -> {
+                final String punyCodedTo = PunycodeConversion.toAscii(to);
+                final String puncyCodedReplyTo = !StringUtils.isEmpty(replyTo)?
+                        PunycodeConversion.toAscii(replyTo) : "";
+
                 final MimeMessageHelper message = new MimeMessageHelper(mimeMessage, MimeMessageHelper.MULTIPART_MODE_NO, "UTF-8");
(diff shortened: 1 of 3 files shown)


whois-update/src/main/java/net/ripe/db/whois/update/mail/MailGatewaySmtp.java
```diff
@@ -82,22 +80,24 @@ public void sendEmail(final String to, final String subject, final String text,
...
-                    if (!StringUtils.isEmpty(replyTo))
-                        message.setReplyTo(replyTo);
-                    message.setSubject(subject);
-                    message.setText(text);
-
-                    mimeMessage.addHeader("Precedence", "bulk");
-                    mimeMessage.addHeader("Auto-Submitted", "auto-generated");
-
-                    loggerContext.log("msg-out.txt", new MailMessageLogCallback(mimeMessage));
+            final String punyCodedTo = PunycodeConversion.toAscii(to);
```
eshryane, line 83: Why are these lines outside the send() method?

whois-update/src/test/java/net/ripe/db/whois/update/mail/MailGatewaySmtpTest.java
```diff
@@ -83,6 +86,20 @@ public void sendResponseAndCheckForEmptyReplyTo() throws Exception {
...
 
+    @Test
+    public void checkRecipientAddressesArePunycoded() throws Exception {
+        MailSenderStub mailSenderStub = new MailSenderStub();
+        MailGatewaySmtp mailGatewaySmtp = new MailGatewaySmtp(loggerContext, mailConfiguration, mailSenderStub);
+        ReflectionTestUtils.setField(mailGatewaySmtp, "outgoingMailEnabled", true);
+
+        when(mailConfiguration.getFrom()).thenReturn("from@from.to");
+
+        mailGatewaySmtp.sendEmail("to@to.to", "subject", "test", "fredrik.egerstad@Åownit.se");
```
eshryane, line 97: Replace real email address with a test one

sbusk: Whoops

whois-update/src/main/java/net/ripe/db/whois/update/mail/MailGatewaySmtp.java
sbusk: Because I started off using the existing mocking setup type tests with argument captures but injecting the email did not work.. now I'm using the MailSenderStub and it works just fine so I've moved them into the send(), well spotted :)

eshryane (approved):

PR #896, opened 2021-09-09
updating health check endpoint before shutdown preventing need for jmx in start up script

dadepo:


whois-api/src/main/java/net/ripe/db/whois/api/httpserver/DelayShutdownHook.java
```diff
@@ -0,0 +1,46 @@
...
+    @Override
+    protected void doStart() throws Exception {
+        super.doStart();
+        readinessUpdater.up();
+    }
+
+    @Override
+    protected void doStop() throws Exception {
+        readinessUpdater.down();
+        Thread.sleep(preShutdownPause * 1000);
```
eshryane, line 43: Use ```Uninterruptibles.sleepUninterruptibly()``` or catch InterruptedException, log and re-throw

whois-db/src/main/java/net/ripe/db/whois/db/WhoisServer.java
```diff
@@ -84,6 +88,12 @@ public void start() {
     }
 
     public void stop() {
+        try {
+            Thread.sleep(preShutdownPause * 1000);
```
eshryane, line 92: Is this necessary? Will the WhoisServer not wait for Jetty to shutdown first anyway?

dadepo: Yes it turns out to be necessary, and no unfortunately the WhoisServer does not wait for Jetty to shutdown first. Not sure if there is another way to enforce this

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/DelayShutdownHook.java
dadepo: Updated

whois-db/src/main/java/net/ripe/db/whois/db/WhoisServer.java
eshryane: OK, can you add a TODO explaining why this sleep is there. Also I suggest adding an extra delay here, so there isn't a race condition with the DelayShutdownHook (if the delay is the same, the WhoisServer may exit first?). Also log here if the wait is interrupted.

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/DelayShutdownHook.java
eshryane: Thanks! I suggest updating WhoisServer to match (I don't know if interrupting the thread is possible during shutdown, this will prevent it)

whois-db/src/main/java/net/ripe/db/whois/db/WhoisServer.java
dadepo: I added some explanatory texts. I do not think there is a need to add an extra delay. There won't be a race condition; worst case scenario is the health check endpoint stays available for couple of milliseconds/seconds shy of the actual delay.

eshryane: OK! Thanks

eshryane (approved):

PR #899, opened 2021-09-13
sentry configurations

maggarwal13:
There are some changes in log4j2.xml file as well.

### DIFF
--- a/whois-commons/src/main/java/net/ripe/db/whois/common/SentryConfigurations.java
+++ b/whois-commons/src/main/java/net/ripe/db/whois/common/SentryConfigurations.java
@@ -17,7 +17,7 @@ public class SentryConfigurations {
 
     private final ApplicationVersion applicationVersion;
     private final String sentryDsn;
-    final String environment;
+    private final String environment;
 
     @Autowired
     public SentryConfigurations(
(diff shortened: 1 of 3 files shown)


whois-commons/src/main/java/net/ripe/db/whois/common/SentryConfigurations.java
```diff
@@ -0,0 +1,44 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(SentryConfigurations.class);
+
+    private final ApplicationVersion applicationVersion;
+    private final String sentryDsn;
+    private final String environment;
+
+    @Autowired
+    public SentryConfigurations(
+            @Value("${sentry.dsn:}") final String sentryDsn,
+            @Value("${sentry.environment:}") final String environment,
```
eshryane, line 25: `sentry.environment` is optional, should you check for empty string and return if not defined? Or should we assume a default value (such as production)?

maggarwal13: Ok I will add a check I guess making default as production is error prone as if someone forgets to add environment property in predev errors will be reported as production then

maggarwal13: Done

eshryane: Thanks!

eshryane (approved):

PR #915, opened 2021-09-24
being more explicit on how much shedlock's lock should last. Increased threads in thread pool

dadepo:


dadepo:
In `whois`, we have 15 `@Scheduled` jobs, 8 of which uses `shedlock` and 6 configured to run at midnight. 
We initially set thread pool size to 10.

In `whois-internal` on the other hand, we have 10 `@Scheduled` jobs, 8 of which uses `shedlock` with 5 configured to run at midnight. Thread pool size for `whois-internal` is 24.

Hence `whois` is running more scheduled jobs than `whois-internal` but it is configured with less amount of threads.

Theory is that, due to the number of jobs being scheduled, and available threads (ie 10) there is a chance between the time the task is put on the queue and the time it is executed for the same task to be put in the queue on another machine. And since the lock is released immediately a task finishes, if it takes some time for identical task to be pulled out of the pool on another machine, then the same task can be run again.

The changes in this PR seek to improve the situation by making the following explicit changes:

1. For tasks configured to run at midnight, have the lock set for at least 7hrs. This will ensure enough time to prevent any other subsequent run of the job.
2. Increase the pool capacity from 10 to 24

whois-scheduler/src/main/java/net/ripe/db/whois/scheduler/SchedulerConfig.java
```diff
@@ -33,7 +28,7 @@
     @Bean("scheduler")
     public TaskScheduler taskScheduler() {
         ThreadPoolTaskScheduler scheduler = new ThreadPoolTaskScheduler();
-        scheduler.setPoolSize(10);
+        scheduler.setPoolSize(20);
```
eshryane, line 31: 20 or 24?

eshryane: Any idea why scheduled jobs at midnight only run sequentially (one at a time)?

whois-scheduler/src/main/java/net/ripe/db/whois/scheduler/task/acl/ResetPersonalObjectAccounting.java
```diff
@@ -17,8 +17,8 @@ public ResetPersonalObjectAccounting(final PersonalObjectAccounting personalObje
     }
 
     @Override
-    @Scheduled(cron = "0 0 0 * * *")
-    @SchedulerLock(name = "ResetPersonalObjectAccounting")
+    @Scheduled(cron = "@midnight")
+    @SchedulerLock(name = "ResetPersonalObjectAccounting", lockAtLeastFor = "23h")
```
dadepo, line 21: As suggested by @sbusk , this is now set to 23h (instead of 7h) to be explicit about the requirement to have these jobs only run once in space of 24hrs. He also suggested to have this in whois-internal...just to be really sure, even though we do not have the double job problem there...any thoughts @eshryane

eshryane: My only concern is whether this will prevent us from running those scheduled jobs during the day via JMX. Can you test to find out? If it's an issue we'll need a workaround (such as manually resetting the timestamp in the DB table beforehand).

dadepo: Done for whois-internal here https://github.com/RIPE-NCC/whois-internal/pull/434/commits/40ce8a6366ee9dbd612158d99408687e3bd3ec43

sbusk: > My only concern is whether this will prevent us from running those scheduled jobs during the day via JMX. Can you test to find out? If it's an issue we'll need a workaround (such as manually resetting the timestamp in the DB table beforehand).

It will definitely prevent you from running the job again during the day via JMX. The workaround is to update the lock time in the database but beware: shedlock also maintains a lock time cache in memory so you may also have to restart the node. Maybe we want to override the lock time in the runDailyScheduledTask JMX action so we can force a job to run when we need it. (it will currently silently refuse to run the job).

dadepo: I reduced this from 23h to 1h since the jobs should now run in parallel...I could remove it totally as it was before, but I think 1hr is safe enough and should not block being able to run jmx job manually

sbusk: I'd either leave it at 23h with a solid explanation as to why (to safeguard against a job running more than 1 time per day for whatever reason) or leave it out completely so it's the default (which is more intuitive).
Setting it to 1h doesn't prevent the job from running more than once per day and also differs from the default so it's confusing as to where the 1h comes from.

dadepo: I'll go with leaving it out. Having it as 23h essentially nullifies the ability to run the jobs via JMX.

dadepo:
PS Do not merge until confirmation that jobs still run with all the new modifications

whois-db/src/main/resources/applicationContext-whois-base.xml
```diff
@@ -11,8 +11,7 @@
     <context:mbean-export/>
 
     <context:annotation-config />
-    <bean class="net.ripe.db.whois.scheduler.SchedulerConfig" />
-
     <import resource="applicationContext-nrtm.xml"/>
+    <bean class="net.ripe.db.whois.scheduler.SchedulerConfig" />
```
sbusk, line 15: You didn't like it being on line 14? ;)

whois-query/src/test/java/net/ripe/db/whois/query/integration/SimpleKeepaliveTestIntegration.java
```diff
@@ -22,6 +23,7 @@
 
 @Category(IntegrationTest.class)
 @DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_EACH_TEST_METHOD)
+@Ignore
```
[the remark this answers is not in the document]
dadepo: Nope. Forgot it there after troubleshooting...Removed now!

whois-db/src/main/resources/applicationContext-whois-base.xml
dadepo: haha...while trying to figure out what could be going wrong...I moved the configurations around a bit to match whois-internal :)

sbusk: Ok :)

PR #918, opened 2021-09-29
PGP-signed message with base64 encoded content

eshryane:
We don't support it yet, but we should - as Thunderbird does this by default now.

### DIFF
--- a/whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/integration/SignedMessageIntegrationSpec.groovy
+++ b/whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/integration/SignedMessageIntegrationSpec.groovy
@@ -3920,4 +3920,70 @@ class SignedMessageIntegrationSpec extends BaseWhoisSourceSpec {
       ack =~ "Create SUCCEEDED: \\[person\\] FP1-TEST   First Person"
   }
 
+  @Ignore("TODO: [ES] We do not support base64 encoded content yet")
+  def "pgp signed message with base64 encoded content"() {
+    given:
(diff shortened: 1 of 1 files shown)


whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/integration/SignedMessageIntegrationSpec.groovy
```diff
@@ -3920,4 +3920,70 @@ class SignedMessageIntegrationSpec extends BaseWhoisSourceSpec {
       ack =~ "Create SUCCEEDED: \\[person\\] FP1-TEST   First Person"
   }
 
+  @Ignore("TODO: [ES] We do not support base64 encoded content yet")
```
dadepo, line 3923: Suggestion: Maybe assert this currently fails, so the test passes? so when this is implemented, then the test will fail, which will then be fixed. 

This prevents a situation, whereby, if the fix is implemented by another developer apart from you, a duplicated test case can be added, since there is a chance they did not know this ignored test existed.

dadepo (approved):

whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/integration/SignedMessageIntegrationSpec.groovy
eshryane: I prefer to `@Ignore` the failing test, then remove the annotation when implementing, then the test itself doesn't have to be changed, it will pass once the change is implemented.

Hopefully we'll get to resolve this in the short term (there's an existing Jira issue to track it).

PR #934, opened 2021-11-02
Lucene 8 10

sbusk:


jbartolini (approved):

pom.xml
```diff
@@ -74,7 +74,7 @@
         <jackson.version>2.12.5</jackson.version>
 
         <log4j.version>2.14.1</log4j.version>
-        <lucene.version>7.7.3</lucene.version>
+        <lucene.version>8.9.0</lucene.version>
```
eshryane, line 77: Add a comment above that the dependency version must be synchronised with Elasticsearch

whois-api/src/test/java/net/ripe/db/whois/api/fulltextsearch/FullTextAnalyzerTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * The {@link org.apache.lucene.analysis.Analyzer} used in whois does *not* split on punctuation at the moment.
+ * This has been done to support finding objects based on exact key containing interpunction.
+ * Downside of not stripping the punctuation is that the index is blown up considerably because for example we store both
+ * "members" and "members:" in the index.
+ * We probably want to implement a filter that strips most punctuation but also emits punctuation that is part of an object's key.
+ * This test has been added to more quickly find the cause of changes in fulltext search results as a result of Lucene upgrades or Analyzer changes :)
+ */
+public class FullTextAnalyzerTest {
```
eshryane, line 25: This test class is not related to the Lucene 8 update? Remove from this PR?

sbusk: It is something I have added as a result of losing time figuring out why the behaviour was different after a Lucene major version bump, trying to use the StandardAnalyzer instead but then figuring out that we can't and I want to have it documented and under test for the next time we do a major version upgrade and run into the same problem. So IMHO yes it is related to the Lucene 8 upgrade.

pom.xml
sbusk: Good one, will do.

whois-api/src/test/java/net/ripe/db/whois/api/fulltextsearch/FullTextAnalyzerTest.java
eshryane: OK! Thanks. I confused this with Jacopo's changes for the logsearch analyser, but they are not related.

eshryane (approved):

PR #940, opened 2021-11-11
Do not run authoritative resource tasks by default during tests

eshryane:
Scheduled authoritative resource import tasks were sometimes run unexpectedly during integration tests, causing them to fail randomly.


eshryane:
Integration tests passed (3f3eb4d)

maggarwal13 (approved):

PR #942, opened 2021-11-12
Support base64 encoding

maggarwal13:
Thunderbird 78+ is using OpenPGP to sign mail messages, rather than the Enigmail plugin.

The default for signed messages is now to base64 encode the message content, which we need to support.


whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageParser.java
```diff
@@ -55,6 +56,7 @@
     private static final Logger LOGGER = LoggerFactory.getLogger(MessageParser.class);
 
     private static final DateTimeFormatter DATE_FORMAT = DateTimeFormatter.ofPattern("EEE MMM dd HH:mm:ss z yyyy");
+    public static final String HEADER_BASE_64 = "Content-Transfer-Encoding: base64";
```
eshryane, line 59: This is both the header and the value, maybe define each separately (also it's not guaranteed there will only be one space separating them).

maggarwal13: Changed it to use pattern matching.

eshryane: 👍

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageParser.java
```diff
@@ -328,4 +353,13 @@ void addCredential(final Credential credential) {
             credentials.add(credential);
         }
     }
+
+    private String decodeBase64Encoding(final String paragraph) {
+        try {
+            return new String(new Base64().decode(paragraph.getBytes()));
+        } catch (Exception e) {
+            LOGGER.info("failed to decode base64 for ");
```
eshryane, line 361: Missing ```e.getClass().getName()``` and ```e.getMessage()``` from LOGGER statement?

eshryane (commented):
commented

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageParser.java
maggarwal13: Yes thanks for pointing it out, I somehow missed it

eshryane (approved):

PR #955, opened 2021-12-09
Require enduser abusec

sbusk:


whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/common/EndUserAbuseCValidator.java
```diff
@@ -0,0 +1,70 @@
...
+
+    private final RpslObjectDao rpslObjectDao;
+
+    @Autowired
+    public EndUserAbuseCValidator(final RpslObjectDao rpslObjectDao) {
+        this.rpslObjectDao = rpslObjectDao;
+    }
+
+    @Override
+    public void validate(final PreparedUpdate update, final UpdateContext updateContext) {
```
eshryane, line 34: Do we need to check the inetnum STATUS? or are *all* status values included? 

The Abuse-c Validation policy [ripe-705](https://www.ripe.net/publications/docs/ripe-705) only refers to direct allocated inetnums needing abuse-c, and we only check certain end user inetnum statuses in our abuse-c validation code.

eshryane (approved):

PR #956, opened 2021-12-09
Domain overlap validator

sbusk:


whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/update/DomainAuthSpec.groovy
```diff
@@ -3,8 +3,9 @@ package net.ripe.db.whois.spec.update
 
 import net.ripe.db.whois.spec.BaseQueryUpdateSpec
 import net.ripe.db.whois.spec.domain.AckResponse
+import org.junit.jupiter.api.Tag
 
-@org.junit.jupiter.api.Tag("IntegrationTest")
+@Tag("IntegrationTest")
```
maggarwal13, line 8: The reason there is fully qualified class name is because we have a Tag class of our own. But I guess since this is not suing that so nice to use the short one

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/domain/DomainIntersectionValidator.java
```diff
@@ -0,0 +1,79 @@
...
+                .orElse(Ipv4Resource.parse("0/0"));
+
+        Ipv4Resource firstIntersecting = null;
+        final List<Ipv4Entry> childEntries = ipv4DomainTree.findFirstMoreSpecific(parentInterval);
+        for (final Ipv4Entry childEntry : childEntries) {
+            final Ipv4Resource child = childEntry.getKey();
+
+            if (child.intersects(ipv4Resource) && !(child.contains(ipv4Resource) || ipv4Resource.contains(child))) {
+                if (firstIntersecting == null || firstIntersecting.singletonIntervalAtLowerBound().compareUpperBound(child.singletonIntervalAtLowerBound()) > 0) {
+                    firstIntersecting = child;
```
maggarwal13, line 60: Personal preference: I would rather  updateContext here and break it then let it loop through whole dataset.

maggarwal13: so initially firstIntersecting will always be null and this if loop will resolve to true , so the next if condition is  irrelevant?

sbusk: This was freely inspired/shamelessly borrowed from the already existing IntersectionValidator but I guess was caught in the act and will properly scrutinise the actual logic ;)

whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/update/DomainAuthSpec.groovy
sbusk: Unless there are clashes I see no reason to litter the code with fully qualified class imports..

whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/update/DomainAuthSpec.groovy
```diff
@@ -3207,4 +3219,42 @@ class DomainAuthSpec extends BaseQueryUpdateSpec {
...
+
+        then:
+        def ack = new AckResponse("", message)
+
+        ack.summary.nrFound == 1
+        ack.summary.assertSuccess(0, 0, 0, 0, 0)
+        ack.summary.assertErrors(1, 1, 0, 0)
+        ack.countErrorWarnInfo(1, 0, 0)
+        ack.errors.any { it.operation == "Create" && it.key == "[domain] 191-193.0.0.193.in-addr.arpa" }
+        ack.errorMessagesFor("Create", "[domain] 191-193.0.0.193.in-addr.arpa") ==
```
eshryane, line 3254: We need to report the overlapping reverse delegation and not the (forward) IPv4 prefix, otherwise it's really confusing for the user - they need to reverse the prefix again to identify the overlapping domain

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/domain/DomainIntersectionValidator.java
```diff
@@ -0,0 +1,79 @@
...
+                }
+            }
+        }
+
+        if (firstIntersecting != null) {
+            updateContext.addMessage(update, UpdateMessages.intersectingRange(firstIntersecting));
```
eshryane, line 66: Report the reverse delegation (i.e. the domain key) of the overlap, not the IPv4 prefix

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/domain/DomainIntersectionValidator.java
sbusk: Simplified :)

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/domain/DomainIntersectionValidator.java
sbusk: Turns out it wasn't necessary, I could lookup the intersecting domain based on the object id from the tree entry.

whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/update/DomainAuthSpec.groovy
sbusk: Done.

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/domain/DomainIntersectionValidator.java
eshryane: Great! Thanks

eshryane (approved):

PR #988, opened 2022-02-10
[DB-3538] Better Error Message when Changing "mnt-by:" on Allocations

jbartolini:
Changed error message. Removed word from other error message to reflect real situation. Fixed tests.


eshryane (approved):
Please wait for integration tests to pass before merging

PR #1003, opened 2022-04-05
Remove WHITEPAGES org type

eshryane:
Following decision on DB-WG:
https://www.ripe.net/ripe/mail/archives/db-wg/2022-March/007316.html

Remove the Whitepages facility from Whois:
https://www.ripe.net/manage-ips-and-asns/db/support/documentation/white-pages


eshryane:
Integration tests have passed

PR #1014, opened 2022-05-12
invalid object type in history

maggarwal13:


whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticFullTextIndex.java
```diff
@@ -120,6 +120,15 @@ protected void update() throws IOException {
         elasticIndexService.updateMetadata(new ElasticIndexMetadata(end, source));
     }
 
+    private SerialEntry getSerialEntry(int serial) {
+        try {
+            return JdbcRpslObjectOperations.getSerialEntry(jdbcTemplate, serial);
+        } catch (Exception e) {
+            LOGGER.debug("SerialDao.getById({})", serial, e);
```
eshryane, line 127: Can you make the message a bit more descriptive, such as `"Caught exception reading serial {} from the database", serial, e`

eshryane: Also add `, ignoring` to be clear we didn't do anything about it

PR #1022, opened 2022-06-15
update jackson

maggarwal13:


pom.xml
```diff
@@ -67,7 +67,7 @@
         <junit.version>5.8.2</junit.version>
         <jetty.version>10.0.7</jetty.version>
         <jersey.version>2.35</jersey.version>
-        <jackson.version>2.12.6</jackson.version>
+        <jackson.version>2.13.2</jackson.version>
```
eshryane, line 70: Can you update to 2.13.3 ?

maggarwal13: Ok but in that case whois-internal jackson dependency also needs to be updated whenever we update whois depenedency

eshryane: OK we can do it some other time. This PR is 👍

eshryane (approved):

PR #1052, opened 2022-07-15
Add comment not noop

maggarwal13:
Currently, changes to comments in Whois updates are considered a NOOP (i.e. the value of the attribute has not changed).

Now allow adding, modifying or removing a comment to count as an update and not a NOOP (i.e. a change to the comment is considered equivalent to changing the attribute value).

### DIFF
--- a/whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/integration/OverrideIntegrationSpec.groovy
+++ b/whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/integration/OverrideIntegrationSpec.groovy
@@ -263,6 +263,26 @@ class OverrideIntegrationSpec extends BaseWhoisSourceSpec {
                 "***Info:    Authorisation override used")
     }
 
+    def "comment is not a noop for whois update"() {
+        given:
+        def data = fixtures["ORG1"].stripIndent() + "override:denis,override1"
+        data = (data =~ /address:      street 5/).replaceFirst("address:      street 5 #test comment noop")
+
+        def update = new SyncUpdate(data: data)
+
+        when:
+        def result = syncUpdate update
+
+        then:
+        !result.contains("Warning: Submitted object identical to database object\n" +
+                "\n" +
+                "***Info:    Authorisation override used")
+        result.contains("Modify SUCCEEDED: [organisation] ORG-TOL1-TEST\n" +
+                "\n" +
+                "\n" +
+                "***Info:    Authorisation override used")
+    }
+
     def "override is noop on case-sensitive change with update-on-noop set to false"() {
         given:
         def data = fixtures["ORG1"].stripIndent() + "override:denis,override1, {update-on-noop=false}"
--- a/whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/integration/OverrideIntegrationSpec.groovy
+++ b/whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/integration/OverrideIntegrationSpec.groovy
@@ -51,7 +51,7 @@ class OverrideIntegrationSpec extends BaseWhoisSourceSpec {
             organisation: ORG-TOL1-TEST
             org-name:     Test Organisation Ltd
             org-type:     OTHER
-            descr:        test org
+            descr:        test org #comment
             address:      street 5
             e-mail:       org1@test.com
             mnt-ref:      TST-MNT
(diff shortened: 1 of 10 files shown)


whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/integration/OverrideIntegrationSpec.groovy
```diff
@@ -263,43 +263,70 @@ class OverrideIntegrationSpec extends BaseWhoisSourceSpec {
...
+        !result.contains("Warning: Submitted object identical to database object\n" +
+                "\n" +
+                "***Info:    Authorisation override used")
+        result.contains("Modify SUCCEEDED: [organisation] ORG-TOL1-TEST\n" +
+                "\n" +
                 "\n" +
                 "***Info:    Authorisation override used")
     }
 
-    def "override on case-sensitive change with update-on-noop set to true"() {
```
eshryane, line 282: Was this test removed?

maggarwal13: No, test case still exists, I think I changed few lines above so somehow it shows deleted

whois-rpsl/src/test/java/net/ripe/db/whois/common/rpsl/RpslObjectTest.java
```diff
@@ -386,20 +386,22 @@ public void compareRpslObjects_single() {
         assertThat(RpslObject.parse("mntner: \t  mnt  \t \n \t     a"), is(RpslObject.parse("mntner: mnt\n\ta")));
         assertThat(RpslObject.parse("mntner: \t  one \t two \t \n \t     a"), is(RpslObject.parse("mntner: one two\n\ta")));
         assertThat(RpslObject.parse("mntner:one \t two \t \n \t     a"), is(RpslObject.parse("mntner: one two\n\ta")));
-        assertThat(RpslObject.parse("mntner: mnt # comment"), is(RpslObject.parse("mntner: mnt")));
+        assertThat(RpslObject.parse("mntner: mnt # comment"), is(RpslObject.parse("mntner: mnt # comment")));
         assertThat(RpslObject.parse("mntner: mnt # comment"), is(RpslObject.parse("mntner: mnt # comment")));
         assertThat(RpslObject.parse("mntner: mnt# comment"), is(RpslObject.parse("mntner: mnt # comment")));
         assertThat(RpslObject.parse("mntner: mnt # com  ment"), is(RpslObject.parse("mntner: mnt # com ment")));
         assertThat(RpslObject.parse("mntner: mnt # com  ment"), is(RpslObject.parse("mntner: mnt #com ment")));
+/*
```
eshryane, line 394: Change commented test to `is(not())` ?

whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/integration/OverrideIntegrationSpec.groovy
maggarwal13: Ahh, I added new tests cases but since lot of lines are similar it treated it as changed , look a. little bit down you will find this test as an addition

whois-rpsl/src/test/java/net/ripe/db/whois/common/rpsl/RpslObjectTest.java
maggarwal13: I commented it because I am not sure if tests is correct. How come   "mntner: mnt two three four" should be equal to "mntner: mnt # one\n two\n+ three\n\tfour". I think this test scenario is wrong and should be deleted

eshryane: That test *was* correct:
`        assertThat(RpslObject.parse("mntner: mnt two three four"), is(RpslObject.parse("mntner: mnt # one\n two\n+ three\n\tfour")));`
The value should be the same regardless of formatting. I suggest just removing the `# one` comment as you're now including this in the equals check, and just test the attribute value comparison.
Or, test the value *and* comment are the same, regardless of formatting 
```
assertThat(RpslObject.parse("mntner: mnt two three four # one"), is(RpslObject.parse("mntner: mnt # one\n two\n+ three\n\tfour # one")));
```

maggarwal13: dded and chenged

eshryane (approved):

PR #1053, opened 2022-07-18
make es.hosts as optional

maggarwal13:


whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticIndexService.java
```diff
@@ -48,23 +48,33 @@ public class ElasticIndexService {
...
                                @Value("${elastic.commit.index:metadata}") final String whoisMetadataIndexName) {
-        this.whoisIndex = whoisIndexName;
+        this.whoisAliasIndex = whoisAliasName;
         this.metadataIndex = whoisMetadataIndexName;
-        final RestClientBuilder clientBuilder = RestClient.builder(elasticHosts.stream().map( (host) -> HttpHost.create(host)).toArray(HttpHost[]::new));
-        client = new RestHighLevelClient(clientBuilder);
+        getEsClient(elasticHosts);
+    }
+
+    private void getEsClient(final List<String> elasticHosts) {
```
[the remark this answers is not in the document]
eshryane: And make Nullable

whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticIndexService.java
```diff
@@ -48,23 +48,33 @@ public class ElasticIndexService {
...
+        } catch (Exception e) {
+            LOGGER.warn("Failed to tsrat the ES client {}", e.getMessage());
+        }
     }
 
     @PreDestroy
     public void preDestroy() throws IOException {
-        client.close();
+        if (isElasticRunning()) {
+            client.close();
```
eshryane, line 76: Check if client is null (uninitialised)

whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticIndexService.java
```diff
@@ -48,23 +48,33 @@ public class ElasticIndexService {
...
+            client = new RestHighLevelClient(clientBuilder);
+        } catch (Exception e) {
+            LOGGER.warn("Failed to tsrat the ES client {}", e.getMessage());
+        }
     }
 
     @PreDestroy
     public void preDestroy() throws IOException {
-        client.close();
+        if (isElasticRunning()) {
```
eshryane, line 75: Maybe add `(client != null)` check to `isElasticRunning()` and call that everywhere

maggarwal13: IsElasticRunning is catching all the exception and returning false. So if the  client is null it will return false

whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticIndexService.java
maggarwal13: Added a null check as well . So now it checks if client is not null and also it is reachable

whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticIndexService.java
```diff
@@ -86,34 +99,54 @@ public boolean isEnabled() {
...
         request.setQuery(QueryBuilders.matchAllQuery());
 
         client.deleteByQuery(request, RequestOptions.DEFAULT);
     }
 
-    public long getWhoisDocCount() throws IOException {
-        final CountRequest countRequest = new CountRequest(whoisIndex);
+    protected long getWhoisDocCount() throws IOException {
+        if (!isElasticRunning()) {
+            throw new IOException("ES is not running");
```
eshryane, line 135: I suggest using `IllegalStateException` as a more generic (and unchecked) exception, this situation is not an IOException. However, IOException maybe fine too, if the caller is already handling it (rather than adding a new catch).

eshryane: 👍

eshryane (approved):

PR #1055, opened 2022-07-21
Use local datasource for References service

eshryane:


eshryane:
Integration tests passed cc00752

eshryane:
Now testing

eshryane:
Integration tests passed

whois-api/src/main/java/net/ripe/db/whois/api/rest/ReferencesService.java
```diff
@@ -105,17 +106,16 @@ public class ReferencesService {
     private final String dummyRole;
     @Autowired
     public ReferencesService(
-            final RpslObjectDao rpslObjectDao,
-            final RpslObjectUpdateDao rpslObjectUpdateDao,
+            @Qualifier("jdbcRpslObjectSlaveDao") final RpslObjectDao rpslObjectDao,
+            final ReferencesDao referencesDao,
```
maggarwal13, line 110: If we are using jdbcRpslObjectSlaveDao , we can also inject ReferencesSlaveDao ? If not we can create it in a same way like jdbcRpslObjectSlaveDao?

PR #1056, opened 2022-07-28
Hazelcast shutdown

maggarwal13:
By default hazelcast.shutdownhook.enabled is true which disables the graceful shutdown so manually setting it to false

### DIFF
--- a/whois-query/src/main/java/net/ripe/db/whois/query/hazelcast/HazelcastInstanceManager.java
+++ b/whois-query/src/main/java/net/ripe/db/whois/query/hazelcast/HazelcastInstanceManager.java
@@ -60,7 +60,9 @@ public class HazelcastInstanceManager {
                 .setProperty("hazelcast.phone.home.enabled", "false")
                 .setProperty("hazelcast.memcache.enabled","false")
                 .setProperty("hazelcast.redo.giveup.threshold","10")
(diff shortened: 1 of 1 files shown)


whois-query/src/main/java/net/ripe/db/whois/query/hazelcast/HazelcastInstanceManager.java
```diff
@@ -60,7 +62,11 @@ private Config getGenericConfig() {
                 .setProperty("hazelcast.phone.home.enabled", "false")
                 .setProperty("hazelcast.memcache.enabled","false")
                 .setProperty("hazelcast.redo.giveup.threshold","10")
-                .setProperty("hazelcast.logging.type","slf4j");
+                .setProperty("hazelcast.logging.type","slf4j")
+                .setProperty("hazelcast.shutdownhook.enabled","false")
+                .setProperty("hazelcast.graceful.shutdown.max.wait","60");
+
+        config.setCPSubsystemConfig(new CPSubsystemConfig().setPersistenceEnabled(false));
```
eshryane, line 69: You could also use `config.getCPSubsystemConfig().setPersistenceEnabled(false);`

eshryane (approved):

PR #1057, opened 2022-07-28
Lucene85

eshryane:


eshryane:
Replaced Lucene with Elasticsearch

PR #1059, opened 2022-08-04
Remove whois tags

maggarwal13:


whois-api/src/test/java/net/ripe/db/whois/api/autocomplete/AutocompleteServiceTestIntegration.java
```diff
@@ -33,7 +34,7 @@
 import static org.hamcrest.Matchers.is;
 import static org.junit.jupiter.api.Assertions.fail;
 
-@org.junit.jupiter.api.Tag("IntegrationTest")
+@Tag("IntegrationTest")
```
eshryane, line 37: This change would be better in a separate branch. Could you create a separate PR for that? It would reduce the changes in this PR a lot (and you could merge that one first).

whois-commons/src/main/resources/patch/whois-1.103.1.sql
```diff
@@ -0,0 +1,3 @@
+-- delete tags functionality from whois
+
+DROP TABLE IF EXISTS `tags`;
```
eshryane, line 3: The *current* whois release is 1.103, please use the *next* Whois release 1.104 for the patch file

whois-api/src/test/java/net/ripe/db/whois/api/autocomplete/AutocompleteServiceTestIntegration.java
maggarwal13: I think this refactoring is related to this PR. I committed in one go with lot of other changes so I need to revert a lot and earlier I had to use fully qualified name as there was another Tag class, now its deleted with this PR. Its just the number of classes else its a one liner change

maggarwal13: But if you still insist on creating another PR let me know I can spend some more time to separate them

eshryane: Thanks for explaining @maggarwal13 I agree with you, keep everything in one PR the changes are related.

whois-commons/src/main/resources/patch/whois-1.103.1.sql
maggarwal13: renamed to 1.104

MiguelAHM (approved):

eshryane (approved):

PR #1061, opened 2022-08-04
Streaming marshal is interface ipo abstract

petchells:
Removed `AbstractStreamingMarshal` in favour of adding default methods to the `StreamingMarshal` interface


eshryane (approved):
Thanks Etch

whois-api/src/main/java/net/ripe/db/whois/api/rest/marshal/StreamingMarshalTextPlain.java
```diff
@@ -32,26 +45,42 @@ public <T> void writeArray(T t) {
...
         try {
+            outputStreamWriter.write(t.toString());
             outputStreamWriter.close();
         } catch (IOException e) {
             throw new StreamingException(e);
         }
     }
+
     @Override
-    public <T> void throwNotFoundError(HttpServletRequest request, List<Message> errorMessages) {
```
MiguelAHM, line 43: I think that you don't need to override this method because you already have the default method with same code in the interface, right?

MiguelAHM: same with createErrorStringEntity and createErrorStringMessages

MiguelAHM: I pull the branch in my local and I see that those methods are different.

MiguelAHM (approved):

PR #1062, opened 2022-08-08
Support the ".txt" extension in REST API requests to override the Accept: header as "text/plain"

eshryane:
This is useful for scripts or for hyperlinks to specify the response type, without needing to separately set an Accept header.

If a supported extension is present, it will overwrite any Accept header as it takes priority.


MiguelAHM (approved):

eshryane:
Integration tests have passed

PR #1067, opened 2022-08-19
feat: add ipv4 and ipv6 name server check

MiguelAHM:


whois-api/src/test/java/net/ripe/db/whois/api/rest/DomainObjectServiceIT.java
```diff
@@ -24,7 +25,7 @@
 import static org.junit.jupiter.api.Assertions.fail;
 
 @Tag("IntegrationTest")
-public class DomainObjectServiceTestIntegration extends AbstractIntegrationTest {
+public class DomainObjectServiceIT extends AbstractIntegrationTest {
```
maggarwal13, line 28: We usually follow a convention of putting IntegrationTest can you rename it to this?

whois-api/src/main/java/net/ripe/db/whois/api/rest/DomainObjectService.java
```diff
@@ -63,14 +64,17 @@ public class DomainObjectService {
     private final WhoisObjectMapper whoisObjectMapper;
     private final LoggerContext loggerContext;
 
+    private final ServiceInputChecker servicesInputChecker;
```
maggarwal13, line 67: Can you rename to NsServerInputChecker? ServiceInputChecker is ambiguous

whois-api/src/main/java/net/ripe/db/whois/api/rest/DomainObjectService.java
```diff
@@ -63,14 +64,17 @@ public class DomainObjectService {
...
+
     @Autowired
     public DomainObjectService(
             final WhoisObjectMapper whoisObjectMapper,
             final InternalUpdatePerformer updatePerformer,
-            final LoggerContext loggerContext) {
+            final LoggerContext loggerContext, final ServiceInputChecker serviceInputChecker) {
```
maggarwal13, line 73: Can you add a BusinessRuleValidator for this? Check any implementation of this interface.

maggarwal13 (changes requested):
I think this logic should  be best placed in BusinessRuleValidator.  Check DomainIntersectionValidator for inspiration.

whois-api/src/main/java/net/ripe/db/whois/api/rest/DomainObjectService.java
maggarwal13: There is already a NServerValidator business rule. Can you have a look

whois-api/src/main/java/net/ripe/db/whois/api/rest/checks/ServiceInputChecker.java
```diff
@@ -0,0 +1,46 @@
...
+                throw new IllegalArgumentException("Is not allowed to use that prefix with ns.ripe.net name server");
+            }
+        }
+    }
+
+    private boolean isIpv4(RpslObject rpslObject){
+        return rpslObject.getAttributes().stream().anyMatch(attribute -> {
+            assert attribute.getType() != null;
+            return attribute.getType().equals(AttributeType.DOMAIN)
+                    && !attribute.getValue().contains("ip6");
```
maggarwal13, line 27: Also use getValueForAttribute or findAttribute instead of iterating over

whois-api/src/main/java/net/ripe/db/whois/api/rest/checks/ServiceInputChecker.java
```diff
@@ -0,0 +1,46 @@
...
+        });
+    }
+    private boolean hasIncorrectPrefixes(RpslObject rpslObject, boolean isIpv4) {
+        return rpslObject.getAttributes().stream().anyMatch(attribute -> {
+            assert attribute.getType() != null;
+            return attribute.getType().equals(AttributeType.DESCR)
+                    && ((!attribute.getValue().contains("/32") && !isIpv4) || (!attribute.getValue().contains("/16") && isIpv4));
```
maggarwal13, line 34: attribute.getType().equals(AttributeType.DESCR), Is it a copy paste mistake or we are actually checking descr value? Also use RpslObject variable ObjectType for checking if its  ipv4 or ipv6

whois-api/src/main/java/net/ripe/db/whois/api/rest/checks/ServiceInputChecker.java
maggarwal13: Also use RpslObject variable ObjectType for checking if its ipv4 or ipv6

whois-api/src/main/java/net/ripe/db/whois/api/rest/checks/ServiceInputChecker.java
MiguelAHM: object type is going to be Domain always here, I think that we can not use this type to know if it is an ipv4 or ipv6 domain

MiguelAHM: on the other hand we need to know the prefix. this appears in the description. This is why I am using description, to check the prefix

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/domain/DomainIntersectionValidator.java
```diff
@@ -41,10 +45,11 @@ public DomainIntersectionValidator(final Ipv4DomainTree ipv4DomainTree,
     public void validate(final PreparedUpdate update, final UpdateContext updateContext) {
         final Domain domain = Domain.parse(update.getUpdatedObject().getKey());
         if (domain.getType() != INADDR) {
```
maggarwal13, line 47: There is already a NServerValidator . Is there a reason we are using intersection validator class for Nserver validation?

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/domain/DomainIntersectionValidator.java
```diff
@@ -41,10 +45,11 @@ public DomainIntersectionValidator(final Ipv4DomainTree ipv4DomainTree,
...
+            validateNserverCorrectPrefixes(update.getUpdate(), false);
             return;
         }
-
         validateIntersections(update, updateContext, (Ipv4Resource)domain.getReverseIp());
+        validateNserverCorrectPrefixes(update.getUpdate(), true);
     }
```
maggarwal13, line 53: I think prefix length check is for both ipv6 and ipv4. This handles only ipv4?

whois-api/src/main/java/net/ripe/db/whois/api/rest/checks/ServiceInputChecker.java
MiguelAHM: use the Domain object to get the prefix length

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/domain/DomainIntersectionValidator.java
```diff
@@ -74,4 +79,20 @@ public ImmutableList<Action> getActions() {
...
+
+    private void validateNserverCorrectPrefixes(Update update, boolean isIpv4){
+        if (hasRipeNserver(update.getSubmittedObject()) && hasIncorrectPrefixes(update.getSubmittedObject(), isIpv4)){
+            throw new BadRequestException();
+        }
+    }
+
+    private boolean hasIncorrectPrefixes(RpslObject rpslObject, boolean isIpv4) {
+        RpslAttribute rpslAttribute = rpslObject.findAttribute(AttributeType.DESCR);
+        return !rpslAttribute.getValue().contains("/32") && !isIpv4 || !rpslAttribute.getValue().contains("/16") && isIpv4;
```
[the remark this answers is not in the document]
MiguelAHM: You are right, thank you for the notification, I am using getLength Domain class method now

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/domain/DomainIntersectionValidator.java
MiguelAHM: I was thinking that we have a validator by each service but I saw yesterday that we can have a lot of them and NServerValidator is one of them in this case. So I moved there

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/domain/DomainIntersectionValidator.java
MiguelAHM: not apply because now I am using the type of the Domain to know if it is ipv4 and ipv6

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/domain/NServerValidator.java
```diff
@@ -66,4 +72,19 @@ public ImmutableList<Action> getActions() {
     public ImmutableList<ObjectType> getTypes() {
         return TYPES;
     }
+
+
+    private void validateNserverCorrectPrefixes(RpslObject rpslObject, int prefixLength, boolean isIpv4){
+        if (hasRipeNserver(rpslObject) && hasIncorrectPrefixes(prefixLength, isIpv4)){
+            throw new BadRequestException();
+        }
```
maggarwal13, line 80: Can you add a message in updateContext and create a new error message in UpdateMessages. Just check the below lines. For example we do it for invalid hostname in the same class "UpdateMessages.hostNameMustEndWith(domain.getValue())"

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/domain/NServerValidator.java
```diff
@@ -66,4 +72,19 @@ public ImmutableList<Action> getActions() {
...
+    }
+
+    private boolean hasIncorrectPrefixes(int prefixLength, boolean isIpv4) {
```
maggarwal13, line 83: use final where we can

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/domain/NServerValidator.java
```diff
@@ -44,6 +48,8 @@ public void validate(final PreparedUpdate update, final UpdateContext updateCont
                 case INADDR:
                 case IP6:
                 {
+                    validateNserverCorrectPrefixes(update.getUpdatedObject(), domain.getReverseIp().getPrefixLength(),
+                            domain.getType().equals(INADDR));
                     final boolean endsWithDomain = domain.endsWithDomain(nServer.getHostname());
```
maggarwal13, line 53: can we have this check after hostname validation is performed as ipInterval can be null and it should fail before that

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/domain/NServerValidator.java
```diff
@@ -66,4 +72,19 @@ public ImmutableList<Action> getActions() {
...
+        }
+    }
+
+    private boolean hasIncorrectPrefixes(int prefixLength, boolean isIpv4) {
+        return prefixLength!=32 && !isIpv4 || prefixLength!=16 && isIpv4;
+    }
+
+    private boolean hasRipeNserver(RpslObject rpslObject) {
+        return rpslObject.findAttributes(AttributeType.NSERVER).stream().anyMatch(nserver -> nserver.getValue().equals("ns.ripe.net"));
+    }
```
maggarwal13, line 89: This needs to be case insensitive? Or CIString is handling this case already?

whois-api/src/test/java/net/ripe/db/whois/api/rest/DomainObjectServiceTestIntegration.java
```diff
@@ -230,7 +231,119 @@ public void create_domain_object_unparsable_json() {
                     "Unrecognized token 'bad': was expecting", "1", "32");
         }
     }
+    @Test
+    public void create_domain_object_fail_nserver_with_correct_prefix_ipv4() {
+
```
maggarwal13, line 236: can you also add a test case where nserver is not ns.ripe.net and it is allowed for /16 ?

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/domain/NServerValidator.java
MiguelAHM: in Domain.endsWithDomain if isDasNotation is false, we still could have the problem that the reverseIp is null. 
I will add it before that and add an if

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/domain/NServerValidator.java
MiguelAHM: need case sensitive

whois-update/src/main/java/net/ripe/db/whois/update/handler/UpdateRequestHandler.java
```diff
@@ -178,11 +178,7 @@ private Collection<Update> attemptUpdatesOneByOne(final UpdateRequest updateRequ
     }
 
     private void processUpdateQueueBatchUpdate(final UpdateRequest updateRequest, final UpdateContext updateContext) {
-        try {
```
eshryane, line 181: Be very careful about changing this, it will have side-effects. I suggest logging the exception in the catch block for now but still swallow the exception, so we can find out when it happens.

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/domain/NServerValidator.java
```diff
@@ -66,4 +71,18 @@ public ImmutableList<Action> getActions() {
     public ImmutableList<ObjectType> getTypes() {
         return TYPES;
     }
+
+
+    private void validateNserverCorrectPrefixes(final Domain domain, final PreparedUpdate update,
```
eshryane, line 76: I suggest renaming to ```validateRipeNsServerPrefixLength``` to be more specific (i.e. this validation is only done for ns.ripe.net)

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/domain/NServerValidator.java
```diff
@@ -66,4 +71,18 @@ public ImmutableList<Action> getActions() {
...
+                                                final boolean isIpv4, final RpslAttribute nServerAttribute, final UpdateContext updateContext){
+        if ("ns.ripe.net".equalsIgnoreCase(nServerAttribute.getValue()) && hasIncorrectPrefixes(domain.getReverseIp().getPrefixLength(),
+                isIpv4)){
+            updateContext.addMessage(update, nServerAttribute, UpdateMessages.hostNameMustEndWith(domain.getValue()));
+        }
+    }
+
+    private boolean hasIncorrectPrefixes(final int prefixLength, final boolean isIpv4) {
```
eshryane, line 84: I suggest renaming to ```hasIncorrectPrefixesForRipeNsServer()``` or similar

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/domain/NServerValidator.java
```diff
@@ -66,4 +71,18 @@ public ImmutableList<Action> getActions() {
     public ImmutableList<ObjectType> getTypes() {
         return TYPES;
     }
+
+
+    private void validateNserverCorrectPrefixes(final Domain domain, final PreparedUpdate update,
+                                                final boolean isIpv4, final RpslAttribute nServerAttribute, final UpdateContext updateContext){
+        if ("ns.ripe.net".equalsIgnoreCase(nServerAttribute.getValue()) && hasIncorrectPrefixes(domain.getReverseIp().getPrefixLength(),
+                isIpv4)){
+            updateContext.addMessage(update, nServerAttribute, UpdateMessages.hostNameMustEndWith(domain.getValue()));
```
eshryane, line 80: We need a **new** error message for this case, e.g. ```Prefix length must be /16 for IPv4 or /32 for IPv6 if ns.ripe.net is used as a nameserver```

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/domain/NServerValidator.java
eshryane: And prefix (singular) not prefixes

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/domain/NServerValidator.java
```diff
@@ -45,7 +47,10 @@ public void validate(final PreparedUpdate update, final UpdateContext updateCont
                 case IP6:
                 {
                     final boolean endsWithDomain = domain.endsWithDomain(nServer.getHostname());
-
+                    if (domain.getReverseIp() != null) {
+                        validateRipeNsServerPrefixLength(domain, update, domain.getType().equals(INADDR),
+                                nServerAttribute, updateContext);
```
maggarwal13, line 52: Suggestion: If you are passing the domain object anyways  you do not need to pass a isIpv4 boolean? 
You can have this check in the method itself. It makes it more readable

whois-api/src/test/java/net/ripe/db/whois/api/rest/DomainObjectServiceTestIntegration.java
```diff
@@ -230,7 +230,152 @@ public void create_domain_object_unparsable_json() {
...
+                    .post(Entity.entity(mapRpslObjects(domain), MediaType.APPLICATION_JSON_TYPE), WhoisResources.class);
+
+        RestTest.assertErrorCount(response, 0);
+        assertThat(response.getWhoisObjects(), hasSize(1));
+    }
+
+    @Test
+    public void create_domain_object_fail_nserver_with_another_nserver() {
+
+        databaseHelper.addObject("" +
```
maggarwal13, line 265: method name says fail but it  actually tests successful case?

maggarwal13 (approved):

PR #1069, opened 2022-08-23
feat: parallelize using threads

MiguelAHM:


.gitlab-ci.yml
```diff
@@ -95,22 +96,150 @@ whois-integration-test:
...
+  <<: *it_env
+  services:
+    - mariadb:10.2
+  tags:
+    - whois
+  stage: integration-test
+  rules:
+    - if: '$CI_COMMIT_REF_NAME == "master"'
+      when: always
+    - when: manual
```
maggarwal13, line 119: can we create a generic step so that we do not have to repeat the rules and all IT tests extends it?

.gitlab-ci.yml
```diff
@@ -95,22 +96,150 @@ whois-integration-test:
     - mariadb:10.2
   tags:
     - whois
+  stage: integration-test
+  rules:
+    - if: '$CI_COMMIT_REF_NAME == "master"'
+      when: always
+    - when: manual
```
maggarwal13, line 103: Do I need to click on run 7 times now, since we separated them? It will be quite irritating and we might forget or miss one of IT tests

.gitlab-ci.yml
```diff
@@ -95,22 +96,150 @@ whois-integration-test:
...
+      allow_failure: true
+  script:
+    - mvn -DskipTests install
+    - mvn $MAVEN_OPTS -Ddb.host=mariadb verify -Pintegration -pl whois-api -T 1C
+
+whois-client-integration-test:
+  <<: *it_env
```
maggarwal13, line 110: Minor: Names are quite large it does not fit in small modal box of gitlab  . Can we rename to whois-client-test or something similar as the stage is itself Integration stage

.gitlab-ci.yml
```diff
@@ -95,22 +96,150 @@ whois-integration-test:
...
+  stage: integration-test
+  rules:
+    - if: '$CI_COMMIT_REF_NAME == "master"'
+      when: always
+    - when: manual
+      allow_failure: true
+  script:
+    - mvn -DskipTests install
+    - mvn $MAVEN_OPTS -Ddb.host=mariadb verify -PintegrationAll -pl whois-db -T 1
+
```
maggarwal13, line 158: Do we need to specify thread count as all other tests runs separately in different gitlab runner

.gitlab-ci.yml
maggarwal13: So for example if we change the mariadb version now we need to change at 7 diffeent places

.gitlab-ci.yml
```diff
@@ -95,22 +96,150 @@ whois-integration-test:
...
+    - whois
+  stage: integration-test
+  rules:
+    - if: '$CI_COMMIT_REF_NAME == "master"'
+      when: always
+    - when: manual
+      allow_failure: true
+  script:
+    - mvn -DskipTests install
+    - mvn $MAVEN_OPTS -Ddb.host=mariadb verify -PintegrationAll -pl whois-client -T 1
```
maggarwal13, line 123: Minor: Although it takes very less time to mvn install but is it possible to share this artifact path using a generic step

.gitlab-ci.yml
MiguelAHM: yes, you are right let me check of to do that

.gitlab-ci.yml
MiguelAHM: some of them are not able to work in parallel, for those it is necessary to specify this. Just query and common I think

.gitlab-ci.yml
```diff
@@ -71,54 +71,115 @@ release:
         - "*/target/surefire-reports/TEST-*.xml"
         - "*/target/failsafe-reports/TEST-*.xml"
 
-whois-groovy-test:
+
+.itest_template: &itest_configuration
   <<: *it_env
```
maggarwal13, line 76: do we really need a new template? Can it be merged with existing one it_template?

PR #1070, opened 2022-08-24
Nro rdap profile 0

maggarwal13:
Minimal implementation of spec: https://bitbucket.org/nroecg/nro-rdap-profile/raw/v1/nro-rdap-profile.txt


eshryane:
Don't merge until we've confirmed what we must implement to comply with the NRO RDAP profile spec.

See here: https://bitbucket.org/nroecg/nro-rdap-profile/raw/v1/nro-rdap-profile.txt

PR #1073, opened 2022-08-26
Mask override param in api calls

petchells:
Avoid printing sensitive information to the Jetty logs. Reuse PasswordFilter which is already used in audit/qry/upd logs.

### DIFF
--- a/whois-api/src/main/java/net/ripe/db/whois/api/httpserver/FilteredPasswordSlf4jRequestLogWriter.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/httpserver/FilteredPasswordSlf4jRequestLogWriter.java
@@ -1,5 +1,6 @@
 package net.ripe.db.whois.api.httpserver;
 
+import net.ripe.db.whois.common.conversion.PasswordFilter;
 import org.eclipse.jetty.server.Slf4jRequestLogWriter;
 
 import javax.annotation.Nonnull;
@@ -13,8 +14,7 @@ public class FilteredPasswordSlf4jRequestLogWriter extends Slf4jRequestLogWriter
 
     @Override
     public void write(@Nonnull final String requestEntry) throws IOException {
-        //super.write(PasswordFilter.filterPasswordsInUrl(requestEntry));
-        super.write(PASSWORD_PATTERN.matcher(requestEntry).replaceAll("FILTERED"));
+        super.write(PasswordFilter.filterPasswordsInUrl(requestEntry));
     }
 
 }
--- a/whois-api/src/main/java/net/ripe/db/whois/api/httpserver/FilteredPasswordSlf4jRequestLogWriter.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/httpserver/FilteredPasswordSlf4jRequestLogWriter.java
@@ -1,17 +1,20 @@
 package net.ripe.db.whois.api.httpserver;
 
-import net.ripe.db.whois.common.conversion.PasswordFilter;
 import org.eclipse.jetty.server.Slf4jRequestLogWriter;
 
 import javax.annotation.Nonnull;
 import java.io.IOException;
 import java.util.regex.Pattern;
 
+
 public class FilteredPasswordSlf4jRequestLogWriter extends Slf4jRequestLogWriter {
+
+    private static final Pattern PASSWORD_PATTERN = Pattern.compile("(?<=(?i)(password=))([^&]*)");
 
     @Override
     public void write(@Nonnull final String requestEntry) throws IOException {
-        super.write(PasswordFilter.filterPasswordsInUrl(requestEntry));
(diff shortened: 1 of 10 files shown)


whois-api/src/test/java/net/ripe/db/whois/api/log/JettyRequestLogTestIntegration.java
```diff
@@ -101,19 +103,43 @@ public void password_filtered() throws Exception {
...
 
 
-        String actual = fileToString(getRequestLogFilename());
+        String actual = fileToString(getRequestLog());
         assertThat(actual, containsString("GET /whois/test/person/TP1-TEST?password=FILTERED"));
         assertThat(actual, not(containsString("some-api_key-123")));
     }
 
+    @Test
+    public void password_and_override_filtered() throws Exception {
```
eshryane, line 112: Add tests for mixed-case (and/or uppercase) password and override. Query parameters are case-sensitive, but if a user mixed the case by mistake we don't want sensitive data in the log.

whois-api/src/test/java/net/ripe/db/whois/api/log/JettyRequestLogTestIntegration.java
```diff
@@ -209,7 +235,17 @@ private String getRequestLogFilename() {
         return String.format("%s/%s", requestLogDirectory, "request.log");
     }
 
-    private String fileToString(final String filename) throws IOException {
-        return new String(Files.readAllBytes(new File(filename).toPath()));
+    private File getRequestLog() {
+        return new File(getRequestLogFilename());
+    }
+
+    private void clearRequestLog() throws IOException {
```
eshryane, line 242: You can also use ```new FileWriter(getRequestLog()).close()``` to truncate the file.

whois-api/src/test/java/net/ripe/db/whois/api/log/JettyRequestLogTestIntegration.java
petchells: done in 2bc549bab

whois-api/src/test/java/net/ripe/db/whois/api/log/JettyRequestLogTestIntegration.java
petchells: done in 964696d65

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/FilteredPasswordSlf4jRequestLogWriter.java
```diff
@@ -1,17 +1,17 @@
...
 import org.eclipse.jetty.server.Slf4jRequestLogWriter;
 
 import javax.annotation.Nonnull;
 import java.io.IOException;
-import java.util.regex.Pattern;
 
-public class FilteredPasswordSlf4jRequestLogWriter extends Slf4jRequestLogWriter {
 
-    private static final Pattern PASSWORD_PATTERN = Pattern.compile("(?<=(?i)(password=))([^&]*)");
+public class FilteredPasswordSlf4jRequestLogWriter extends Slf4jRequestLogWriter {
```
eshryane, line 10: Rename back to `FilteredSlf4RequestLogWriter` as it's not just filtering passwords

whois-api/src/test/java/net/ripe/db/whois/api/log/JettyRequestLogTestIntegration.java
```diff
@@ -152,11 +178,23 @@ public void password_filtered_case_insensitive() throws Exception {
...
 
+    @Test
+    public void override_filtered_case_insensitive() throws Exception {
+        RestTest.target(getPort(), "whois/test/person/TP1-TEST?oVeRrIdE=overrideUser,overPASS1")
+                .request()
+                .get(WhoisResources.class);
+
+
+        String actual = fileToString(getRequestLog());
+        assertThat(actual.toLowerCase(), containsString("override=overrideUser,FILTERED".toLowerCase()));
```
eshryane, line 194: Should we test that the query parameter name has not changed (as in, match on `oVeRrIdE`). This allows us to detect if a user is not using all lowercase in a query param name.

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/FilteredPasswordSlf4jRequestLogWriter.java
petchells: done in 44045027c

whois-api/src/test/java/net/ripe/db/whois/api/log/JettyRequestLogTestIntegration.java
petchells: yes, i think we should. i copied this test from the "PassWord" test but now you mention it, i don't think it's a good idea to 'lowerCase' everything in the test. updated in 44045027c

eshryane (approved):

PR #1076, opened 2022-09-09
Cache Auth Service (SSO) Responses

eshryane:
We often make multiple calls for the same SSO token in a short space of time, cache the responses.


pom.xml
```diff
@@ -50,6 +50,7 @@
 
         <diffutils.version>1.3.0</diffutils.version>
 
+        <ehcache.version>2.10.4</ehcache.version>
         <ezvcard.version>0.11.3</ezvcard.version>
```
maggarwal13, line 54: In whois we already have a caching mechanism from spring I think. I sit possible to use that instead of adding a new dependency?

MiguelAHM:
If we add a new technology maybe we should create some integration.

pom.xml
eshryane: Do you mean a different library that implements `org.springframework.cache.annotation.Cacheable` ?

eshryane:
> If we add a new technology maybe we should create some integration tests.

Caching is transparent so testing will be difficult. Also, for integration tests the Crowd Server is mocked.

MiguelAHM:
> > If we add a new technology maybe we should create some integration tests.
> 
> Caching is transparent so testing will be difficult. Also, for integration tests the Crowd Server is mocked.

I used to test the caches with integration test, in my previous job but using a company framework that is based on spring boot.
In Spring the config and the annotations seems more or least the same so probably there is a way to perform integration test, however I am not sure how to do that.
I am ok just checking in the logs if it works

> > If we add a new technology maybe we should create some integration tests.
> 
> Caching is transparent so testing will be difficult. Also, for integration tests the Crowd Server is mocked.

I used to test the caches with integration test, in my previous job but we were working with a framework based on spring boot.
In Spring the config and the annotations seems more or least the same so probably there is a way to perform integration test as well, however I am not sure how to do that.
I am ok just checking in the logs if it works.

Apart from this, I remember that the cached object must be serialized, but I am not sure if that applies here. We used to receive an error in the logs and the cache didn't work in those situations

eshryane:
> I used to test the caches with integration test, in my previous job but we were working with a framework based on spring boot. In Spring the config and the annotations seem more or least the same so there is probably a way to perform integration test as well, however, I am not sure how to do that. I am ok just checking in the logs to see if it works.

Caching is a "nice to have" so I'm ok not to add integration tests for it. That said, if there's a simple way to test that it's working then let's add that separately.

> Apart from this, I remember that the cached object must be serialized, but I am not sure if that applies here. We used to receive an error in the logs and the cache didn't work in those situations

```Objects used in Ehcache only need to implement Serializable if you need to store the cache on disk or are using cache replication```

See https://www.ehcache.org/documentation/3.0/serializers-copiers.html

I've confirmed this is the case by testing, a crowd response is cached and returned on subsequent requests.

eshryane:
@MiguelAHM I added an integration test for the caching behaviour, it was straightforward to just check the `cacheManager`

MiguelAHM:
> @MiguelAHM I added an integration test for the caching behaviour, it was straightforward to just check the `cacheManager`

ok, thank you!

pom.xml
eshryane: Please clarify @maggarwal13

MiguelAHM (approved):

pom.xml
maggarwal13: Sorry I got confused between whois and whois-internal. We already have this in whois-internal. Thanks for clarifying

PR #1085, opened 2022-09-27
Fixed poetic form integration test (only a single mnt-by is allowed).

eshryane:
See poetic-form template change in #1080


eshryane:
Integration tests have passed

PR #1086, opened 2022-09-29
Control Health Check via JMX

eshryane:


whois-api/src/main/java/net/ripe/db/whois/api/healthcheck/HealthCheckJmx.java
```diff
@@ -0,0 +1,39 @@
...
+    @Autowired
+    public HealthCheckJmx() {
+        super(LOGGER);
+    }
+
+    @Override
+    @ManagedOperation(description = "Get JMX healthcheck status")
+    public boolean check() {
+        return enabled;
+    }
```
maggarwal13, line 28: Minor: feel free to ignore. Can we rename to isEnabled? It is not checking just enabling or disabling the health check via jmx?  I do not think it needs to implement healthcheck interface

whois-api/src/main/java/net/ripe/db/whois/api/healthcheck/HealthCheckService.java
```diff
@@ -0,0 +1,39 @@
...
+        this.healthChecks = healthChecks;
+    }
+
+    @GET
+    public Response check() {
+        boolean healthy = true;
+        for (HealthCheck healthCheck : healthChecks) {
+            if (!healthCheck.check()) {
+                healthy = false;
+                break;
```
maggarwal13, line 29: can we return Response.status(Status.SERVICE_UNAVAILABLE).entity("DISABLED").build(); from here itself instead of break statement? This way we will not be needing the local variable healthy and we can just

whois-api/src/main/java/net/ripe/db/whois/api/healthcheck/HealthCheckJmx.java
eshryane: This method is used both by the HealthCheckService to check the status of the JMX healthcheck, and also as a JMX operation for a user to check the status. It's not enabling or disabling (those are separate operations)

eshryane: It's checking the current status

whois-api/src/main/java/net/ripe/db/whois/api/healthcheck/HealthCheckService.java
eshryane: Good idea I'll change it

maggarwal13 (approved):

PR #1097, opened 2022-10-14
Fix 3 flaky tests in RdapResponseJsonTest

YiweiLi4:
Test `domain_serialization_test` is flaky since the JSON generator doesn't write the values in a deterministic order. So I configured the mapper and changed the order of values in the expected String.

### DIFF
--- a/whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapResponseJsonTest.java
+++ b/whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapResponseJsonTest.java
@@ -151,7 +151,7 @@ public class RdapResponseJsonTest {
         lastChangedEvent.setEventActor("joe@example.com");
         nameserver.getEvents().add(lastChangedEvent);
 
-        assertThat(marshal(nameserver), equalTo("" +
+                assertThat(marshal(nameserver), equalTo("" +
                 "{\n" +
                 "  \"handle\" : \"handle\",\n" +
                 "  \"ldhName\" : \"ns1.xn--fo-5ja.example\",\n" +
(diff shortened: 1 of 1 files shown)


eshryane:
Thanks for your contribution @YiweiLi4 we are working on the RDAP service now and we will include this for reviw

maggarwal13 (approved):

eshryane:
@YiweiLi4 the unit tests in ```RdapResponseJsonTest.java``` are not successful for me, can you check please?

YiweiLi4:
Sorry for the late reply. It is because I modified the helper function `marshal`, which leads to the failure of the unit tests for `nameserver_serialization_test` and `ip_serialization_test`. I just rearranged the order of values in the corresponding strings. It should work now. Let me know if there are still any problems:)

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapResponseJsonTest.java
```diff
@@ -494,6 +495,7 @@ private JsonFactory createJsonFactory() {
 
         objectMapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
         objectMapper.configure(SerializationFeature.WRITE_EMPTY_JSON_ARRAYS, true);
+        objectMapper.configure(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY, true);
```
petchells, line 498: This call is deprecated -- replace with...

JsonMapper.builder().configure(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY, true);

petchells:
Hi, I just tweaked your PR to remove the deprecated calls. For me it works exactly the same as before -- if you can confirm that then I'll merge it. Tnx!
~Paul

YiweiLi4:
[petchells](https://github.com/petchells) Looks good for me. Thank you!

petchells:
merged to fix-unstable-rdap-test so it runs on RIPE's ci/cd

PR #1099, opened 2022-10-20
Initialise the nrtmv4 database

petchells:


whois-commons/src/test/java/net/ripe/db/whois/common/dao/jdbc/DatabaseHelper.java
```diff
@@ -249,9 +257,9 @@ private static DataSource createDataSource( final String databaseName) {
     }
 
     static void ensureLocalhost(final JdbcTemplate jdbcTemplate) {
-        final boolean isLocalhostOrRdonly = jdbcTemplate.execute(new ConnectionCallback<Boolean>() {
+        final boolean isLocalhostOrRdonly = Boolean.TRUE.equals(jdbcTemplate.execute(new ConnectionCallback<Boolean>() {
```
eshryane, line 260: Can execute() return `null` ? Better to fail on assigning to a boolean rather than default to `true`

whois-commons/src/main/resources/nrtm_data.sql
```diff
@@ -0,0 +1,2 @@
+insert into source(id, name)
+values (100, 'RIPE_TEST_SOURCE');
```
eshryane, line 2: use the existing source name `TEST` or the schema name `WHOIS_TEST_TEST`

whois-commons/src/test/java/net/ripe/db/whois/common/dao/jdbc/DatabaseHelper.java
```diff
@@ -541,7 +550,7 @@ private static void validateFilePerTable(final JdbcTemplate jdbcTemplate) {
             return rs.getBoolean(1);
         });
 
-        if (filePerTable) {
+        if (Boolean.TRUE.equals(filePerTable)) {
```
eshryane, line 553: Don't allow `filePerTable` to be null (warn or fail)

whois-commons/src/main/resources/nrtm_schema.sql
```diff
@@ -0,0 +1,58 @@
...
+/*!40103 SET TIME_ZONE = '+00:00' */;
+/*!40014 SET @OLD_UNIQUE_CHECKS = @@UNIQUE_CHECKS, UNIQUE_CHECKS = 0 */;
+/*!40014 SET @OLD_FOREIGN_KEY_CHECKS = @@FOREIGN_KEY_CHECKS, FOREIGN_KEY_CHECKS = 0 */;
+/*!40101 SET @OLD_SQL_MODE = @@SQL_MODE, SQL_MODE = 'NO_AUTO_VALUE_ON_ZERO' */;
+/*!40111 SET @OLD_SQL_NOTES = @@SQL_NOTES, SQL_NOTES = 0 */;
+
+DROP TABLE IF EXISTS `source`;
+create table `source`
+(
+    `id`   int(11) unsigned NOT NULL AUTO_INCREMENT,
```
eshryane, line 15: Do you need an `id` column if the `name` is unique?

whois-commons/src/main/resources/nrtm_schema.sql
```diff
@@ -0,0 +1,58 @@
...
+    `name` varchar(40)      NOT NULL DEFAULT '',
+    PRIMARY KEY (`id`),
+    UNIQUE KEY `source__name_uk` (`name`)
+) ENGINE = InnoDB
+  DEFAULT CHARSET = latin1;
+
+DROP TABLE IF EXISTS `version_information`;
+create table `version_information`
+(
+    `id`         int(11) unsigned NOT NULL AUTO_INCREMENT,
```
eshryane, line 25: Do you need an `id` column?

whois-commons/src/main/resources/nrtm_data.sql
petchells: this has since been removed -- data used in the tests is set up explicitly in the test class

whois-nrtm4/src/test/java/net/ripe/db/nrtm4/AbstractNrtm4IntegrationBase.java
```diff
@@ -0,0 +1,8 @@
+package net.ripe.db.nrtm4;
+
+import net.ripe.db.whois.common.dao.jdbc.AbstractDatabaseHelperIntegrationTest;
+
+
+public class AbstractNrtm4IntegrationBase extends AbstractDatabaseHelperIntegrationTest {
```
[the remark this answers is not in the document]
petchells: it's been removed since i wasn't using it. but it was working without the annotation.

whois-commons/src/main/resources/nrtm_schema.sql
petchells: not for the code itself but i think in general it's always a good idea to use one in the db: you never know when someone might need to make the other columns non-unique and then you'll need an ID. it's always much harder to add later, whereas it's really easy to remove. so basically the reason is 'experience says use an explicit ID' but i'm ok to remove it if it bothers someone.

whois-commons/src/test/java/net/ripe/db/whois/common/dao/jdbc/DatabaseHelper.java
petchells: i would suppose that autoboxing would come into play then it's not so clear what would happen -- i just wanted to make it explicit. but yes, your suggestion is better still. i'll do that.

petchells: added extra check

whois-commons/src/test/java/net/ripe/db/whois/common/dao/jdbc/DatabaseHelper.java
petchells: added extra check

whois-commons/src/main/resources/applicationContext-commons.xml
```diff
@@ -74,6 +74,12 @@
         <property name="password" value="${whois.db.slave.password}"/>
     </bean>
 
+    <bean id="nrtmDataSource" parent="abstractDataSource">
```
eshryane, line 77: Why should this go into commons? It's only needed for nrtmv4?

whois-commons/src/main/resources/nrtm_schema.sql
eshryane: OK

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/NrtmSource.java
```diff
@@ -0,0 +1,9 @@
+package net.ripe.db.nrtm4.persist;
```
eshryane, line 1: Can you define a new property instead (or use the existing property `whois.source`)

whois-commons/src/main/resources/nrtm_schema.sql
MiguelAHM: maybe here we can use the name. I understand the point of Paul, to use Id always to identify the row instead of another column (doesn't matter if that column is unique or not), I guess that this is just different point of view.

whois-commons/src/main/resources/applicationContext-commons.xml
petchells: yup, moved to whois-nrtm4 module

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/NrtmSource.java
petchells: i could, but then there would be only one source -- the story says we need three. is the idea to have three servers, one for each source? i assumed one server for n sources.

whois-commons/src/main/resources/nrtm_schema.sql
petchells: yes that's really it. i'm not making any assumptions about uniqueness coz it's not 100% designed yet. when we've got a working implementation it might turn out that this could be optimized. even then, thinking about it purely as an abstract data source, it's almost always a good idea to have a dedicated unique identifier rather than overloading the role of the 'name'  column to be both a business code and technically unique, since then it has mixed responsibility, and introduces a constraint you might not always want.

whois-commons/src/test/java/net/ripe/db/whois/common/dao/jdbc/DatabaseHelper.java
```diff
@@ -249,20 +253,19 @@ private static DataSource createDataSource( final String databaseName) {
...
+            final String username = metaData.getUserName();
+
+            return url.contains("localhost")
+                    || url.contains("mariadb")
+                    || url.contains("127.0.0.1")
+                    || username.startsWith("rdonly");
         });
-
+        if (isLocalhostOrRdonly == null) {
+            throw new NullPointerException("Result of query was null in 'ensureLocalhost(...)'");
```
eshryane, line 267: throw `IllegalStateException` instead of NullPointerException? No big deal as it's in test scaffolding

eshryane (approved):

whois-commons/src/test/java/net/ripe/db/whois/common/dao/jdbc/DatabaseHelper.java
petchells: yes, better.

whois-commons/src/test/java/net/ripe/db/whois/common/dao/jdbc/DatabaseHelper.java
```diff
@@ -315,6 +318,10 @@ public JdbcTemplate getInternalsTemplate() {
         return internalsTemplate;
     }
 
+    public JdbcTemplate getNrtmTemplate() {
```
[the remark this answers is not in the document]
petchells: yes, done

PR #1102, opened 2022-11-01
Create an empty snapshot file

petchells:
This PR is an implementation of the second NRTM story but is built on top of the first: 'initialise-the-nrtmv4-database' (see #1099). Both are therefore shown in the diff. It's better to review and merge the earlier one before doing this one.


whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NotificationFileGenerationService.java
```diff
@@ -0,0 +1,43 @@
...
+
+@Service
+public class NotificationFileGenerationService {
+
+    private final NrtmVersionInformationDao versionDao;
+
+    public NotificationFileGenerationService(
+            final NrtmVersionInformationDao nrtmVersionInformationDao
+    ) {
+        versionDao = nrtmVersionInformationDao;
```
MiguelAHM, line 21: this.versionDao

PR #1103, opened 2022-11-02
Db 2337 include networks autnums rdap

MiguelAHM:
https://jira.ripe.net/browse/DB-2337

### DIFF
--- a/whois-api/src/main/java/net/ripe/db/whois/api/TopLevelFilter.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/TopLevelFilter.java
@@ -20,14 +20,14 @@ public class TopLevelFilter<T extends Interval<T>>  {
 
     }
 
-    private List<RpslObject> matchRpslWithTree(List<RpslObject> rpslObjects, NestedIntervalMap<T, CIString> tree) {
+    private List<RpslObject> matchRpslWithTree(final List<RpslObject> rpslObjects, final NestedIntervalMap<T, CIString> tree) {
         return rpslObjects.stream().filter(rpslObject -> !tree.findExact((T) IpInterval.parse(rpslObject.getKey())).isEmpty()).collect(Collectors.toList());
     }
 
-    private void buildTree(List<RpslObject> rpslObjects, NestedIntervalMap<T, CIString> tree) {
+    private void buildTree(final List<RpslObject> rpslObjects, final NestedIntervalMap<T, CIString> tree) {
         for (RpslObject rpslObject : rpslObjects) {
             final T key = (T) IpInterval.parse(rpslObject.getKey());
-            List<CIString> moreSpecific = tree.findFirstMoreSpecific(key);
+            final List<CIString> moreSpecific = tree.findFirstMoreSpecific(key);
             if (!moreSpecific.isEmpty()){
                 final T moreSpecificKey = (T) IpInterval.parse(moreSpecific.get(0));
                 tree.remove(moreSpecificKey);
--- a/whois-api/src/main/java/net/ripe/db/whois/api/TopLevelFilter.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/TopLevelFilter.java
@@ -5,39 +5,39 @@ import net.ripe.db.whois.common.etree.NestedIntervalMap;
 import net.ripe.db.whois.common.ip.Interval;
 import net.ripe.db.whois.common.ip.IpInterval;
 import net.ripe.db.whois.common.rpsl.RpslObject;
+import org.apache.commons.compress.utils.Lists;
 
 import java.util.List;
 import java.util.stream.Collectors;
 
 public class TopLevelFilter<T extends Interval<T>>  {
+    private final NestedIntervalMap<T, CIString> tree;
+    private final List<RpslObject> rpslObjects = Lists.newArrayList();
+    public TopLevelFilter(){
+        this.tree = new NestedIntervalMap<>();
+    }
 
-
-    public List<RpslObject> filter(final List<RpslObject> rpslObjects) {
-        final NestedIntervalMap<T, CIString> tree = new NestedIntervalMap<>();
-
-        buildTree(rpslObjects, tree);
-        return matchRpslWithTree(rpslObjects, tree);
+    public List<RpslObject> filter() {
+        return matchRpslWithTree(this.rpslObjects);
 
     }
 
-    private List<RpslObject> matchRpslWithTree(final List<RpslObject> rpslObjects, final NestedIntervalMap<T, CIString> tree) {
-        return rpslObjects.stream().filter(rpslObject -> !tree.findExact((T) IpInterval.parse(rpslObject.getKey())).isEmpty()).collect(Collectors.toList());
-    }
-
-    private void buildTree(final List<RpslObject> rpslObjects, final NestedIntervalMap<T, CIString> tree) {
-        for (RpslObject rpslObject : rpslObjects) {
-            final T key = (T) IpInterval.parse(rpslObject.getKey());
-            final List<CIString> moreSpecific = tree.findFirstMoreSpecific(key);
-            if (!moreSpecific.isEmpty()){
-                final T moreSpecificKey = (T) IpInterval.parse(moreSpecific.get(0));
-                tree.remove(moreSpecificKey);
+    public void addElementToTree(final RpslObject rpslObject) {
+        final T key = (T) IpInterval.parse(rpslObject.getKey());
+        final List<CIString> moreSpecific = tree.findFirstMoreSpecific(key);
+        if (!moreSpecific.isEmpty()){
+            final T moreSpecificKey = (T) IpInterval.parse(moreSpecific.get(0));
+            tree.remove(moreSpecificKey);
+            tree.put(key, rpslObject.getKey());
+        } else {
+            if (tree.findExactAndAllLessSpecific(key).isEmpty()){
                 tree.put(key, rpslObject.getKey());
-            } else {
-                if (tree.findExactAndAllLessSpecific(key).isEmpty()){
-                    tree.put(key, rpslObject.getKey());
-                }
             }
         }
+        rpslObjects.add(rpslObject);
     }
 
+    private List<RpslObject> matchRpslWithTree(final List<RpslObject> rpslObjects) {
+        return rpslObjects.stream().filter(rpslObject -> !tree.findExact((T) IpInterval.parse(rpslObject.getKey())).isEmpty()).collect(Collectors.toList());
+    }
 }
--- a/whois-api/src/main/java/net/ripe/db/whois/api/TopLevelFilter.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/TopLevelFilter.java
@@ -1,43 +1,42 @@
 package net.ripe.db.whois.api;
 
-import net.ripe.db.whois.common.domain.CIString;
 import net.ripe.db.whois.common.etree.NestedIntervalMap;
 import net.ripe.db.whois.common.ip.Interval;
 import net.ripe.db.whois.common.ip.IpInterval;
 import net.ripe.db.whois.common.rpsl.RpslObject;
-import org.apache.commons.compress.utils.Lists;
 
+import java.util.Collections;
 import java.util.List;
-import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class TopLevelFilter<T extends Interval<T>>  {
-    private final NestedIntervalMap<T, CIString> tree;
-    private final List<RpslObject> rpslObjects = Lists.newArrayList();
-    public TopLevelFilter(){
-        this.tree = new NestedIntervalMap<>();
+    private NestedIntervalMap<T, RpslObject> tree;
+    public TopLevelFilter(Stream<RpslObject> rpslObjectStream){
+        this.buildTree(rpslObjectStream);
     }
 
     public List<RpslObject> filter() {
-        return matchRpslWithTree(this.rpslObjects);
-
-    }
-
-    public void addElementToTree(final RpslObject rpslObject) {
-        final T key = (T) IpInterval.parse(rpslObject.getKey());
-        final List<CIString> moreSpecific = tree.findFirstMoreSpecific(key);
-        if (!moreSpecific.isEmpty()){
-            final T moreSpecificKey = (T) IpInterval.parse(moreSpecific.get(0));
-            tree.remove(moreSpecificKey);
-            tree.put(key, rpslObject.getKey());
-        } else {
-            if (tree.findExactAndAllLessSpecific(key).isEmpty()){
-                tree.put(key, rpslObject.getKey());
-            }
+        if(tree == null){
+            return Collections.emptyList();
         }
-        rpslObjects.add(rpslObject);
+        return tree.mapToValues();
+
+    }
+    private void buildTree(Stream<RpslObject> rpslObjects) {
+        rpslObjects.forEach(rpslObject -> {
+            this.tree = new NestedIntervalMap<>();
+            final T key = (T) IpInterval.parse(rpslObject.getKey());
+            List<RpslObject> moreSpecific = tree.findFirstMoreSpecific(key);
+            if (!moreSpecific.isEmpty()) {
+                final T moreSpecificKey = (T) IpInterval.parse(moreSpecific.get(0).getKey());
+                tree.remove(moreSpecificKey);
+                tree.put(key, rpslObject);
+            } else {
+                if (tree.findExactAndAllLessSpecific(key).isEmpty()) {
+                    tree.put(key, rpslObject);
+                }
+            }
+        });
     }
 
-    private List<RpslObject> matchRpslWithTree(final List<RpslObject> rpslObjects) {
-        return rpslObjects.stream().filter(rpslObject -> !tree.findExact((T) IpInterval.parse(rpslObject.getKey())).isEmpty()).collect(Collectors.toList());
-    }
 }
--- a/whois-api/src/main/java/net/ripe/db/whois/api/TopLevelFilter.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/TopLevelFilter.java
@@ -11,22 +11,22 @@ import java.util.stream.Stream;
 
 public class TopLevelFilter<T extends Interval<T>>  {
     private NestedIntervalMap<T, RpslObject> tree;
-    public TopLevelFilter(Stream<RpslObject> rpslObjectStream){
+    public TopLevelFilter(final Stream<RpslObject> rpslObjectStream){
         this.buildTree(rpslObjectStream);
     }
 
-    public List<RpslObject> filter() {
+    public List<RpslObject> getTopLevelValues() {
         if(tree == null){
             return Collections.emptyList();
         }
         return tree.mapToValues();
 
     }
-    private void buildTree(Stream<RpslObject> rpslObjects) {
+    private void buildTree(final Stream<RpslObject> rpslObjects) {
+        this.tree = new NestedIntervalMap<>();
         rpslObjects.forEach(rpslObject -> {
-            this.tree = new NestedIntervalMap<>();
             final T key = (T) IpInterval.parse(rpslObject.getKey());
-            List<RpslObject> moreSpecific = tree.findFirstMoreSpecific(key);
+            final List<RpslObject> moreSpecific = tree.findFirstMoreSpecific(key);
             if (!moreSpecific.isEmpty()) {
                 final T moreSpecificKey = (T) IpInterval.parse(moreSpecific.get(0).getKey());
                 tree.remove(moreSpecificKey);
--- a/whois-api/src/main/java/net/ripe/db/whois/api/TopLevelFilter.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/TopLevelFilter.java
@@ -1,50 +1,43 @@
 package net.ripe.db.whois.api;
 
-import com.google.common.collect.Maps;
 import net.ripe.db.whois.common.domain.CIString;
 import net.ripe.db.whois.common.etree.NestedIntervalMap;
 import net.ripe.db.whois.common.ip.Interval;
 import net.ripe.db.whois.common.ip.IpInterval;
-import net.ripe.db.whois.common.ip.Ipv4Resource;
-import net.ripe.db.whois.common.ip.Ipv6Resource;
-import net.ripe.db.whois.common.rpsl.ObjectType;
 import net.ripe.db.whois.common.rpsl.RpslObject;
 
-import java.util.HashMap;
 import java.util.List;
 import java.util.stream.Collectors;
 
-public class TopLevelFilter<T extends Interval<T>> {
-    private static final HashMap<ObjectType, TopLevelFilter> resourceFilterMap;
-    static {
-        resourceFilterMap = Maps.newHashMap();
-        resourceFilterMap.put(ObjectType.INETNUM, new TopLevelFilter(Ipv4Resource.parse("0/0")));
-        resourceFilterMap.put(ObjectType.INET6NUM, new TopLevelFilter(Ipv6Resource.parse("::/0")));
-    }
+public class TopLevelFilter<T extends Interval<T>>  {
 
 
-    private final T root;
-
-    public TopLevelFilter(final T root) {
-        this.root = root;
-    }
-
-    private List<RpslObject> filter(final List<RpslObject> rpslObjects) {
+    public List<RpslObject> filter(final List<RpslObject> rpslObjects) {
         final NestedIntervalMap<T, CIString> tree = new NestedIntervalMap<>();
 
+        buildTree(rpslObjects, tree);
+        return matchRpslWithTree(rpslObjects, tree);
+
+    }
+
+    private List<RpslObject> matchRpslWithTree(List<RpslObject> rpslObjects, NestedIntervalMap<T, CIString> tree) {
+        return rpslObjects.stream().filter(rpslObject -> !tree.findExact((T) IpInterval.parse(rpslObject.getKey())).isEmpty()).collect(Collectors.toList());
+    }
+
+    private void buildTree(List<RpslObject> rpslObjects, NestedIntervalMap<T, CIString> tree) {
         for (RpslObject rpslObject : rpslObjects) {
             final T key = (T) IpInterval.parse(rpslObject.getKey());
-            tree.put(key, rpslObject.getKey());
+            List<CIString> moreSpecific = tree.findFirstMoreSpecific(key);
+            if (!moreSpecific.isEmpty()){
+                final T moreSpecificKey = (T) IpInterval.parse(moreSpecific.get(0));
+                tree.remove(moreSpecificKey);
+                tree.put(key, rpslObject.getKey());
+            } else {
+                if (tree.findExactAndAllLessSpecific(key).isEmpty()){
+                    tree.put(key, rpslObject.getKey());
+                }
+            }
         }
-
-        final List<CIString> firstMoreSpecifics = tree.findFirstMoreSpecific(root);
-
-        return rpslObjects.stream()
-                .filter(rpslObject -> firstMoreSpecifics.contains(rpslObject.getKey()))
-                .collect(Collectors.toList());
     }
 
-    public static List<RpslObject> filter(final List<RpslObject> rpslObjects, final ObjectType objectType) {
-        return resourceFilterMap.get(objectType).filter(rpslObjects);
-    }
 }
--- a/whois-api/src/main/java/net/ripe/db/whois/api/TopLevelFilter.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/TopLevelFilter.java
@@ -1,0 +1,50 @@
+package net.ripe.db.whois.api;
+
+import com.google.common.collect.Maps;
+import net.ripe.db.whois.common.domain.CIString;
+import net.ripe.db.whois.common.etree.NestedIntervalMap;
+import net.ripe.db.whois.common.ip.Interval;
+import net.ripe.db.whois.common.ip.IpInterval;
+import net.ripe.db.whois.common.ip.Ipv4Resource;
+import net.ripe.db.whois.common.ip.Ipv6Resource;
+import net.ripe.db.whois.common.rpsl.ObjectType;
+import net.ripe.db.whois.common.rpsl.RpslObject;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class TopLevelFilter<T extends Interval<T>> {
+    private static final HashMap<ObjectType, TopLevelFilter> resourceFilterMap;
+    static {
+        resourceFilterMap = Maps.newHashMap();
+        resourceFilterMap.put(ObjectType.INETNUM, new TopLevelFilter(Ipv4Resource.parse("0/0")));
+        resourceFilterMap.put(ObjectType.INET6NUM, new TopLevelFilter(Ipv6Resource.parse("::/0")));
+    }
+
+
+    private final T root;
+
+    public TopLevelFilter(final T root) {
+        this.root = root;
+    }
+
+    private List<RpslObject> filter(final List<RpslObject> rpslObjects) {
+        final NestedIntervalMap<T, CIString> tree = new NestedIntervalMap<>();
+
+        for (RpslObject rpslObject : rpslObjects) {
+            final T key = (T) IpInterval.parse(rpslObject.getKey());
+            tree.put(key, rpslObject.getKey());
+        }
+
+        final List<CIString> firstMoreSpecifics = tree.findFirstMoreSpecific(root);
+
+        return rpslObjects.stream()
+                .filter(rpslObject -> firstMoreSpecifics.contains(rpslObject.getKey()))
+                .collect(Collectors.toList());
+    }
+
+    public static List<RpslObject> filter(final List<RpslObject> rpslObjects, final ObjectType objectType) {
+        return resourceFilterMap.get(objectType).filter(rpslObjects);
+    }
+}
--- a/whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
@@ -6,6 +6,7 @@ import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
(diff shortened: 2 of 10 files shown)


whois-api/src/main/java/net/ripe/db/whois/api/TopLevelFilter.java
```diff
@@ -0,0 +1,50 @@
...
+import net.ripe.db.whois.common.ip.Ipv4Resource;
+import net.ripe.db.whois.common.ip.Ipv6Resource;
+import net.ripe.db.whois.common.rpsl.ObjectType;
+import net.ripe.db.whois.common.rpsl.RpslObject;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class TopLevelFilter<T extends Interval<T>> {
```
MiguelAHM, line 17: We have this same class in whois-internal, should we move this class to whois-common and change the reference?

whois-api/src/main/java/net/ripe/db/whois/api/TopLevelFilter.java
```diff
@@ -0,0 +1,50 @@
...
+    public TopLevelFilter(final T root) {
+        this.root = root;
+    }
+
+    private List<RpslObject> filter(final List<RpslObject> rpslObjects) {
+        final NestedIntervalMap<T, CIString> tree = new NestedIntervalMap<>();
+
+        for (RpslObject rpslObject : rpslObjects) {
+            final T key = (T) IpInterval.parse(rpslObject.getKey());
+            tree.put(key, rpslObject.getKey());
```
eshryane, line 37: I suggest only putting top-level objects in the tree - for each rpslObject, check if there is already a less-specific object in the tree, and if so don't add it. Then return the entire tree.

whois-api/src/main/java/net/ripe/db/whois/api/TopLevelFilter.java
```diff
@@ -0,0 +1,50 @@
...
+import net.ripe.db.whois.common.ip.Ipv6Resource;
+import net.ripe.db.whois.common.rpsl.ObjectType;
+import net.ripe.db.whois.common.rpsl.RpslObject;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class TopLevelFilter<T extends Interval<T>> {
+    private static final HashMap<ObjectType, TopLevelFilter> resourceFilterMap;
```
eshryane, line 18: There should be no need to filter out 0/0 or ::/0 as they are not referenced by the organisation entity

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -156,6 +159,63 @@ public Object mapSearch(final String requestUrl, final List<RpslObject> objects,
         return mapCommons(searchResult, requestUrl);
     }
 
+    public Object mapMergeSearch(final String requestUrl, List<RpslObject> objects,
+                                 final Iterable<LocalDateTime> localDateTimes, final int maxResultSize){
+        RdapObject organisation = new RdapObject();
+        List<Ip> ipv4Networks = new ArrayList<>();
```
eshryane, line 165: Use `Lists.newArrayList()` instead of `new ArrayList` for consistency

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -156,6 +159,63 @@ public Object mapSearch(final String requestUrl, final List<RpslObject> objects,
         return mapCommons(searchResult, requestUrl);
     }
 
+    public Object mapMergeSearch(final String requestUrl, List<RpslObject> objects,
+                                 final Iterable<LocalDateTime> localDateTimes, final int maxResultSize){
+        RdapObject organisation = new RdapObject();
```
eshryane, line 164: Use final keyword wherever possible

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -156,6 +159,63 @@ public Object mapSearch(final String requestUrl, final List<RpslObject> objects,
...
+            } else if (object.getType() == ObjectType.ORGANISATION) {
+                organisation = getRdapObject(requestUrl, object, iterator.next(), Optional.empty());
+            }
+        }
+
+        organisation.setAutnums(autnums);
+        organisation.setNetworks(ipv4Networks, ipv6Networks);
+
+        if(outOfLimitNotice != null){
+            organisation.getNotices().add(outOfLimitNotice);
```
eshryane, line 199: `outOfLimitNotice` is added *twice*. I suggest removing the `Notice outOfLimitNotice = null` above and the not null code here.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -156,6 +159,63 @@ public Object mapSearch(final String requestUrl, final List<RpslObject> objects,
...
+            List<RpslObject> ipv6Rpsl = TopLevelFilter.filter(objects.stream().filter(rpsl -> ObjectType.INET6NUM.equals(rpsl.getType()))
+                    .collect(Collectors.toList()), ObjectType.INET6NUM);
+
+            objects = new ArrayList<>();
+            objects.addAll(autnumRpsl);
+            objects.addAll(organisationRpsl);
+            objects.addAll(ipv4Rpsl);
+            objects.addAll(ipv6Rpsl);
+
+            outOfLimitNotice = new Notice();
```
eshryane, line 173: Only define `outOfLimitNotice` (once) here and not above (as null)

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
MiguelAHM: we are setting the value of the organisation below, we can not use final

whois-api/src/main/java/net/ripe/db/whois/api/rdap/WhoisRdapService.java
```diff
@@ -76,7 +78,8 @@ public WhoisRdapService(final RdapQueryHandler rdapQueryHandler,
                             final SourceContext sourceContext,
                             @Value("${rdap.public.baseUrl:}") final String baseUrl,
                             final RdapRequestValidator rdapRequestValidator,
-                            @Value("${rdap.search.max.results:100}") final int maxResultSize) {
+                            @Value("${rdap.search.max.results:100}") final int maxResultSize,
```
eshryane, line 81: These arguments are very similar, I suggest documenting what the difference between them is (i.e. in JavaDoc)

whois-api/src/main/java/net/ripe/db/whois/api/rdap/WhoisRdapService.java
```diff
@@ -119,7 +123,7 @@ public Response lookup(@Context final HttpServletRequest request,
             }
             case ENTITY: {
                 rdapRequestValidator.validateEntity(key);
-                return lookupObject(request, whoisObjectTypes, key);
+                return key.startsWith("ORG-") ? lookupForOrganisation(request, whoisObjectTypes, key) : lookupObject(request, whoisObjectTypes, key);
```
eshryane, line 126: Org is case-insensitive?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/WhoisRdapService.java
```diff
@@ -209,6 +213,33 @@ protected Response lookupObject(final HttpServletRequest request, final Set<Obje
...
+                String.format("%s %s %s %s %s %s %s %s",
+                        QueryFlag.NO_GROUPING.getLongFlag(),
+                        QueryFlag.NO_REFERENCED.getLongFlag(),
+                        QueryFlag.SELECT_TYPES.getLongFlag(),
+                        objectTypesToString(List.of(AUT_NUM, ObjectType.INETNUM, ObjectType.INET6NUM)),
+                        QueryFlag.NO_FILTERING.getLongFlag(),
+                        QueryFlag.INVERSE.getLongFlag(),
+                        AttributeType.ORG.getName(),
+                        key));
+        List<RpslObject> organisationResult = rdapQueryHandler.handleQuery(getQueryObject(objectTypes, key), request);
```
eshryane: Use final

whois-api/src/main/java/net/ripe/db/whois/api/rdap/WhoisRdapService.java
```diff
@@ -209,6 +213,33 @@ protected Response lookupObject(final HttpServletRequest request, final Set<Obje
         return getResponse(request, result);
     }
 
+    protected Response lookupForOrganisation(final HttpServletRequest request, final Set<ObjectType> objectTypes,
+                                             final String key) {
+        Query autnumInetnumForOrganisationQuery = Query.parse(
```
eshryane: Move this variable down past the `organisationResult` check as it's not needed until afterwards

whois-api/src/main/java/net/ripe/db/whois/api/rdap/WhoisRdapService.java
```diff
@@ -209,6 +213,33 @@ protected Response lookupObject(final HttpServletRequest request, final Set<Obje
...
+                        QueryFlag.INVERSE.getLongFlag(),
+                        AttributeType.ORG.getName(),
+                        key));
+        List<RpslObject> organisationResult = rdapQueryHandler.handleQuery(getQueryObject(objectTypes, key), request);
+
+        if (organisationResult.size() > 1){
+            throw new IllegalStateException("Unexpected result size: " + organisationResult.size());
+        }
+
+        List<RpslObject> organisationAsResult = rdapQueryHandler.handleQuery(autnumInetnumForOrganisationQuery,
```
eshryane: Rename variable it's confusingly similar to `organisationResult`, how about `resourceResult` ?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/WhoisRdapService.java
```diff
@@ -209,6 +213,33 @@ protected Response lookupObject(final HttpServletRequest request, final Set<Obje
...
+                request);
+
+        List<RpslObject> result = new ArrayList<>();
```
eshryane, line 239: You can avoid creating a new List by using `com.google.common.collect.Iterables.concat(listA, listB)`

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/RdapObject.java
```diff
@@ -27,13 +30,13 @@
 @XmlSeeAlso({
     Nameserver.class,
     Entity.class,
-    Ip.class,
```
eshryane, line 30: Why remove Ip and Autnum ?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/RdapObject.java
```diff
@@ -155,4 +172,16 @@ public String getObjectClassName() {
     public void setObjectClassName(final String value) {
         this.objectClassName = value;
     }
+
+    public void setAutnums(final List<Autnum> autnums){
+        this.autnums = autnums;
+    }
+
+    public void setNetworks(List<Ip> ipv4Networks, List<Ip> ipv6Networks) {
+        List<Ip> networks = new ArrayList<>();
```
eshryane, line 181: Use Lists.newArrayList() for consistency

whois-api/src/test/java/net/ripe/db/whois/api/rdap/WhoisRdapServiceTestIntegration.java
```diff
@@ -1341,8 +1342,166 @@ public void lookup_org_entity_handle() {
                 .get(Entity.class);
 
         assertThat(response.getHandle(), equalTo("ORG-TEST1-TEST"));
+        assertTrue(response.getAutnums().isEmpty());
+        assertTrue(response.getNetworks().isEmpty());
     }
 
+    @Test
```
eshryane, line 1349: Add a test for references to a *hierarchy* of IPv4 LEGACY resources, only the top-level resource should be returned.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -156,6 +159,69 @@ public Object mapSearch(final String requestUrl, final List<RpslObject> objects,
...
+            outOfLimitNotice.setTitle(String.format("limited search results to %s maximum" , maxResultSize));
+        }
+
+        final List<Ip> ipv4Networks = new ArrayList<>();
+        final List<Ip> ipv6Networks = new ArrayList<>();
+        RdapObject organisation = new RdapObject();
+        for (final RpslObject object : objects) {
+            if (object.getType() == AUT_NUM ) {
```
eshryane, line 181: Use `switch(object.getType())` as it's more readable. Also throw IllegalStateException in the default case.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
eshryane: OK. What happens if you don't find an organisation? Is an empty RdapObject() returned? Is that case handled elsewhere?

whois-api/src/main/java/net/ripe/db/whois/api/TopLevelFilter.java
eshryane: Good idea

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -156,6 +159,69 @@ public Object mapSearch(final String requestUrl, final List<RpslObject> objects,
         return mapCommons(searchResult, requestUrl);
     }
 
+    public Object mapOrganisationEntity(final String requestUrl, List<RpslObject> objects,
```
eshryane, line 162: `List<RpslObject>` can be 100,000's in size. We should refactor `RdapQueryHandler.handleQuery()` to accept a `ResponseHandler` or return an `Iterable` of RpslObjects instead.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/WhoisRdapService.java
```diff
@@ -205,10 +227,44 @@ private Boolean isRedirect(ObjectType objectType, final String key) {
     }
 
     protected Response lookupObject(final HttpServletRequest request, final Set<ObjectType> objectTypes, final String key) {
-        List<RpslObject> result =  rdapQueryHandler.handleQuery(getQueryObject(objectTypes, key), request);
+        Iterable<RpslObject> result =  rdapQueryHandler.handleQuery(getQueryObject(objectTypes, key), request);
```
petchells, line 230: Why not use the stream method here, too?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/WhoisRdapService.java
```diff
@@ -205,10 +227,44 @@ private Boolean isRedirect(ObjectType objectType, final String key) {
...
         return getResponse(request, result);
     }
 
+    protected Response lookupForOrganisation(final HttpServletRequest request, final Set<ObjectType> objectTypes,
+                                             final String key) {
+        final List<RpslObject> organisationStream = rdapQueryHandler.handleQuery(getQueryObject(objectTypes,
```
petchells, line 236: organisationStream isn't actually a stream, it's a list. better to use the method that returns a stream.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/WhoisRdapService.java
```diff
@@ -205,10 +227,44 @@ private Boolean isRedirect(ObjectType objectType, final String key) {
...
+                        QueryFlag.NO_GROUPING.getLongFlag(),
+                        QueryFlag.NO_REFERENCED.getLongFlag(),
+                        QueryFlag.SELECT_TYPES.getLongFlag(),
+                        objectTypesToString(List.of(AUT_NUM, ObjectType.INETNUM, ObjectType.INET6NUM)),
+                        QueryFlag.NO_FILTERING.getLongFlag(),
+                        QueryFlag.INVERSE.getLongFlag(),
+                        AttributeType.ORG.getName(),
+                        key));
+
+        final Stream.Builder<RpslObject> resourcesStream = rdapQueryHandler.handleQueryIter(autnumInetnumForOrganisationQuery,
```
petchells, line 258: handleQueryIter should be renamed to streamQueryResults or sth like that. btw it's probably better to return a Stream than a Stream.Builder

whois-api/src/main/java/net/ripe/db/whois/api/rdap/WhoisRdapService.java
```diff
@@ -220,17 +276,27 @@ private Query getQueryObject(final Set<ObjectType> objectTypes, final String key
...
+                .header(CONTENT_TYPE, CONTENT_TYPE_RDAP_JSON)
+                .build();
+    }
+    private Response getResponse(HttpServletRequest request, Iterable<RpslObject> result) {
+        Iterator<RpslObject> rpslIterator = result.iterator();
 
-        if (result.size() > 1) {
-            throw new IllegalStateException("Unexpected result size: " + result.size());
+        if (!rpslIterator.hasNext()) {
+            throw new NotFoundException("not found");
```
petchells, line 292: it's already a NotFoundException, so saying "not found" doesn't add any information. better say "result is empty"

whois-api/src/main/java/net/ripe/db/whois/api/rdap/WhoisRdapService.java
```diff
@@ -220,17 +276,27 @@ private Query getQueryObject(final Set<ObjectType> objectTypes, final String key
...
         }
 
+        final RpslObject resultObject = rpslIterator.next();
 
-        final RpslObject resultObject = result.get(0);
+        if (rpslIterator.hasNext()) {
```
petchells, line 297: so there's only ever a single RpslObject returned? if so then better name this method `getSingleObjectReponse`

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/RdapObject.java
```diff
@@ -155,4 +173,16 @@ public String getObjectClassName() {
     public void setObjectClassName(final String value) {
         this.objectClassName = value;
     }
+
+    public void setAutnums(final List<Autnum> autnums){
+        this.autnums = autnums;
+    }
+
+    public void setNetworks(List<Ip> ipv4Networks, List<Ip> ipv6Networks) {
```
petchells, line 181: missing 'final'

petchells, line 181: would it make sense to overload this into two methods -- one for v4 and one for v6? otherwise a developer might be tempted to use a null instead of an empty list, which would go badly.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/RdapRequestType.java
```diff
@@ -38,7 +38,7 @@ public Set<ObjectType> getWhoisObjectTypes(String key) {
 
     ENTITY {
         public Set<ObjectType> getWhoisObjectTypes(String key) {
```
petchells, line 40: missing 'final'

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapQueryHandler.java
```diff
@@ -33,6 +34,36 @@ public RdapQueryHandler(final QueryHandler queryHandler) {
         this.queryHandler = queryHandler;
     }
 
+    public Stream.Builder<RpslObject> handleQueryIter(final Query query, final HttpServletRequest request) {
+        Stream.Builder<RpslObject> out = Stream.builder();
```
petchells, line 38: missing 'final'

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/RdapObject.java
```diff
@@ -155,4 +173,22 @@ public String getObjectClassName() {
...
         this.objectClassName = value;
     }
+
+    public void setAutnums(final List<Autnum> autnums){
+        this.autnums = autnums;
+    }
+
+    public void setIpv4(final List<Ip> ipv4Networks) {
+        if (networks == null) {
+            networks = Lists.newArrayList();
```
maggarwal13, line 183: Why there is a different setter for ipv4 and ipv6? IP class has a ip version which should be enough to distinguish between ipv4 and ipv6. We should use the one setter method as its ultimately only one IP class

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -156,6 +161,80 @@ public Object mapSearch(final String requestUrl, final List<RpslObject> objects,
...
+
+            switch (object.getType()) {
+                case AUT_NUM:
+                    autnums.add((Autnum) getRdapObject(requestUrl, object, objectDao.getLastUpdated(object.getObjectId()),
+                            Optional.empty()));
+                    break;
+                case INETNUM:
+                    ipv4Networks.add((Ip) getRdapObject(requestUrl, object, objectDao.getLastUpdated(object.getObjectId()),
+                            Optional.empty()));
+                    break;
```
maggarwal13, line 182: Inetnum and inet6num is doing the same thing, We should not use two different lists for same IP class

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -156,6 +161,80 @@ public Object mapSearch(final String requestUrl, final List<RpslObject> objects,
...
+                    ipv6Rpsl.add(rpsl);
+                    break;
+                default:
+                    throw new IllegalStateException("Incorrect object type " + rpsl.getType());
+            }
+        });
+
+        return Stream.concat(Stream.concat(Stream.concat(autnumRpsl.stream(), organisationRpsl.stream()),
+                topLevelFilter.filter(ipv4Rpsl).stream()), topLevelFilter.filter(ipv6Rpsl).stream()).iterator();
+    }
```
maggarwal13, line 236: I think Toplevel filter uses IpInterval class you do not have to create a separate list and filter them separately for ipv4 and ipv6.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/RdapObject.java
```diff
@@ -91,6 +95,20 @@ public String getLang() {
         return lang;
     }
 
+    public List<Ip> getNetworks() {
+        if (networks == null) {
+            networks = Lists.newArrayList();
+        }
+        return networks;
```
maggarwal13, line 102: We use the same class to map other object types , will this have an empty array in json structure for them as well?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/RdapObject.java
```diff
@@ -91,6 +95,20 @@ public String getLang() {
...
+    }
+
+    public List<Autnum> getAutnums() {
+        if (autnums == null) {
+            autnums = Lists.newArrayList();
+        }
```
maggarwal13, line 108: We use the same class to map other object types , will this have an empty array in json structure for them as well?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -156,6 +161,75 @@ public Object mapSearch(final String requestUrl, final List<RpslObject> objects,
...
+
+        if (networks.size() + autnums.size()  > maxResultSize) {
+            final Notice outOfLimitNotice = new Notice();
+            outOfLimitNotice.setTitle(String.format("limited search results to %s maximum" , maxResultSize));
+            organisation.getNotices().add(outOfLimitNotice);
+        }
+        return mapCommons(organisation, requestUrl);
+    }
+
+    private Iterator<RpslObject> getTopLevelObjects(final Iterator<RpslObject> objects) {
```
petchells, line 203: is this method for grouping the objects? if so, then specify 'grouping' in the query, then there's no need for the lists or the switch -- just filter the inet(6)nums.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -156,6 +161,75 @@ public Object mapSearch(final String requestUrl, final List<RpslObject> objects,
...
     }
 
+    public Object mapOrganisationEntity(final String requestUrl, Iterator<RpslObject> objects,
+                                        final RpslObjectDao objectDao, final int maxResultSize){
+
+        final List<Autnum> autnums = Lists.newArrayList();
+        final List<Ip> networks = Lists.newArrayList();
+        RdapObject organisation = new RdapObject();
+        for (Iterator<RpslObject> it = getTopLevelObjects(objects); it.hasNext(); ) {
+            final RpslObject object = it.next();
```
maggarwal13, line 171: why we are passing org objects as well to toplevelObjects? It should just filter the top level resources . Can we just pass on the inetnum and inet6num to this function and later on add it in rdap networks setter?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/WhoisRdapService.java
```diff
@@ -205,10 +227,44 @@ private Boolean isRedirect(ObjectType objectType, final String key) {
...
+                        objectTypesToString(List.of(AUT_NUM, ObjectType.INETNUM, ObjectType.INET6NUM)),
+                        QueryFlag.NO_FILTERING.getLongFlag(),
+                        QueryFlag.INVERSE.getLongFlag(),
+                        AttributeType.ORG.getName(),
+                        key));
+
+        final Stream<RpslObject> resourcesResult = rdapQueryHandler.handleQueryStream(autnumInetnumForOrganisationQuery,
+                request);
+        return getOrganisationResponse(request, joinResults(organisationResult, resourcesResult));
+    }
```
maggarwal13, line 261: What is the use of joining the results? You can get rid of switch statements in getOrganisationResponse if you pass it separately.  We are searchimng for organisation objects and later on resources underneath it, we should not join them in one list and later on seperate them

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -156,6 +161,75 @@ public Object mapSearch(final String requestUrl, final List<RpslObject> objects,
...
+        final List<RpslObject> ipv6Rpsl = Lists.newArrayList();
+
+        objects.forEachRemaining(rpsl -> {
+            switch (rpsl.getType()){
+                case AUT_NUM:
+                    autnumRpsl.add(rpsl);
+                    break;
+                case ORGANISATION:
+                    organisationRpsl.add(rpsl);
+                    break;
```
maggarwal13, line 217: No need for org and autnum objects here ? Just inetnum and inet6num

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -156,6 +161,75 @@ public Object mapSearch(final String requestUrl, final List<RpslObject> objects,
...
+                case INET6NUM:
+                    ipv6Rpsl.add(rpsl);
+                    break;
+                default:
+                    throw new IllegalStateException("Incorrect object type " + rpsl.getType());
+            }
+        });
+
+        return Stream.concat(Stream.concat(Stream.concat(autnumRpsl.stream(), organisationRpsl.stream()),
+                topLevelFilter.filter(ipv4Rpsl).stream()), topLevelFilter.filter(ipv6Rpsl).stream()).iterator();
```
maggarwal13, line 230: Can we get rid of switch statements and use stream to filter the ipv4 and ipv6 object type and pass on to toplevel filter instead of iterating them first adding it in new list? Its more readable

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -156,6 +161,75 @@ public Object mapSearch(final String requestUrl, final List<RpslObject> objects,
...
+                    break;
+                case INET6NUM:
+                    ipv6Rpsl.add(rpsl);
+                    break;
+                default:
+                    throw new IllegalStateException("Incorrect object type " + rpsl.getType());
+            }
+        });
+
+        return Stream.concat(Stream.concat(Stream.concat(autnumRpsl.stream(), organisationRpsl.stream()),
```
petchells, line 229: is it better to add the lists with .addAll() then stream.iterate() on the full result? this method converts iterator -> list -> stream -> iterator -- seems a bit inefficient.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/WhoisRdapService.java
```diff
@@ -76,7 +95,8 @@ public WhoisRdapService(final RdapQueryHandler rdapQueryHandler,
                             final SourceContext sourceContext,
                             @Value("${rdap.public.baseUrl:}") final String baseUrl,
                             final RdapRequestValidator rdapRequestValidator,
-                            @Value("${rdap.search.max.results:100}") final int maxResultSize) {
+                            @Value("${rdap.search.max.results:100}") final int maxResultSize,
+                            @Value("${rdap.entity.max.results:100}") final int maxEntityResultSize) {
         this.rdapQueryHandler = rdapQueryHandler;
```
maggarwal13, line 100: Where we are using this limit to restrict the number of search entries we retrieve from database?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapQueryHandler.java
```diff
@@ -33,6 +34,36 @@ public RdapQueryHandler(final QueryHandler queryHandler) {
         this.queryHandler = queryHandler;
     }
 
+    public Stream<RpslObject> handleQueryStream(final Query query, final HttpServletRequest request) {
+        final Stream.Builder<RpslObject> out = Stream.builder();
+
+        final InetAddress remoteAddress = InetAddresses.forString(request.getRemoteAddr());
+        queryHandler.streamResults(query, remoteAddress, 0, new ApiResponseHandler() {
+            @Override
+            public void handle(final ResponseObject responseObject) {
```
petchells, line 43: missing try { ... } catch (final QueryException e) { ... }

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapQueryHandler.java
```diff
@@ -33,6 +34,36 @@ public RdapQueryHandler(final QueryHandler queryHandler) {
...
+                        }
+                    }
+                }
+            }
+        });
+        return out.build();
+    }
+
+
     public List<RpslObject> handleQuery(final Query query, final HttpServletRequest request) {
```
petchells, line 67: handleQuery should call handleQueryStream to reduce duplicated code.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/SearchResult.java
```diff
@@ -47,4 +55,18 @@ public void addDomainSearchResult(final Domain domain) {
         }
         domainResults.add(domain);
     }
+
+    public void addAutnumSeachResult(final Autnum autnum){
+        if (autnumResults == null){
```
maggarwal13, line 60: This is not called from anywhere?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/SearchResult.java
```diff
@@ -47,4 +55,18 @@ public void addDomainSearchResult(final Domain domain) {
...
+            autnumResults = Lists.newArrayList();
+        }
+        autnumResults.add(autnum);
+    }
+
+    public void addNetworkSeachResult(final Ip ip){
+        if (networkResults == null){
+            networkResults = Lists.newArrayList();
+        }
```
maggarwal13, line 69: This is not called from anywhere?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/RdapObject.java
MiguelAHM: Attached evidences on the task. Tested with a Domain object

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/SearchResult.java
```diff
@@ -14,7 +14,9 @@
 @XmlAccessorType(XmlAccessType.FIELD)
 @XmlType(name = "entity", propOrder = {
         "entityResults",
-        "domainResults"
+        "domainResults",
+        "autnumResults",
+        "networkResults"
 })
```
maggarwal13, line 20: why autnumResults and networkResults are added? are they being used somewhere through searchResult bean?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -156,6 +162,58 @@ public Object mapSearch(final String requestUrl, final List<RpslObject> objects,
...
+                    break;
+                default:
+                    throw new IllegalStateException("Incorrect object type " + object.getType());
+            }
+        });
+
+        List<Ip> networks = mergeTopLevelResources(requestUrl, objectDao, ipv4Rpsl, ipv6Rpsl);
+
+        if (networks.size() > maxResultSize) {
+            final Notice outOfLimitNotice = new Notice();
```
maggarwal13, line 199: what about autnums? Is limit not applied to that?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/SearchResult.java
MiguelAHM: I added it in the beginning because I thought that I would need to use seachResult and I totally forgot to delete them

maggarwal13 (approved):

eshryane (approved):

PR #1111, opened 2022-11-15
Perform filtering on nrtm4 objects

petchells:


whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NotificationFileGenerationService.java
```diff
@@ -0,0 +1,43 @@
...
+        } else {
+            version = lastVersion.get();
+            // If it's a delta then use this version. If it's a snapshot then increment it.
+            if (version.getType() == NrtmDocumentType.snapshot) {
+                version = versionDao.save(version.increment());
+            }
+        }
+        final SnapshotFile snapshotFile = new SnapshotFile(version);
+        return snapshotFile.setObjectsString("");
+    }
```
maggarwal13, line 41: Setter method should have a void return type? May be use use getter or rename to setAndGet?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/NrtmDocumentType.java
```diff
@@ -0,0 +1,5 @@
+package net.ripe.db.nrtm4.persist;
+
+public enum NrtmDocumentType {
+    snapshot,
```
maggarwal13, line 4: Enum should be in uppercase by convention? Also you do not need a comma if there is only one type

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/RpslAttributeTransformer.java
```diff
@@ -0,0 +1,26 @@
...
+
+    private final static int nrtmVersionNumber = 4;
+
+    RpslAttributeTransformer(final DummifierNrtm dummifierNrtm) {
+        this.dummifierNrtm = dummifierNrtm;
+    }
+
+    RpslObject filter(final RpslObject in) {
+        return dummifierNrtm.dummify(nrtmVersionNumber, in);
+    }
```
maggarwal13, line 24: Are we Dummyfying the data or filtering it? if its just one line I would rather call  it directly instead of creating a new class

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/NrtmVersionDao.java
```diff
@@ -35,7 +36,7 @@ public NrtmVersionDao(@Qualifier("nrtmDataSource") final DataSource dataSource)
     public Optional<VersionInformation> findLastVersion(final NrtmSource source) {
         try {
```
maggarwal13, line 37: minor: feel free to ignore. May be rename to VersionInfo ?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/NrtmVersionDao.java
```diff
@@ -56,11 +57,12 @@ public void createNew(final NrtmSource source) {
         final long version = 1L;
         final UUID sessionID = UUID.randomUUID();
         jdbcTemplate.update(
-                "insert into version_information (source_id, version, session_id) " +
-                        "values (?, ?, ?)",
```
maggarwal13, line 60: Formatting. we use he convention of uppercase for reserved keywords like select, insert. It makes it more readable

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/NrtmVersionInformationDao.java
```diff
@@ -0,0 +1,92 @@
...
+import javax.sql.DataSource;
+import java.sql.PreparedStatement;
+import java.sql.Statement;
+import java.util.Optional;
+import java.util.UUID;
+
+
+@Repository
+public class NrtmVersionInformationDao {
+
```
maggarwal13, line 23: rename to NrtmVersionInfoDao?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/NrtmVersionInformationDao.java
```diff
@@ -0,0 +1,92 @@
...
+    @Autowired
+    public NrtmVersionInformationDao(@Qualifier("nrtmDataSource") final DataSource dataSource) {
+        this.jdbcTemplate = new JdbcTemplate(dataSource);
+    }
+
+    public Optional<VersionInformation> findLastVersion(final NrtmSource source) {
+        try {
+            return Optional.ofNullable(jdbcTemplate.queryForObject(
+                    "select vi.id, src.name, vi.version, vi.session_id, vi.type " +
```
maggarwal13, line 43: no need for optional.ofNullable if it throws empty result exception?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/VersionInformation.java
```diff
@@ -41,4 +45,12 @@ public UUID getSessionID() {
         return sessionID;
     }
 
+    public NrtmDocumentType getType() {
+        return type;
+    }
+
+    public VersionInformation increment() {
+        return new VersionInformation(0L, this.source, this.version + 1, this.sessionID, this.type);
+    }
```
maggarwal13, line 54: I would prefer to not have this logic here and keep it as plain pojo. This will be used for delta files as well and version increment will depend of documentType.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NotificationFileGenerationService.java
```diff
@@ -0,0 +1,43 @@
...
+        // Get last version from database.
+        final Optional<VersionInformation> lastVersion = versionDao.findLastVersion(source);
+        VersionInformation version;
+        if (lastVersion.isEmpty()) {
+            version = versionDao.createNew(source);
+        } else {
+            version = lastVersion.get();
+            // If it's a delta then use this version. If it's a snapshot then increment it.
+            if (version.getType() == NrtmDocumentType.snapshot) {
+                version = versionDao.save(version.increment());
```
maggarwal13, line 36: Since the function is of generating SnapshotFile type will always be of snapshot? Do we need this else condition? 
May be throw an error for if its not snapshot?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NotificationFileGenerationService.java
```diff
@@ -0,0 +1,43 @@
...
+            }
+        }
+        final SnapshotFile snapshotFile = new SnapshotFile(version);
+        return snapshotFile.setObjectsString("");
```
maggarwal13, line 40: Seems like it is possible to generate a snapshot file using versionType delta?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NotificationFileGenerationService.java
petchells: usually yes, i was thinking about making it a builder (to make SnapshotFile immutable) so this is somewhere in between. i'll change it in the follow-up branch.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/NrtmDocumentType.java
petchells: the RFC specifies lower case -- so instead of making the Enum upper and then adding a method just to get the same string in lower case i just defined it as lower in the first place. intellij is happy with it so i just forgot about it. there will also be "delta" as well, so we'll get back to this code and change it then.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/NrtmVersionDao.java
petchells: there's already a VersionInfo in common, which the nrtm module depends on. keep this on the backburner, though -- VersionInformation is quite long and not all that descriptive. there's probably a better name for this class (NrtmPublishableDocumentVersion? nope, still too long. but you get the idea).

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/NrtmVersionDao.java
petchells: sure. i just like lower case for everything ;)

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/NrtmVersionInformationDao.java
petchells: true

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/VersionInformation.java
petchells: i'm not really happy with this either. let's figure out a better place for it. but actually it will /only/ be used for delta files -- snapshots don't increment the version, they just use the one from the last delta. if there are no deltas, then we don't publish a new snapshot (though i think we might still produce a new update notification for the same one).

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NotificationFileGenerationService.java
petchells: see comment above -- we won't actually need to do anything with the version if the last published file is a snapshot.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NotificationFileGenerationService.java
petchells: yes -- this is how i've coded it in the latest branch.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/NrtmVersionInformationDao.java
petchells: actually changed in later PR to NrtmVersionInfoRepository -- coz i think it's cleaner to reflect the `@Repository` annotation in the name and also because DAO is logically separate in that it can use a repository to do data manipulation but not necessarily directly coupled to the db. think of it like a data service, or a thin layer on top of the database.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NotificationFileGenerationService.java
maggarwal13: Ok, may be we get back to this when we start with other stories

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/NrtmVersionInformationDao.java
maggarwal13: Not really you are sending the optional.empty in case of empty result exception

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/NrtmVersionDao.java
maggarwal13: I guess you have to start pressing few more keys till Intellij becomes smart enough ;)

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/RpslAttributeTransformer.java
maggarwal13: Agree with you, I guess it too much. if we need more logic in filtering we can revisit this and create a separate class but for now its best to call directly

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/NrtmDocumentType.java
MiguelAHM: I think that we should use the uppercase adding a value in lower case

maggarwal13: Javadocs suggest constants should always be in uppercase. https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NotificationFileGenerationService.java
maggarwal13: Ok, I guess this will be revisited when we add delta file type. Thanks for explaining

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/VersionInformation.java
maggarwal13: When we have a logic in place for delta file I think we will have a clearer picture. Lets resolve this in later issues

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NotificationFileGenerationService.java
```diff
@@ -0,0 +1,43 @@
...
+@Service
+public class NotificationFileGenerationService {
+
+    private final NrtmVersionInformationDao versionDao;
+
+    public NotificationFileGenerationService(
+            final NrtmVersionInformationDao nrtmVersionInformationDao
+    ) {
+        versionDao = nrtmVersionInformationDao;
+    }
```
maggarwal13, line 22: this.versionDao? Also rename the variable to nrtmVersionDao ?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/NrtmVersionInformationDao.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.Optional;
+import java.util.UUID;
+
+
+@Repository
+public class NrtmVersionInformationDao {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(NrtmVersionInformationDao.class);
+    private final JdbcTemplate jdbcTemplate;
+    private final RowMapper<VersionInformation> rowMapper = (rs, rowNum) ->
```
maggarwal13, line 26: Do you think it's better to create a VersionInfoRowMapper class ? It's more clearer to me rather than creating a private field for it. It might also be needed in few other classes.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/NrtmVersionInformationDao.java
MiguelAHM: It is a bit odd here to catch an exception that we don't want to throw.
I suggest 
* to catch DataAccessException exception and return an error in that case. 
* move that error (LOGGER.debug) to NotificationFileGenerationService behind lastVersion.isEmpty() 
I don't understand why we are returning "null" or empty value in case that we receive an error from the query.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/publish/NrtmObject.java
```diff
@@ -0,0 +1,5 @@
+package net.ripe.db.nrtm4.publish;
+
+public class NrtmObject {
```
MiguelAHM, line 3: empty class?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/publish/PublishableNrtmDocument.java
```diff
@@ -0,0 +1,48 @@
...
+import net.ripe.db.nrtm4.persist.NrtmDocumentType;
+import net.ripe.db.nrtm4.persist.NrtmSource;
+import net.ripe.db.nrtm4.persist.VersionInformation;
+
+
+public abstract class PublishableNrtmDocument {
+
+    @JsonProperty("nrtm_version")
+    private int nrtmVersion = 4;
+    private NrtmDocumentType type;
```
MiguelAHM, line 13: final

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/publish/PublishableNrtmDocument.java
```diff
@@ -0,0 +1,48 @@
...
+    private NrtmSource source;
```
MiguelAHM, line 14: final

whois-nrtm4/src/test/java/net/ripe/db/nrtm4/NotificationFileGenerationServiceIntegrationTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+@Tag("IntegrationTest")
+@ContextConfiguration(locations = {"classpath:applicationContext-nrtm4-test.xml"})
+public class NotificationFileGenerationServiceIntegrationTest extends AbstractDatabaseHelperIntegrationTest {
+
+    @Autowired
+    private NotificationFileGenerationService notificationFileGenerationService;
+
+    @Autowired
+    private NrtmVersionInformationDao nrtmVersionInformationDao;
```
MiguelAHM, line 27: not used in the test class

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NotificationFileGenerationService.java
```diff
@@ -0,0 +1,43 @@
...
+
+    // TODO: Add a global lock to ensure that no other instance can run until this method exits
+    public SnapshotFile generateSnapshot(final NrtmSource source) {
```
MiguelAHM, line 25: I know that we are using IT for testing it, but I think that for this kind of code (code with some logic) and add unit tests as well.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/NrtmVersionInformationDao.java
petchells: the idea of keeping it private is to hint that all version row mapping should be in this class. and the idea of putting it all in this class is to be sure that it doesn't get used or mutated in ways that aren't written here -- coz to do so would cause big problems.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/NrtmVersionInformationDao.java
```diff
@@ -0,0 +1,92 @@
...
+        } catch (final EmptyResultDataAccessException ex) {
+            LOGGER.debug("findLastVersion threw exception: {}", ex.getMessage());
+            return Optional.empty();
+        }
+    }
+
+    public VersionInformation createNew(final NrtmSource source) {
+
+        jdbcTemplate.update("insert into source (name) values (?)", source.name());
+        final long version = 1L;
```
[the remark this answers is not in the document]
petchells: this is fixed in the 'transform-rpsl-to-nrtm-json' branch

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/publish/NrtmObject.java
petchells: yes, it's removed in 'transform-rpsl-to-nrtm-json'

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/publish/PublishableNrtmDocument.java
petchells: how dare you!

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NotificationFileGenerationService.java
petchells: i agree. i've done some refactoring in 'transform-rpsl-to-nrtm-json' to make it easier to unit test parts of the implementation. there could, and *should* be more though. unit test will be better than integration tests for some things, but they're also fragile coz they depend heavily on the code, and i'm not sure how all of that will look yet.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/NrtmVersionDao.java
petchells: done in 'transform-rpsl-to-nrtm-json'

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NotificationFileGenerationService.java
petchells: done in 'transform-rpsl-to-nrtm-json'

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/NrtmDocumentType.java
petchells: pff. what do they know about writing Java code. changed in 'transform-rpsl-to-nrtm-json'

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NotificationFileGenerationService.java
MiguelAHM: yes, this is why I think that we should add (never replace) to IT in this case I mean

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/VersionInformation.java
petchells: done

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/WhoisSlaveDao.java
```diff
@@ -0,0 +1,32 @@
...
+
+    public WhoisSlaveDao(
+        final WhoisSlaveRepository whoisSlaveRepository
+    ) {
+        this.whoisSlaveRepository = whoisSlaveRepository;
+    }
+
+    public List<Pair<SerialModel, RpslObjectModel>> findSerialsAndObjectsSinceSerial(final int serialId) {
+        final List<Pair<SerialModel, RpslObjectModel>> lastObjects = whoisSlaveRepository.findSerialsInLastSince(serialId);
+        final List<Pair<SerialModel, RpslObjectModel>> historyObjects = whoisSlaveRepository.findSerialsInHistorySince(serialId);
```
maggarwal13, line 24: we are already have a versionDao in whois-commons. Can we use it instead creating new?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/WhoisSlaveRepository.java
```diff
@@ -0,0 +1,61 @@
...
+        "s.serial_id, s.object_id, s.sequence_id, s.atlast, s.operation, " +
+        "r.object_id, r.sequence_id, r.timestamp, r.object_type, r.object, r.pkey " +
+        "FROM serials s JOIN %s r ON r.object_id = s.object_id and r.sequence_id = s.sequence_id " +
+        "WHERE s.serial_id > ? " +
+        "ORDER BY s.serial_id ASC";
+
+    private final RowMapper<Pair<SerialModel, RpslObjectModel>> serialAndRpslObjectMapper = (rs, n) ->
+        new Pair<>(
+            new SerialModel(
+                rs.getInt(1),
```
maggarwal13, line 32: We already have a bean name VersionInfo bean for the same (SerialModel) in whois-commons, can you have a look at the  existing code? May be refactor or shift this code to versionDao instead?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/RpslObjectModel.java
```diff
@@ -0,0 +1,56 @@
+package net.ripe.db.nrtm4.persist;
+
+import net.ripe.db.whois.common.rpsl.ObjectType;
+import net.ripe.db.whois.common.rpsl.RpslObject;
+
+
+public class RpslObjectModel {
+
```
maggarwal13, line 8: check RpslObjectUpdateInfo bean, I think it's the same you are trying to use. We should re-use the exisiting beans and methods where we can.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/DeltaChange.java
```diff
@@ -0,0 +1,50 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import net.ripe.db.whois.common.rpsl.ObjectType;
+import net.ripe.db.whois.common.rpsl.RpslObject;
+
+
+public class DeltaChange {
+
+    enum Action {
+        delete,
+        add_modify,
```
maggarwal13, line 12: Enum values should be in uppercase

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/RpslObjectTransformer.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    RpslObject filter(final RpslObject rpslObject) {
+        // do we need to convert latin1 to utf8? nrtm db is utf8 (coz we serve json as utf8)
+        return dummifierNrtm.dummify(NRTM_VERSION, rpslObject);
+    }
+
+    boolean isAllowed(final RpslObject rpslObject) {
+        return dummifierNrtm.isAllowed(NRTM_VERSION, rpslObject);
+    }
```
maggarwal13, line 29: Calling dummifierNrtm directly is way more easier and readable then calling rpslObjectTransformer?

whois-commons/src/main/resources/nrtm_schema.sql
```diff
@@ -12,41 +12,68 @@
 DROP TABLE IF EXISTS `source`;
 create table `source`
 (
```
maggarwal13, line 14: We usually put all our sql files including patch files in whois-commons resources.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/NrtmDocumentType.java
```diff
@@ -0,0 +1,10 @@
+package net.ripe.db.nrtm4.persist;
+
+public enum NrtmDocumentType {
+    DELTA,
+    SNAPSHOT;
+
+    public String nameToLowerCase() {
+        return name().toLowerCase();
```
maggarwal13, line 8: If we need the value in lowercase always, we should override toString method, so whenever we call just DELTA , it will automatically returns me the lowercase instead of remembering to call nameToLowerCase. But feel free to ignore my comment.

whois-nrtm4/src/test/java/net/ripe/db/nrtm4/NrtmProcessorIntegrationTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @Test
+    public void test_whois_slave_dao() {
+        loadSerials("nrtm_sample.sql");
+        final List<Pair<SerialModel, RpslObjectModel>> changes = whoisSlaveDao.findSerialsAndObjectsSinceSerial(0);
+        assertThat(changes.size(), is(64));
+    }
+
+    @Test
+    public void test_delta_file_cannot_be_generated() {
```
MiguelAHM, line 69: This test is not working for me

whois-nrtm4/src/test/java/net/ripe/db/nrtm4/NrtmProcessorIntegrationTest.java
```diff
@@ -0,0 +1,84 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+
+
+@Tag("IntegrationTest")
+@ContextConfiguration(locations = {"classpath:applicationContext-nrtm4-test.xml"})
+public class NrtmProcessorIntegrationTest extends AbstractDatabaseHelperIntegrationTest {
+
+    @Autowired
+    private DeltaProcessor deltaProcessor;
```
MiguelAHM, line 32: don't need this autowire

whois-nrtm4/src/test/java/net/ripe/db/nrtm4/NrtmProcessorIntegrationTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @Autowired
```
MiguelAHM, line 34: don't need this auto wire

whois-nrtm4/src/test/java/net/ripe/db/nrtm4/NotificationFileGenerationServiceIntegrationTest.java
```diff
@@ -0,0 +1,62 @@
...
+            final PublishableSnapshotFile publishableSnapshotFile = notificationFileGenerationService.generateSnapshot(nrtmSourceHolder.getSource());
+            assertThat(publishableSnapshotFile.getVersion(), is(1L));
+            sessionID = publishableSnapshotFile.getSessionID();
+            assertThat(publishableSnapshotFile.getSource(), is(nrtmSourceHolder.getSource()));
+            assertThat(publishableSnapshotFile.getNrtmVersion(), is(4));
+            assertThat(publishableSnapshotFile.getType(), is(SNAPSHOT.nameToLowerCase()));
+        }
+        {
+            final PublishableSnapshotFile publishableSnapshotFile = notificationFileGenerationService.generateSnapshot(nrtmSourceHolder.getSource());
+            assertThat(publishableSnapshotFile.getVersion(), is(2L));
```
MiguelAHM, line 46: I have an error here:
Expected :is <2L>
Actual   :<1L>

MiguelAHM: NotificationFileGenerationService -> line 33 you get the lastVersion, you are not incrementing it because deltas are not changing, so you need to modify the test or to create some delta in the middle

whois-nrtm4/src/test/java/net/ripe/db/nrtm4/NrtmProcessorIntegrationTest.java
```diff
@@ -0,0 +1,84 @@
...
+        loadSerials("nrtm_sample_sm.sql");
+        insertSnapshot();
+        final DeltaFileModel deltas = nrtmProcessor.generateDeltaFile(NrtmSourceHolder.valueOf("TEST"));
+        assertThat(deltas.getPayload(), is(sampleSm));
+    }
+
+    private String sampleSm = "[{\"action\":\"delete\",\"object_class\":\"AUT_NUM\",\"primary_key\":\"AS8928\"},{\"action\":\"add_modify\",\"object\":\"inetnum:        195.77.187.144 - 195.77.187.151\\nnetname:        Netname\\ndescr:          Description\\ncountry:        es\\nadmin-c:        DUMY-RIPE\\ntech-c:         DUMY-RIPE\\nstatus:         ASSIGNED PA\\nmnt-by:         MAINT-AS3352\\nsource:         RIPE\\nremarks:        ****************************\\nremarks:        * THIS OBJECT IS MODIFIED\\nremarks:        * Please note that all data that is generally regarded as personal\\nremarks:        * data has been removed from this object.\\nremarks:        * To view the original object, please query the RIPE Database at:\\nremarks:        * http://www.ripe.net/whois\\nremarks:        ****************************\\n\",\"object_class\":\"INETNUM\",\"primary_key\":\"195.77.187.144 - 195.77.187.151\"}]";
```
MiguelAHM, line 83: should be final static right?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmProcessor.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    public DeltaFileModel generateDeltaFile(final NrtmSource source) {
+
+        // Find changes since the last delta
+        final Optional<VersionInformation> lastVersion = nrtmVersionInfoRepository.findLastVersion(source);
+        if (lastVersion.isEmpty()) {
+            throw new IllegalStateException("Cannot create a delta without an initial snapshot");
+        }
+        final List<Pair<SerialModel, RpslObjectModel>> whoisChanges = whoisSlaveDao.findSerialsAndObjectsSinceSerial(lastVersion.get().getLastSerialId());
```
MiguelAHM, line 64: it is always more clear to use an object than a pair. You could have a list of a descriptive objects instead of a list of pairs I mean

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/WhoisSlaveRepository.java
petchells: it doesn't have the serial_id column, which we need for ordering the changes. maybe it's possible to extend so i can add the extra column. but there's sth odd about that class -- it implements `identifiable` on the object_id column but that table is not unique on object_id. let's discuss it.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/NrtmDocumentType.java
petchells: no, that's a good idea. i'll do that.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/RpslObjectModel.java
petchells: it's missing the `object` field which contains the RPSL. I've moved it to common and extended RpslObjectUpdateInfo instead.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/DeltaChange.java
petchells: done

whois-commons/src/main/resources/nrtm_schema.sql
petchells: yes, that's where it is already :) or do you mean actually in the 'patch' directory?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmProcessor.java
petchells: for me it's enough to see the <type> but ok -- implemented as an explicit containing class

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/WhoisSlaveDao.java
petchells: yes, i moved the code there

whois-commons/src/main/java/net/ripe/db/whois/common/dao/jdbc/JdbcVersionDao.java
```diff
@@ -144,4 +175,15 @@ public List<VersionInfo> getVersionsForTimestamp(final ObjectType type, final St
                 "WHERE temp.timestamp=:cutoffTime ", parameters, new VersionInfoRowMapper()));
         return versionInfos;
     }
+
+    public List<SerialRpslObjectTuple> findSerialsInLastSince(final int serialId) {
+        final String sql = String.format(sqlSerialAndRpslTemplate, "last");
+        return jdbcTemplate.query(sql, serialAndRpslObjectMapper, serialId);
+    }
```
maggarwal13, line 182: Can we create a method which returns result of jdbcTemplate.query(sql, serialAndRpslObjectMapper, serialId); and pass on the "last" or "history" ?
This way we can get rid of unnecessary sql query constant at the top

maggarwal13: Also this way we can use the serialAndRpslObjectMapper directly

whois-commons/src/main/java/net/ripe/db/whois/common/dao/VersionInfo.java
```diff
@@ -5,23 +5,18 @@
...
-    private final int objectId;
-    private final int sequenceId;
+
+public class VersionInfo extends Serial implements Identifiable, Comparable<VersionInfo> {
+
     private final VersionDateTime timestamp;
-    private final Operation operation;
 
     public VersionInfo(final boolean inLast, final int objectId, final int sequenceId, final long timestamp, final Operation operation) {
-        this.inLast = inLast;
```
maggarwal13, line 16: If versionInfo extends new class Serial, which means serialId is also available to versionInfo? and this way equals method and toString method of versionInfo now is wrongly implemented? Or am I missing something?

whois-commons/src/main/java/net/ripe/db/whois/common/dao/jdbc/SerialRpslObjectTuple.java
```diff
@@ -0,0 +1,29 @@
+package net.ripe.db.whois.common.dao.jdbc;
+
+import net.ripe.db.whois.common.dao.RpslObjectModel;
+import net.ripe.db.whois.common.dao.Serial;
+
+
+public class SerialRpslObjectTuple {
+
+    private final Serial serial;
+    private final RpslObjectModel rpslObjectModel;
```
maggarwal13, line 10: I think what you are looking for is SerialEntry bean. Please go through existing NRTM code , we have almost all he beans we need. I am not sure why SerialEntry does not contain serial_id but please feel free to add it there as its already available. Let's discuss over call

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/WhoisSlaveDao.java
```diff
@@ -0,0 +1,33 @@
...
+import org.springframework.stereotype.Service;
+
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+
+
+@Service
+public class WhoisSlaveDao {
+
```
maggarwal13, line 14: Name is very generic and seems like connecting to readOnlyWhois datasource but its actually versionDoa and should not be in nrtm module?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/RpslObjectTransformer.java
petchells: done

whois-commons/src/main/java/net/ripe/db/whois/common/dao/jdbc/JdbcVersionDao.java
petchells: done

whois-commons/src/main/java/net/ripe/db/whois/common/dao/VersionInfo.java
petchells: good point. added serialId to equals and hash methods (but not compareTo, or sorting would be different)

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/WhoisSlaveDao.java
petchells: the objects it gets and the way they are sorted are specific for NRTM. i think the name could be better, though, it's a hangover from when it did sth a bit different but now it's not good. i'll change the name.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/WhoisSlaveRepository.java
petchells: as discussed -- it doesn't have to be unique to be 'Identifiable' (although it's used in the code like it is!). adding serial_id to the SELECT doesn't have any effect anyway.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/DeltaChange.java
```diff
@@ -0,0 +1,57 @@
...
+
+public class DeltaChange {
+
+    @JsonSerialize(using = EnumToStringSerializer.class)
+    enum Action {
+        DELETE,
+        ADD_MODIFY;
+
+        @Override
+        public String toString() {
```
maggarwal13, line 17: I guess you can directly call  @JsonValue on a getter method instead of creating a custom serializer?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/DeltaProcessor.java
```diff
@@ -0,0 +1,44 @@
...
+    public DeltaProcessor(final Dummifier dummifierNrtm) {
+        this.dummifierNrtm = dummifierNrtm;
+    }
+
+    List<DeltaChange> process(final List<SerialEntry> changes) {
+        return changes.stream()
+            .filter(change -> dummifierNrtm.isAllowed(NRTM_VERSION, change.getRpslObject()))
+            .map(serialRpsl -> {
+                if (serialRpsl.isAtLast()) {
+                    return new DeltaChange(
```
maggarwal13, line 27: This should be based on serialRpsl .operation?

whois-commons/src/main/java/net/ripe/db/whois/common/domain/serials/SerialEntry.java
```diff
@@ -24,10 +25,20 @@ public SerialEntry(final Operation operation, final boolean atLast, final int ob
...
     }
 
+    public SerialEntry(final int serialId, final Operation operation, final boolean atLast, final int objectId, final byte[] blob) {
+        this(operation, atLast, objectId, 0, 0, blob);
+        rpslObject = RpslObject.parse(objectId, blob);
+        this.serialId = serialId;
+    }
+
     public static SerialEntry createSerialEntryWithoutTimestamps(final Operation operation, final boolean atLast, final int objectId, final byte[] blob){
         return new SerialEntry(operation, atLast, objectId, 0, 0, blob);
```
maggarwal13, line 35: Serial id here? Also I do not thunk we need timestamp in that case use this constructor?

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/DummifierNrtm.java
```diff
@@ -9,12 +9,12 @@
...
 
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
-@Component
+@Component("dummifierNrtm")
 public class DummifierNrtm implements Dummifier {
```
maggarwal13, line 18: Do we need this change?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/VersionInformation.java
```diff
@@ -2,27 +2,34 @@
 
 import java.util.UUID;
 
+
 public class VersionInformation {
```
maggarwal13, line 7: can we rename to NrtmVersionInfo as rest of classes includes Nrtm as prefix and also it will avoid confusion with existing versionInfo

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/NrtmVersionInfoRepository.java
```diff
@@ -0,0 +1,128 @@
...
+
+
+@Repository
+public class NrtmVersionInfoRepository {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(NrtmVersionInfoRepository.class);
+    private final JdbcTemplate jdbcTemplate;
+    private final RowMapper<VersionInformation> rowMapper = (rs, rowNum) ->
+        new VersionInformation(
+            rs.getLong(1),
```
maggarwal13, line 28: You have duplicated this rowMapper in repository layer as well. May be create a class out of it? And also can we have all insert or select statement in Dao, its easier to maintain at one class

whois-nrtm4/src/test/java/net/ripe/db/nrtm4/NrtmProcessorIntegrationTest.java
```diff
@@ -0,0 +1,67 @@
...
+import static org.hamcrest.Matchers.is;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+
+@Tag("IntegrationTest")
+@ContextConfiguration(locations = {"classpath:applicationContext-nrtm4-test.xml"})
+public class NrtmProcessorIntegrationTest extends AbstractDatabaseHelperIntegrationTest {
+
+    @Autowired
+    private RpslObjectDao rpslObjectDao;
```
MiguelAHM, line 27: I think that you are not using this @Autowired

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/DeltaChange.java
```diff
@@ -0,0 +1,57 @@
+package net.ripe.db.nrtm4;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonValue;
+import net.ripe.db.whois.common.rpsl.ObjectType;
+import net.ripe.db.whois.common.rpsl.RpslObject;
+
+
+public class DeltaChange {
+
```
maggarwal13, line 10: I think this is the class we are going to publish? Can you add JsonInclude.Include.NON_NULL instead of payloadProcessor it's easier to see on bean level.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/DeltaChange.java
```diff
@@ -0,0 +1,57 @@
...
+        }
+    }
+
+    private final Action action;
+    @JsonProperty("object_class")
+    private final ObjectType objectClass;
+    @JsonProperty("primary_key")
+    private final String primaryKey;
+    private final RpslObject object;
+
```
maggarwal13, line 28: You do not need to add a custom serializer in payloadProcessor, we can get rid of that class all together ? You can directly put annotation  here and write a converter here  @JsonSerialize(converter = RplsObjectSerializer.class).  You need to extend StdConverter.

whois-commons/src/main/java/net/ripe/db/whois/common/dao/jdbc/SerialRpslObjectTuple.java
maggarwal13: Already changed in new commit

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmProcessor.java
```diff
@@ -0,0 +1,80 @@
...
+        // Find changes since the last delta
+        final Optional<NrtmVersionInfo> lastVersion = nrtmVersionInfoRepository.findLastVersion(source);
+        if (lastVersion.isEmpty()) {
+            throw new IllegalStateException("Cannot create a delta without an initial snapshot");
+        }
+        final List<SerialEntry> whoisChanges = serialDao.getSerialEntriesSince(lastVersion.get().getLastSerialId());
+
+        final List<DeltaChange> deltas = deltaProcessor.process(whoisChanges);
+        snapshotSynchronizer.synchronizeDeltasToSnapshot(deltas);
+        if (whoisChanges.size() < 1) {
```
MiguelAHM, line 66: could you move this to line 63? just after retrieving the serials from last entry? we don't need to get the deltas and perform the sync if whoisChanges is empty, right?

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/DummifierNrtm.java
petchells: yes, because there's more than one implementation of Dummifier

whois-nrtm4/src/test/java/net/ripe/db/nrtm4/NrtmProcessorIntegrationTest.java
petchells: fixed

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmProcessor.java
petchells: done

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/DeltaChange.java
petchells: i tried that approach earlier but was getting NPE from Jackson on the objectId field. i'm not sure how to get it working coz the config looks ok to me. can we come back to this when we have a solution? it has a very low impact on the code.

MiguelAHM (approved):

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/DeltaChange.java
petchells: done

PR #1112, opened 2022-11-15
Rdap events and notices

MiguelAHM:


whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -130,20 +130,18 @@ public RdapObjectMapper(
...
 
-    public Object mapSearch(final String requestUrl, final List<RpslObject> objects, final Iterable<LocalDateTime> localDateTimes, final int maxResultSize) {
+    public Object mapSearch(final String requestUrl, final List<RpslObject> objects, final int maxResultSize) {
         final SearchResult searchResult = new SearchResult();
-        final Iterator<LocalDateTime> iterator = localDateTimes.iterator();
 
         for (final RpslObject object : objects) {
             if (object.getType() == DOMAIN) {
-                searchResult.addDomainSearchResult((Domain) getRdapObject(requestUrl, object, iterator.next(), Optional.empty()));
+                searchResult.addDomainSearchResult((Domain) getRdapObject(requestUrl, object, Optional.empty()));
```
eshryane, line 142: Make sure there are tests for both these cases (i.e. result is read from the RIPE database directly, or via the Lucene index).

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -215,7 +212,15 @@ private RdapObject getRdapObject(final String requestUrl,
             rdapResponse.getRemarks().add(createRemark(rpslObject));
         }
 
-        rdapResponse.getEvents().add(createEvent(lastChangedTimestamp));
+        if (rpslObject.containsAttribute(AttributeType.CREATED)){
```
eshryane, line 215: These attributes *must* always appear in an RPSL object (they are mandatory), i.e. remove the 'if' check

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapObjectMapperTest.java
```diff
@@ -566,11 +568,11 @@ public void organisation_should_not_have_notice_for_language() {
     @Test
     public void mapSearch_twoObjects() {
         final List<RpslObject> objects = Lists.newArrayList(
-                RpslObject.parse("organisation: ORG-TOL-TEST\norg-name: Test Organisation\nstatus: OTHER\ndescr: comment 1\nsource: TEST"),
-                RpslObject.parse("organisation: ORG-TST-TEST\norg-name: Test Company\nstatus: OTHER\ndescr: comment 2\nsource: TEST")
+                RpslObject.parse("organisation: ORG-TOL-TEST\norg-name: Test Organisation\nstatus: OTHER\ndescr: comment 1\nsource: TEST\ncreated: 1996-02-05T03:52:05.938\nlast-modified: 1970-04-14T09:22:14.857"),
```
petchells, line 571: missing timezone in the date strings

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
MiguelAHM: Yeah, I know. However: 
Each response that includes an RDAP object as the topmost object SHOULD include a "registration" event and a "last changed" event for that object. This may not be possible for all objects: for example, event data for legacy IP network objects is often missing or unreliable.

This may not be possible for all objects: for example, event data for legacy IP network objects is often missing or unreliable.

Maybe the legacy object does not have this object. I know that it is autogenerated attributes but maybe there are some data in database that do not have this data, could be that possible?

eshryane: Every existing object in the database contains both created and last-modified attributes, so this check should never fail, and if it does it's a bug

MiguelAHM: I'm going to add a check and throw an error if this happens.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
MiguelAHM: I am looking at WhoisRdapServieTestIntegration and there are some tests that are already using FullTextIndex.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -215,7 +212,12 @@ private RdapObject getRdapObject(final String requestUrl,
             rdapResponse.getRemarks().add(createRemark(rpslObject));
         }
 
-        rdapResponse.getEvents().add(createEvent(lastChangedTimestamp));
+        if (!rpslObject.containsAttribute(AttributeType.CREATED) || !rpslObject.containsAttribute(AttributeType.LAST_MODIFIED)){
```
eshryane, line 215: No need for this block, `getValueForAttribute` will throw an exception if it can't find the attribute

eshryane (approved):

PR #1115, opened 2022-11-17
Fix Filtered Jetty Logging Integration Tests

eshryane:
JettyRequestLogTestIntegration tests do not like to run in parallel, as the Jetty logfile is cleared after every test when a different test is reading from it. Add a workaround to run each test separately.


MiguelAHM (approved):

eshryane:
Integration tests passed 3 times before merging.

PR #1118, opened 2022-11-22
feat: add trailing dot in domains ldh name

MiguelAHM:


whois-rpsl/src/main/java/net/ripe/db/whois/common/ip/IpInterval.java
```diff
@@ -15,6 +15,13 @@ public static String removeTrailingDot(final String address) {
         return address;
     }
 
+    public static String addTrailingDot(final String address){
```
eshryane, line 18: The "trailing dot" signifies the (reverse) domain root, it doesn't belong in an IpInterval. I suggest adding logic to the Domain class to always return a trailing dot.

eshryane: Correction, `removeTrailingDot` is already in IpInterval! So I guess `addTrailingDot` is fine there too 🤷

eshryane (approved):

PR #1119, opened 2022-11-22
Db 4316 add remaining profiles

MiguelAHM:


whois-api/src/test/java/net/ripe/db/whois/api/rdap/WhoisRdapServiceTestIntegration.java
```diff
@@ -237,7 +237,8 @@ public void lookup_inetnum_range_reserved() {
 
         assertThat(ip.getPort43(), is("whois.ripe.net"));
         assertThat(ip.getRdapConformance(), hasSize(2));
-        assertThat(ip.getRdapConformance(), containsInAnyOrder("rdap_level_0", "cidr0"));
+        assertThat(ip.getRdapConformance(), containsInAnyOrder(RdapConformanceEnum.LEVEL_0.name(),
```
eshryane, line 240: I suggest comparing with an expected **string** value, not the enum name itself (i.e. what will the string look like? Is it uppercase?)

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapConformanceEnum.java
```diff
@@ -0,0 +1,7 @@
+package net.ripe.db.whois.api.rdap;
+
+public enum RdapConformanceEnum {
```
eshryane, line 3: No need for 'Enum' in the class name

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapConformanceEnum.java
```diff
@@ -0,0 +1,7 @@
...
+    CIRD_0("cidr0"), LEVEL_0("rdap_level_0"), FLAT_MODEL("nro_rdap_profile_asn_flat_0");
+
+    RdapConformanceEnum(final String value){}
```
eshryane, line 6: Can you store the value in a field? I think these are the expected values that should appear in the RDAP response?

whois-api/src/test/java/net/ripe/db/whois/api/rdap/WhoisRdapServiceTestIntegration.java
eshryane: Just tested that "CIRD_0" appears in the response rather than the expected "cidr_0"

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapConformanceEnum.java
```diff
@@ -0,0 +1,7 @@
+package net.ripe.db.whois.api.rdap;
+
+public enum RdapConformanceEnum {
+    CIRD_0("cidr0"), LEVEL_0("rdap_level_0"), FLAT_MODEL("nro_rdap_profile_asn_flat_0");
```
eshryane, line 4: Typo: replace CIRD_0 with CIDR_0

whois-api/src/test/java/net/ripe/db/whois/api/rdap/WhoisRdapServiceTestIntegration.java
MiguelAHM: yes, I am taking the name() instead of the value. I fixed

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapConformance.java
```diff
@@ -0,0 +1,15 @@
+package net.ripe.db.whois.api.rdap;
+
+public enum RdapConformance {
+    CIDR_0("cidr0"), LEVEL_0("rdap_level_0"), FLAT_MODEL("nro_rdap_profile_asn_flat_0");
```
eshryane, line 4: Do we **also** need nro_rdap_profile_0 ? Or do we not meet the requirements yet? Is that mentioned in the document?

I just found the extension registry: https://www.iana.org/assignments/rdap-extensions/rdap-extensions.xhtml

MiguelAHM: To signify conformance with this specification, the string literal
   "nro_rdap_profile_0" MUST be included as an rdapConformance value in
   each response.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -222,7 +226,7 @@ private RdapObject getRdapObject(final String requestUrl,
             case INETNUM:
             case INET6NUM:
                 rdapResponse = createIp(rpslObject);
-                rdapResponse.getRdapConformance().add("cidr0");
+                rdapResponse.getRdapConformance().add(RdapConformance.CIDR_0.getValue());
```
eshryane, line 229: Add *all* extensions here?

MiguelAHM: we must include cidr0 in all the responses, not just in INET6NUM:
https://bitbucket.org/nroecg/nro-rdap-profile/raw/v1/nro-rdap-profile.txt

4.1.
   The service MUST implement the "cidr0" extension (registered with
   IANA), so the string literal "cidr0" MUST be included as an
   rdapConformance value in **each response.**

MiguelAHM: after converting the object from RPSL to Rdap we are calling mapCommons, so we are still adding all the extensions but the FLAT_MODEL. 
We need CIDR_0 in all the responses so I will move to mapCommons in order to have CIRD_0, LEVEL_0 and PROFILE_0 in all the responses

'''
   Each response MUST include an rdapConformance element.  The service
   MUST conform with RFC 7483 [RFC7483], so "rdap_level_0" will be
   included as an rdapConformance value in each response.

   To signify conformance with this specification, the string literal
   "nro_rdap_profile_0" MUST be included as an rdapConformance value in
   each response.

   The service MUST implement the "cidr0" extension (registered with
   IANA), so the string literal "cidr0" MUST be included as an
   rdapConformance value in each response.
'''

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -469,6 +475,7 @@ private Autnum createAutnumResponse(final RpslObject rpslObject) {
         autnum.setStatus(getResourceStatus(rpslObject));
 
         autnum.getEntitySearchResults().addAll(createContactEntities(rpslObject));
+        autnum.getRdapConformance().add(RdapConformance.FLAT_MODEL.getValue());
         return autnum;
```
maggarwal13, line 479: Did we already implemented nro_rdap_profile_asn_flat_0? I see the DB-4199 still open

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -257,7 +261,9 @@ private RdapObject getRdapObject(final String requestUrl,
     private RdapObject mapCommons(final RdapObject rdapResponse, final String requestUrl) {
         rdapResponse.getNotices().add(noticeFactory.generateTnC(requestUrl));
 
-        rdapResponse.getRdapConformance().addAll(RDAP_CONFORMANCE_LEVEL);
+        rdapResponse.getRdapConformance().addAll(List.of(RdapConformance.CIDR_0.getValue(),
+                RdapConformance.LEVEL_0.getValue(),
```
maggarwal13, line 265: I think cidr0 is only for inetnum and inet6num, this will add  it to every object type? Can you confirm ?

maggarwal13 (changes requested):
please read my comments above

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
maggarwal13: I just read your reply to Ed, seems like it should be included in every response

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
MiguelAHM: right now I think that we are following the flat model:

6.1.2.3.  Flat model

   An RIR that receives a query for an autnum for which it has
   registration authority but that has not been further delegated by the
   RIR to one of its own accounts MUST respond with a "404 Not Found"
   response.

   To signify use of the flat model, the string literal
   "nro_rdap_profile_asn_flat_0" MUST be included as an rdapConformance
   value in each response.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapConformance.java
```diff
@@ -0,0 +1,15 @@
+package net.ripe.db.whois.api.rdap;
+
+public enum RdapConformance {
+    CIDR_0("cidr0"), LEVEL_0("rdap_level_0"), PROFILE_0("nro_rdap_profile_0"), FLAT_MODEL("nro_rdap_profile_asn_flat_0");
+
```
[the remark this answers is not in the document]
MiguelAHM: I am not able to find anything else in the RFC about it, just to add it in all the responses

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -193,7 +196,9 @@ public RdapObject mapError(final int errorCode, final String errorTitle, final L
     }
 
     public RdapObject mapHelp(final String requestUrl) {
-        return mapCommons(new RdapObject(), requestUrl);
+        RdapObject rdapObject =  mapCommons(new RdapObject(), requestUrl);
+        rdapObject.setRdapConformance(COMMON_RDAP_CONFORMANCE);
```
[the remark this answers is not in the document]
MiguelAHM: In /help I am adding FLAT_MODEL as well. Maybe is better to take all the values that we have in RdapConformance, because in /help we will always show the supported profiles that we have, so if in the future we need to add another one we just need to add it in RdapCOnformance and it will automatically add the value in /help

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
maggarwal13: Ok can you comment close the JIRA  DB-4199

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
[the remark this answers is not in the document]
maggarwal13: OK but use the addAll not setter as it will replace the previously added RdapConformance.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapConformance.java
maggarwal13, line 5: nro_rdap_profile_asn_flat_0 is only for Autnum objects. It should not be added in commons Can you confirm?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapConformance.java
```diff
@@ -0,0 +1,15 @@
+package net.ripe.db.whois.api.rdap;
+
+public enum RdapConformance {
```
maggarwal13, line 3: I think this enum is not being used now? There is already a constant list for this?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -95,7 +95,10 @@ class RdapObjectMapper {
     private static final String TERMS_AND_CONDITIONS = "http://www.ripe.net/data-tools/support/documentation/terms";
     private static final Link COPYRIGHT_LINK = new Link(TERMS_AND_CONDITIONS, "copyright", TERMS_AND_CONDITIONS, null, null);
 
-    private static final List<String> RDAP_CONFORMANCE_LEVEL = Lists.newArrayList("rdap_level_0");
+    private static final List<String> COMMON_RDAP_CONFORMANCE =
+            Lists.newArrayList(RdapConformance.PROFILE_0.getValue(), RdapConformance.CIDR_0.getValue(),
+            RdapConformance.LEVEL_0.getValue(),
+            RdapConformance.FLAT_MODEL.getValue());
```
maggarwal13, line 102: nro_rdap_profile_asn_flat_0 is only for Autnum objects. It should not be added in commons Can you confirm? I do not see it mentioned in section 4.1

MiguelAHM: We are not adding flat in common, we are adding it in /help and in autnums. You are right, we should add it just in autnums. This is why I need a set in /help, because it is calling commons but I can try to refactor this and use a set + add

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapConformance.java
MiguelAHM: we are using this enum everywhere instead of in the tests (Ed told me to use directly the profile to make the tests more clear)

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
maggarwal13: Ahh, I got confuse with the name of constant variable COMMON_RDAP_CONFORMANCE. Can you rename to HELP_RDAP_CONFORMANCE?

MiguelAHM: I will delete it, this is not used now in my branch I will directly read all the values from the Enum to have everything in one place

maggarwal13: Since mapCommons is already adding the common rdapConformance may be just add the remaining conformances required for help section?

MiguelAHM: I updated the PR right now

MiguelAHM: In that case if tomorrow we need to add a new profile, we will need to add in this list and in the enum as well. To me it makes not sense, I prefer just to add it in the enum and /help will read all of the values that we have in the enum

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -193,7 +192,9 @@ public RdapObject mapError(final int errorCode, final String errorTitle, final L
     }
 
     public RdapObject mapHelp(final String requestUrl) {
-        return mapCommons(new RdapObject(), requestUrl);
+        final RdapObject rdapObject =  mapCommons(new RdapObject(), requestUrl);
+        rdapObject.setRdapConformance(Lists.newArrayList(RdapConformance.values()).stream().map(RdapConformance::getValue).collect(Collectors.toList()));
+        return rdapObject;
```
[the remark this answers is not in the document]
MiguelAHM: yes true, thank you!

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
maggarwal13: Discussed over chat for refactoring

maggarwal13 (approved):

eshryane:
As discussed with Miguel, the "nro_rdap_profile_asn_flat_0" profile is *only* added on autnum responses.

Otherwise the common profiles on all responses are "cidr0", "rdap_level_0", "nro_rdap_profile_0".

PR #1120, opened 2022-11-22
Fix sonar

MiguelAHM:


petchells (approved):
worked on command line. SQ report looks good.

PR #1125, opened 2022-11-25
feat: do no return as-block if the autnum does not exist

MiguelAHM:


whois-api/src/test/java/net/ripe/db/whois/api/rdap/WhoisRdapServiceTestIntegration.java
```diff
@@ -1067,60 +1067,6 @@ public void lookup_reserved_autnum() {
         assertThat(autnum.getStatus(), contains("reserved"));
     }
 
-    @Test
```
eshryane, line 1070: add a negative test, e.g. create an as-block "AS1 - AS6" then query for AS1, expect 404 not found response (i.e. confirm the as-block is not returned)

maggarwal13 (approved):

MiguelAHM:
https://gitlab.ripe.net/swe-database-team/whois/-/pipelines/281787
Test ok

PR #1126, opened 2022-11-25
Db 4317 fix jcard properties

MiguelAHM:
TODO: 
* adr: the Type may be WORK AND HOME (discuss with the team about the list of empty strings) 
* tel: The Type may be URI or TEXT -> We are already doing this (discuss with the team about "type": "voice") 
* email: The Type may be HOME AND WORK (discuss with the team because now we are using abuse or email and I think that it makes sense) 
* We have to add "language", "altid", and "pref" parameter to the current properties (discuss with the team about which values must I add there) 

adr -> Why we have a list of empty strings? 
tel -> are the types voice/fax already valid? should we continue using it? 
email -> Are the types abuse and email already valid? it seems that we need to use HOME/WORK 
It seems that we need to add an extra parameters "language", "altid" and "pref" to our current vcard's properties. Which should be the values for those new parameters?

### DIFF
--- a/whois-api/src/main/java/net/ripe/db/whois/api/rdap/VCardBuilder.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/rdap/VCardBuilder.java
@@ -1,7 +1,6 @@
 package net.ripe.db.whois.api.rdap;
 
 import com.google.common.base.Joiner;
-import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
 import net.ripe.db.whois.api.rdap.domain.vcard.VCard;
 import net.ripe.db.whois.api.rdap.domain.vcard.VCardKind;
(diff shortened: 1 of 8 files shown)


whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/vcard/VCardType.java
```diff
@@ -3,7 +3,9 @@
 public enum VCardType {
 
     TEXT("text"),
-    URI("uri");
+    URI("uri"),
+    WORK("work");
+
```
maggarwal13, line 8: We should not add work or home in my opinion, we do not know for sure, better to keep it as it is and RFC says  may be values not must or only vales .

whois-nrtm/src/main/resources/rdap.json
```diff
@@ -0,0 +1,73 @@
+{
+  "vcardArray": [
```
maggarwal13, line 2: Are these being used for tests? If yes can you move to test/resources?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/VCardBuilder.java
```diff
@@ -1,97 +1,116 @@
...
+import static java.util.Collections.nCopies;
+import static net.ripe.db.whois.api.rdap.domain.vcard.VCardName.ADDRESS;
+import static net.ripe.db.whois.api.rdap.domain.vcard.VCardName.EMAIL;
+import static net.ripe.db.whois.api.rdap.domain.vcard.VCardName.FN;
+import static net.ripe.db.whois.api.rdap.domain.vcard.VCardName.GEO;
+import static net.ripe.db.whois.api.rdap.domain.vcard.VCardName.KIND;
+import static net.ripe.db.whois.api.rdap.domain.vcard.VCardName.ORG;
+import static net.ripe.db.whois.api.rdap.domain.vcard.VCardName.TELEPHONE;
+import static net.ripe.db.whois.api.rdap.domain.vcard.VCardName.VERSION;
+import static net.ripe.db.whois.api.rdap.domain.vcard.VCardType.TEXT;
```
maggarwal13, line 26: We do not have to include all the vcard properties in json only those that RFC says  must or those that we know for sure.

maggarwal13:
I think we are already compliant  with jcard according to RFC

PR #1129, opened 2022-11-28
Db 4334 solve plain text error

MiguelAHM:


whois-api/src/main/java/net/ripe/db/whois/api/rest/CustomMessageBodyWriter.java
```diff
@@ -0,0 +1,40 @@
...
+import javax.ws.rs.core.MultivaluedMap;
+import javax.ws.rs.ext.MessageBodyWriter;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import java.nio.charset.StandardCharsets;
+
+@Produces({"text/plain"})
+public class CustomMessageBodyWriter implements MessageBodyWriter<WhoisResources> {
```
eshryane, line 18: Use a more specific class name than "Custom", without being too verbose ("WhoisResourcesPlainTextWriter" ?)

whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisServletDeployer.java
```diff
@@ -104,6 +106,9 @@ public void deploy(WebAppContext context) {
         jaxbJsonProvider.configure(SerializationFeature.WRAP_ROOT_VALUE, false);
         resourceConfig.register(jaxbJsonProvider);
 
+        final MessageBodyWriter<WhoisResources> customMessageBodyWriter = new WhoisResourcesPlainTextWriter();
+        resourceConfig.register(customMessageBodyWriter);
+
```
maggarwal13, line 111: This servlet deployer is being used for lot of other services and not every service returns whoisResources. Also this issue exists only when we throw exception else streamingMarshal takes care of it. Check 404 responses are handled properly. May be the best place to make changes is in RestServiceHelper where we create errorEntity for whoisResources but it's just a suggestion.

MiguelAHM: yes, you are right, other services that uses whoisResources are affected (some IT breaks because of that). To fix that I did next modification;

   @Override
     public boolean isWriteable(Class<?> aClass, Type type, Annotation[] annotations, MediaType mediaType) {
         return aClass == WhoisResources.class && MediaType.TEXT_PLAIN.contains(mediaType.getType());
     }
     
Now we are checking the class and the MediaType...not just the class

MiguelAHM: I will create a new PR with this change. However, it will be a very big change

MiguelAHM: https://github.com/RIPE-NCC/whois/pull/1136 -> Another approach
1) We are not able to throw the errors as we do for the 404, because we throw those errors from the Mashal strategy (StreamingMarshalTextPlainText). If we want to do this we need to move all the logic that we have in whoisRestService.lookup and whoisSeachService to that Marshal strategy and perform a big try catch to return the error in the correct format.
2) Jackson is able to transform the error in xml or json using @XmlElement and @JsonProperty but it does not support plain/text. This is why we don't have this problem with XML.

whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisResourcesPlainTextWriter.java
```diff
@@ -0,0 +1,44 @@
...
+
+    @Override
+    public void writeTo(WhoisResources whoisResources, Class<?> aClass, Type type, Annotation[] annotations, MediaType mediaType, MultivaluedMap<String, Object> multivaluedMap, OutputStream outputStream) throws IOException, WebApplicationException {
+        StringBuilder sb = new StringBuilder();
+        for (ErrorMessage errorMessage : whoisResources.getErrorMessages()) {
+            if (errorMessage.getSeverity() != null){
+                sb.append("Severity: ").append(errorMessage.getSeverity()).append('\n');
+            }
+            if (errorMessage.getText() != null){
+                sb.append("Text: ").append(errorMessage.getText()).append('\n');
```
eshryane, line 38: I suggest `errorMessage(.toString()` instead of `errorMessage.getText()` so it's more readable in plaintext

whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisServletDeployer.java
eshryane: I prefer this approach @MiguelAHM

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceTestIntegration.java
```diff
@@ -729,6 +729,35 @@ public void lookup_object_text_plain_not_found_extension() {
...
+    public void lookup_object_text_plain_bad_source_accept_header() {
+        try {
+            RestTest.target(getPort(), "whois/oez/org/MKQ-RIPE.txt")
+                    .request()
+                    .get(String.class);
+            fail();
+        } catch (BadRequestException e) {
+            final String response = e.getResponse().readEntity(String.class);
+            assertThat(response, is("Severity: Error\n" +
+                    "Text: Invalid source '%s'\n" +
```
eshryane, line 742: use `ErrorMessage.toString()` rather than `getText()`

whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisResourcesPlainTextWriter.java
MiguelAHM: .toString is not necessary, but you are right that is better to append the errorMessage directly doing .append(errorMessage)

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceTestIntegration.java
MiguelAHM: ok, now, returning the errorMessage we have oez instead of %s

whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisServletDeployer.java
MiguelAHM: ok, applied your requested changes

MiguelAHM:
TEST OK https://gitlab.ripe.net/swe-database-team/whois/-/pipelines/285510

PR #1134, opened 2022-12-01
Only as-sets with a hierarchical name can be created

eshryane:


eshryane:
Integration tests have passed

PR #1136, opened 2022-12-02
feat: another way to deal with text/plain errors

MiguelAHM:


MiguelAHM:
https://gitlab.ripe.net/swe-database-team/whois/-/pipelines/282632 Test ok

whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisRestService.java
```diff
@@ -290,12 +290,19 @@ public Response lookup(
...
-                addCommaList(QueryFlag.SELECT_TYPES, ObjectType.getByName(objectType).getName());
+        final QueryBuilder queryBuilder;
+        try {
+            queryBuilder = new QueryBuilder().
+                    addFlag(QueryFlag.EXACT).
+                    addFlag(QueryFlag.NO_GROUPING).
+                    addFlag(QueryFlag.NO_REFERENCED).
+                    addCommaList(QueryFlag.SOURCES, source).
+                    addCommaList(QueryFlag.SELECT_TYPES, ObjectType.getByName(objectType).getName());
+        } catch (final IllegalArgumentException ex){
```
eshryane, line 301: You can get an unexpected `IllegalArgumentException` anywhere, not just at this place. You are solving the `Invalid object type` case but we need to generalise it

MiguelAHM: Fixed, I am tackle all the problem in DefaultExceptionMapper, if the mediaType is Text/plain we will create a text/plain error, if not we will create a WhoisResources object (as we are doing until now)

PR #1140, opened 2022-12-06
Populate nrtm4 database

petchells:


whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/NrtmVersionInfoRepository.java
```diff
@@ -2,7 +2,6 @@
 
 import org.slf4j.Logger;
```
MiguelAHM, line 3: Change the java doc from createInitialSnapshot, you swap the same of the table from version_information to version

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmFileWriter.java
```diff
@@ -0,0 +1,25 @@
...
+
+import net.javacrumbs.shedlock.spring.annotation.SchedulerLock;
+import net.ripe.db.whois.common.scheduler.DailyScheduledTask;
+import org.springframework.scheduling.annotation.Scheduled;
+import org.springframework.stereotype.Component;
+
+
+@Component
+public class NrtmFileWriter implements DailyScheduledTask {
+
```
maggarwal13, line 11: This is not actually a filewriter, can we rename it something that suggests it's a scheduler?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/JsonSerializer.java
```diff
@@ -0,0 +1,26 @@
...
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.json.JsonMapper;
+import org.springframework.stereotype.Service;
+
+
+@Service
+public class JsonSerializer {
+
```
maggarwal13, line 11: Can we get rid of this class and use the object mapper directly as there is no logic here? Also we should get rid of  test class for this as well?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmFileRepo.java
```diff
@@ -0,0 +1,48 @@
...
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.charset.StandardCharsets;
+
+
+@Service
+public class NrtmFileRepo {
+
```
maggarwal13, line 17: THis is not a repository layer, can we rename it to may be NrtmFileStreamer or something related?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmFileRepo.java
```diff
@@ -0,0 +1,48 @@
...
+    private final String path;
+
+    NrtmFileRepo(
+        @Value("${nrtm.file.path:/tmp}") final String path
+    ) {
+        this.path = path;
+    }
+
+    boolean checkIfFileExists(final String name) {
+        return new File(path, name).exists();
```
maggarwal13, line 27: Can we put access modifier?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/NrtmFileUtil.java
```diff
@@ -0,0 +1,49 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.Random;
+import java.util.UUID;
+
+import static net.ripe.db.nrtm4.NrtmConstants.DELTA_PREFIX;
+import static net.ripe.db.nrtm4.NrtmConstants.SNAPSHOT_PREFIX;
+
+
+@Service
+public class NrtmFileUtil {
```
maggarwal13, line 17: there is no need for this class to be a service?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/NrtmFileUtil.java
```diff
@@ -0,0 +1,49 @@
...
+
+    private static final String CHARS = "0123456789abcdef";
+    private static final int RAND_STRING_LENGTH = 40;
+
+    public String fileName(final PublishableDeltaFile file) {
+        return String.format("%s.%d.%s.json", DELTA_PREFIX, file.getVersion(), randomHexString());
+    }
+
+    public String fileName(final PublishableSnapshotFile file) {
+        return String.format("%s.%d.%s.json", SNAPSHOT_PREFIX, file.getVersion(), randomHexString());
```
maggarwal13, line 27: These two function looks identical. Can we pass on the prefix from the deltaFileGenartor class (same goes for ssnapshot) ? If both snapshot and delta need to have same file format?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/NrtmFileUtil.java
```diff
@@ -0,0 +1,49 @@
...
+    public String sessionId() {
+        return UUID.randomUUID().toString();
+    }
+
+    private static String randomHexString() {
+        final Random random = new Random();
+        final StringBuilder sb = new StringBuilder(RAND_STRING_LENGTH);
+        for (int i = 0; i < RAND_STRING_LENGTH; i++) {
+            sb.append(CHARS.charAt(random.nextInt(CHARS.length())));
+        }
```
maggarwal13, line 45: can we use the Integer.toHexString(randomInt) instead of looping ? Also if there is no fix file format in RFC , do you thiunk its better to use timestamp as this file will be generator only on one server?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/NrtmFileUtil.java
```diff
@@ -0,0 +1,49 @@
...
+        return String.format("%s.%d.%s.json", DELTA_PREFIX, file.getVersion(), randomHexString());
+    }
+
+    public String fileName(final PublishableSnapshotFile file) {
+        return String.format("%s.%d.%s.json", SNAPSHOT_PREFIX, file.getVersion(), randomHexString());
+    }
+
+    public String hashString(final String payload) {
+        return Hashing.sha256()
+            .hashString(payload, StandardCharsets.UTF_8)
```
maggarwal13, line 32: We already use DigestUtils class to generate md5Hex , can we use the same library to generate sha256?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/DeltaFileGenerator.java
```diff
@@ -1,58 +1,50 @@
...
         final NrtmVersionInfoRepository nrtmVersionInfoRepository,
-        final SnapshotSynchronizer snapshotSynchronizer,
-        final SerialDao serialDao
-        ) {
-        this.deltaFileModelRepository = deltaFileModelRepository;
+        final SerialDao serialDao,
+        final DeltaFileRepository deltaFileRepository,
+        final JsonSerializer jsonSerializer,
+        final NrtmFileUtil nrtmFileUtil
+    ) {
```
maggarwal13, line 38: You do not need to inject NrtmFileUtil as a component?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/DeltaFileGenerator.java
```diff
@@ -1,58 +1,50 @@
...
         final DeltaTransformer deltaTransformer,
         final NrtmVersionInfoRepository nrtmVersionInfoRepository,
-        final SnapshotSynchronizer snapshotSynchronizer,
-        final SerialDao serialDao
-        ) {
-        this.deltaFileModelRepository = deltaFileModelRepository;
+        final SerialDao serialDao,
+        final DeltaFileRepository deltaFileRepository,
+        final JsonSerializer jsonSerializer,
+        final NrtmFileUtil nrtmFileUtil
```
maggarwal13, line 37: Can we remove the JsonSerializer ? Its a extra injection just to create a objectmapper which should be a private field here?

whois-commons/src/main/resources/nrtm_schema.sql
```diff
@@ -19,58 +19,78 @@ create table `source`
...
+create table `version`
 (
     `id`             int unsigned NOT NULL AUTO_INCREMENT,
     `source_id`      int unsigned NOT NULL,
-    `version`        int unsigned NOT NULL DEFAULT '0',
-    `session_id`     varchar(128) NOT NULL DEFAULT '',
-    `type`           varchar(128) NOT NULL DEFAULT '',
-    `last_serial_id` int          NOT NULL DEFAULT '0',
+    `version`        int unsigned NOT NULL,
+    `session_id`     varchar(128) NOT NULL,
```
maggarwal13, line 28: we are mapping a FOREIGN KEY version_id  from notification_file table to id here, what is use of version column?

whois-commons/src/main/resources/nrtm_schema.sql
```diff
@@ -19,58 +19,78 @@ create table `source`
...
-    CONSTRAINT `snapshot_file__version_information_fk` FOREIGN KEY (`version_id`) REFERENCES `version_information` (`id`)
+    UNIQUE KEY `snapshot_file__version_id__uk` (`version_id`),
+    UNIQUE KEY `snapshot_file__name__uk` (`name`),
+    CONSTRAINT `snapshot_file__version_id__fk` FOREIGN KEY (`version_id`) REFERENCES `version` (`id`)
 ) ENGINE = InnoDB
   DEFAULT CHARSET = utf8;
 
 DROP TABLE IF EXISTS `delta_file`;
 create table `delta_file`
 (
```
maggarwal13, line 68: can we rename to delta_object? if we are using snapshot_object naming convention for saving payload?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmFileRepo.java
```diff
@@ -0,0 +1,48 @@
...
+    void streamFromFile(final String name, final OutputStream out) throws IOException {
+        final File file = new File(path, name);
+        try (final FileInputStream fis = new FileInputStream(file)) {
+            fis.transferTo(out);
+        }
+    }
+
+    public void storeFile(final String name, final String payload) throws IOException {
+        final File file = new File(path, name);
+        try (final FileOutputStream fos = new FileOutputStream(file)) {
```
maggarwal13, line 39: These looks very generic File functions, we have some in FileHelper class of whois-commons. May be re-use hem or add them there? You can get rid of this extra component?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/publish/SnapshotFileStreamer.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public void writeSnapshotAsJson(
+        final PublishableSnapshotFile snapshotFile,
+        final OutputStream outputStream
+    ) throws IOException {
+        final ObjectMapper mapper = new ObjectMapper();
+        final JsonGenerator jGenerator = mapper.getFactory().createGenerator(outputStream, JsonEncoding.UTF8);
+        jGenerator.writeStartObject();
+        jGenerator.writeNumberField("nrtm_version", snapshotFile.getNrtmVersion());
+        jGenerator.writeStringField("type", NrtmDocumentType.SNAPSHOT.lowerCaseName());
```
maggarwal13, line 33: This is being used to write to fileSystem correct? I thought we already save json in database and we just need to write to file? Why we are converting to json again?

maggarwal13: Sorry I think I get it now

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotSynchronizer.java
```diff
@@ -8,13 +9,31 @@
...
+        final DeltaTransformer deltaTransformer
+    ) {
+        this.snapshotObjectRepository = snapshotObjectRepository;
+        this.deltaTransformer = deltaTransformer;
+    }
+
+    void synchronizeDeltasToSnapshot(final List<DeltaChange> deltas, final long versionId) {
+        for (final DeltaChange change : deltas) {
+            if (change.getAction() == DeltaChange.Action.ADD_MODIFY) {
+                snapshotObjectRepository.insert(
```
maggarwal13, line 26: If its a modify, don't we need op delee or modify the existing object instead of inserting new? Also it should not need any delta change just rpslObject or serial entry is suffice? Serial Entry has the operation enum

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotInitializer.java
```diff
@@ -0,0 +1,57 @@
...
+    ) {
+        this.nrtmVersionInfoRepository = nrtmVersionInfoRepository;
+        this.serialDao = serialDao;
+        this.snapshotObjectRepository = snapshotObjectRepository;
+        this.dummifierNrtm = dummifierNrtm;
+    }
+
+    NrtmVersionInfo init(final NrtmSource source) {
+        final List<SerialEntry> objects = serialDao.getSerialEntriesFromLast().stream()
+            .filter((obj) -> dummifierNrtm.isAllowed(NrtmConstants.NRTM_VERSION, obj.getRpslObject()))
```
maggarwal13, line 38: If it's a initialisation of snapshot there will be thousands of objects, do you need a streamer or may be partition?

whois-commons/src/main/java/net/ripe/db/whois/common/dao/jdbc/JdbcRpslObjectOperations.java
```diff
@@ -493,6 +511,27 @@ public SerialEntry mapRow(ResultSet rs, int rowNum) throws SQLException {
...
+            "       serials.operation, " +
+            "       serials.atlast," +
+            "       serials.object_id," +
+            "       last.object," +
+            "       last.pkey " +
+            "FROM   serials " +
+            "       JOIN last " +
+            "              ON last.object_id = serials.object_id " +
+            "WHERE serials.atlast = 1", (rs, rowNum) -> new SerialEntry(
+            rs.getInt(1),
```
maggarwal13, line 526: If it is used for iniializing snapshot file, last will also contain deleted object entries ? Do we need to put sequence_id>0 check?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotInitializer.java
```diff
@@ -0,0 +1,57 @@
...
+            .collect(Collectors.toList());
+        if (objects.isEmpty()) {
+            throw new IllegalStateException("generateSnapshot() failed because there are no objects in whois");
+        }
+        final int lastSerial = objects.get(objects.size() - 1).getSerialId();
+        final NrtmVersionInfo version = nrtmVersionInfoRepository.createInitialSnapshot(source, lastSerial);
+        for (final SerialEntry obj : objects) {
+            snapshotObjectRepository.insert(
+                version.getId(),
```
maggarwal13, line 47: We can use DeltaTransfer code and use the same logic we use in SnapshotSynchronizer?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/NrtmVersionInfoRepository.java
petchells: done

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmFileRepo.java
petchells: it's a repository in the sense that it's a directory for a certain type of file, like a git repo. i'll call it NrtmFileStore

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/JsonSerializer.java
petchells: done

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/NrtmFileUtil.java
petchells: changed to @Component.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/DeltaFileGenerator.java
petchells: done

whois-commons/src/main/resources/nrtm_schema.sql
petchells: that contains the version number -- "version" in the JSON, so i used the same name here.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/DeltaChange.java
```diff
@@ -16,46 +18,54 @@ enum Action {
...
         @JsonValue
-        public String toString() {
+        public String toLowerCaseName() {
             return name().toLowerCase();
         }
     }
 
+    @JsonIgnore
+    private final int serialId; // Only set if action is ADD_MODIFY
     private final Action action;
```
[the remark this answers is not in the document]
petchells: it's used to track an object. the story said we should be able to trace an object in a snapshot back to whois. it was discussed with Ed -- he asked for the serial to be stored.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmFileRepo.java
petchells: it's not needed so i left it out because in general it's a good idea to encapsulate functionality at package level. i'll make it public if you prefer.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/NrtmFileUtil.java
petchells: we also need a 'notification' type which will not have a file name, since at any one point in time we only have one with a fixed name (no randomness). still, it could be that we decide to map the fixed name to the latest randomly generated one in a service. i'll change it.

petchells: but also because i try to take advantage of the Java type system (polymorphism) instead of detecting types and using if...else to explicitly make decisions. it's usually less buggy that way. but still, i don't mind either way.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotSynchronizer.java
petchells: whois only has add and delete operations, same for the RFC

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmFileRepo.java
maggarwal13: No, Its cover in my later comments, I think we should use FileHelper in whois-commons. Thanks

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/DeltaChange.java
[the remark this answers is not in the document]
maggarwal13: Ok, but if it's ignored, it will not be persisted in database when you serialize it.  Or you are saving it in a different column in database?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotSynchronizer.java
maggarwal13: Yes but if we are synchronizing the existing persisted snapshot, there will already be entry of modified object in snapshot json? In this case will there be two entries in snapshot file for same object?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/NrtmFileUtil.java
maggarwal13: Sorry I meant you do not need component or service annotation for a util class

whois-commons/src/main/resources/nrtm_schema.sql
maggarwal13: Too many version names to keep track of now ;)  There is version table name with id, version column and version_id mapping to version. But if it all makes sense I do not mind :)

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmFileRepo.java
maggarwal13: methods in this class are more related to FileHelper than related to Nrtm. May be try to move or re-use existing methods in FileHelper in whois-commons?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/SnapshotUpdateWindow.java
```diff
@@ -0,0 +1,39 @@
...
+
+    SnapshotUpdateWindow(@Value("${nrtm.snapshot.window:00:00-04:00}") final String snapshotWindowString) {
+        // parse it into a representation that includes midnight -- so 10:00-06:00 starts at 10am and ends at 6am the following day
+        final Pattern PATTERN = Pattern.compile("(\\d{2}):(\\d{2})-(\\d{2}):(\\d{2})");
+        final Matcher matcher = PATTERN.matcher(snapshotWindowString);
+        if (!matcher.matches()) {
+            throw new IllegalArgumentException("Could not parse: " + snapshotWindowString + " from property nrtm.snapshot.window");
+        }
+        startTime = new Date();
+        endTime = new Date();
```
maggarwal13, line 26: where are we using snapshotWindowString? if both startTime and endTime are new Date()?

whois-commons/src/main/java/net/ripe/db/whois/common/dao/jdbc/JdbcRpslObjectOperations.java
petchells: not necessary, because when sequence_id is 0 atlast == 0 as well, so the WHERE clause will filter them out. having said that there is no db constraint on that condition, so it might be worth adding. what do you think?

maggarwal13: No you are correct its not necessary.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/DeltaChange.java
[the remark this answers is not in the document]
maggarwal13: OK now I see it, you are persisting in snapshot_object table. But if its for snapshot only can we just pass the serial_id to function synchronizeDeltasToSnapshot like we pass version_id ? Instead of putting a json ignore on deltaChange as it is not used in delta change table and not to be serialized

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotInitializer.java
petchells: yes, i'll try a stream.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotFileGenerator.java
```diff
@@ -0,0 +1,116 @@
...
+            snapshotSynchronizer.synchronizeDeltasToSnapshot(deltas, version.getId());
+        }
+        final PublishableSnapshotFile snapshotFile = new PublishableSnapshotFile(version);
+        final ByteArrayOutputStream bos = new ByteArrayOutputStream(4096);
+        try {
+            final String fileName = nrtmFileUtil.fileName(snapshotFile);
+            snapshotFileStreamer.writeSnapshotAsJson(snapshotFile, bos);
+            final String sha256hex = nrtmFileUtil.hashString(bos.toString(StandardCharsets.UTF_8));
+            snapshotFileRepository.save(
+                snapshotFile.getVersionId(),
```
maggarwal13, line 99: Why not pass the snapShotFile mean Directly to save function? Instead of creating new instance of snapShotFile in Save method and here as well?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/DeltaChange.java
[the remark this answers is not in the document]
petchells: it's saved in snapshot_object so we can see which serial_id caused an object to appear in a snapshot. we'll also be able to use that to see which version it appeared in, if we ever need that.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/SnapshotFileRepository.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public SnapshotFile save(
+        final long versionId,
+        final String name,
+        final String hash
+    ) {
+        final KeyHolder keyHolder = new GeneratedKeyHolder();
+        final long now = System.currentTimeMillis();
+        jdbcTemplate.update(connection -> {
```
maggarwal13, line 45: Can we get the timestamp same way we get for created and lastmodified attribute for whois?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/SnapshotUpdateWindow.java
```diff
@@ -0,0 +1,39 @@
+package net.ripe.db.nrtm4.util;
+
+import net.ripe.db.nrtm4.persist.SnapshotFile;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+import java.util.Date;
```
eshryane, line 7: Don't use `java.util.Date` but use `java.time` instead. There's a `net.ripe.db.whois.common.domain.Timestamp` class in whois-commons that wraps a SQL timestamp.

For more background, see e.g.
https://programminghints.com/2017/05/still-using-java-util-date-dont/

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/SnapshotUpdateWindow.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.regex.Pattern;
+
+
+@Component
+public class SnapshotUpdateWindow {
+
+    private final Date startTime;
+    private final Date endTime;
+
+    SnapshotUpdateWindow(@Value("${nrtm.snapshot.window:00:00-04:00}") final String snapshotWindowString) {
```
eshryane, line 18: Can we hard-code a schedule for generating a snapshot? e.g. run the update job every minute and use logic if enough time has passed or deltas generated to create a new snapshot.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/SnapshotFileRepository.java
maggarwal13: check TimestampAttributeGenerator class

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/NrtmFileUtil.java
eshryane: I agree you can make all methods static and don't allow `NrtmFileUtil` to be istantiated, so there is no temptation to store any state here, e.g. see `RestClientUtils`

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/DeltaFileRepository.java
```diff
@@ -0,0 +1,78 @@
...
+                return pst;
+            }, keyHolder
+        );
+        return new DeltaFile(keyHolder.getKeyAs(Long.class), versionId, name, hash, payload, now);
+    }
+
+    public DeltaFile getByVersionId(final long id) {
+        final String sql = "" +
+            "SELECT " + deltaFileFields +
+            "FROM delta_file " +
```
maggarwal13, line 65: Its way easier to read if you mention the fields here rather than in private field? Its not too many

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/NrtmFileUtil.java
eshryane: Or use built-in `MessageDigest.getInstance("SHA-256");`

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/SnapshotUpdateWindow.java
```diff
@@ -0,0 +1,39 @@
...
+        // parse it into a representation that includes midnight -- so 10:00-06:00 starts at 10am and ends at 6am the following day
+        final Pattern PATTERN = Pattern.compile("(\\d{2}):(\\d{2})-(\\d{2}):(\\d{2})");
```
eshryane, line 20: Use `java.time.format.DateTimeFormatter`

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/DeltaFile.java
```diff
@@ -0,0 +1,52 @@
+package net.ripe.db.nrtm4.persist;
```
eshryane, line 1: I suggest using the package name `dao` for consistency with the rest of Whois

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/publish/PublishableDeltaFile.java
```diff
@@ -1,5 +1,6 @@
 package net.ripe.db.nrtm4.publish;
```
eshryane, line 1: I suggest the package name `domain` for connsistency with the rest of Whois

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/DeltaFile.java
maggarwal13: and he class names too

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmFileService.java
```diff
@@ -0,0 +1,115 @@
...
+            out.write(notificationFile.getPayload().getBytes(StandardCharsets.UTF_8));
+            return;
+        }
+        syncNrtmFileToFileSystem(name);
+        nrtmFileRepo.streamFromFile(name, out);
+    }
+
+    void syncNrtmFileToFileSystem(final String name) throws IOException {
+        if (name.startsWith(DELTA_PREFIX)) {
+            if (!nrtmFileRepo.checkIfFileExists(name)) {
```
maggarwal13, line 72: Can we simplify this logic? may be create two separate methods? There are too many if else conditions

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmFileService.java
```diff
@@ -0,0 +1,115 @@
...
+                        } else {
+                            throw new FileNotFoundException("NRTM has no delta files with name: " + name);
+                        }
+                    }
+                } finally {
+                    deltaMutex.leave();
+                }
+            }
+        } else if (name.startsWith(SNAPSHOT_PREFIX)) {
+            if (!nrtmFileRepo.checkIfFileExists(name)) {
```
maggarwal13, line 89: You must have got the filename from the nrtm database when you are trying to write it? If yes , I would rather rely on documnetType enum rather than filename.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotFileGenerator.java
```diff
@@ -0,0 +1,116 @@
...
+            }
+        }
+        // TODO: apply pending deltas
+        //       * if snapshot version > 1...
+        //           - find list of deltas since the last snapshot
+        //           - process them with SnapshotSynchronizer to bring snapshot_objects up to date
+        if (version.getVersion() > 1) {
+            final NrtmVersionInfo lastSnapshot = nrtmVersionInfoRepository.findLastSnapshotVersion(source);
+            final List<SerialEntry> whoisChanges = serialDao.getSerialEntriesBetween(lastSnapshot.getLastSerialId(), version.getLastSerialId());
+            final List<DeltaChange> deltas = deltaTransformer.toDeltaChange(whoisChanges);
```
maggarwal13, line 84: This looks like job of snapshotSynchronizer? As Initiliazor is doing all the work reletd to serialDoa ? This way we do noit need to inject SErialDoa bean here ?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotFileGenerator.java
```diff
@@ -0,0 +1,116 @@
...
+public class SnapshotFileGenerator {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(SnapshotFileGenerator.class);
+
+    private final DeltaTransformer deltaTransformer;
+    private final NrtmVersionInfoRepository nrtmVersionInfoRepository;
+    private final SnapshotInitializer snapshotInitializer;
+    private final SnapshotFileStreamer snapshotFileStreamer;
+    private final SnapshotFileRepository snapshotFileRepository;
+    private final SnapshotSynchronizer snapshotSynchronizer;
```
maggarwal13, line 35: Why this class needs SnapshotFileRepository? If both Initializer and synchronizer is already injecting the SnapshotFileRepository and saving to database ?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotFileGenerator.java
```diff
@@ -0,0 +1,116 @@
...
+            if (deltas.size() < 1) {
+                LOGGER.error("Changes found but no delta has been published since serialID {}. Now at {}",
+                    lastSnapshot.getLastSerialId(), version.getLastSerialId());
+                return Optional.empty();
+            }
+            snapshotSynchronizer.synchronizeDeltasToSnapshot(deltas, version.getId());
+        }
+        final PublishableSnapshotFile snapshotFile = new PublishableSnapshotFile(version);
+        final ByteArrayOutputStream bos = new ByteArrayOutputStream(4096);
+        try {
```
maggarwal13, line 94: Should creating outstream be part of FileStreamer?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/NrtmFileUtil.java
petchells: done

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/NrtmFileUtil.java
petchells: the RFC says the string should not be predictable before publication, so i took that to mean that it should be completely random rather than generated from timestamp. changed the implementation to be similar to what you suggest (but using two longs, since Integer produces only 8 chars)

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/NrtmFileUtil.java
petchells: i made it injectable it so that the randomness can be overridden for use in unit tests -- otherwise it can't be tested. do you think it's better to use an interface and use a different implementation for tests? i'm thinking of the session id, which is put into JSON files and hashed.

maggarwal13: We should avoid coding thinking about writing unit tests and I am not sure how making it injectable avoid randomness.  Lets talk separately may be I can help

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmFileRepo.java
petchells: this idea of this class is to encapsulate all the functionality we need for NRTM specific files. FileHelper looks like a generic util class for dealing with Zip files. i suggest it's better to keep this as a dedicated NRTM file manager component rather than putting NRTM-specific functionality into FileHelper, because then it would have two responsibilities -- one common and one NRTM-specific. i think it's better to divide responsibilities than stack them up, especially since the functionality is not 'common'. at least, it shouldn't be common -- the NRTM module should be the only module reading and writing NRTM files.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotSynchronizer.java
petchells: yes, you're right. as i mentioned in mattermost there's a missing piece here.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/SnapshotUpdateWindow.java
petchells: nowhere yet, i was just sketching out some ideas in code to ensure the snapshots are generated at the right time. i'll remove this stuff.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmFileRepo.java
maggarwal13: But I think functions in this class has nothing to do with NRTM like getting FileInputStream or even checking if file exists or not, these can easily be converted to static method and used in any where where we need to do something with file stream and we can avoid injecting unncessary beans. Its compnent is just using nrtm path that you can pass from NrtmFileService, thats the NRtm specific component and should reside there.

maggarwal13: May be create a FileStreamerHelper class ? if you do not like commons, lets put it in nrtm module :)

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/NrtmFileUtil.java
maggarwal13: Thanks for explaining, in that case we should not use timestamp.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/NrtmFileUtil.java
```diff
@@ -0,0 +1,42 @@
...
+import static net.ripe.db.nrtm4.NrtmConstants.NOTIFICATION_PREFIX;
+import static net.ripe.db.nrtm4.NrtmConstants.SNAPSHOT_PREFIX;
+
+
+@Component
+public class NrtmFileUtil {
+
+    private static final String CHARS = "0123456789abcdef";
+    private static final int RAND_STRING_LENGTH = 40;
+
```
maggarwal13, line 21: These are no longer used?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotFileGenerator.java
petchells: so the snapshot gets published as soon as it can and this thread/transaction can close sooner (which is a good thing). it might take a while to write the file so the transaction will be kept open also (and potentially rolled back). also, other servers won't pick up this change until the read job executes -- so doing it this way means all servers create the file at pretty much the same time. none of this is really necessary, though. it just "feels" a bit neater. thoughts?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/SnapshotUpdateWindow.java
petchells: that's pretty much the idea behind this code. it's not finished yet -- i'll remove this until we have a nice way to do it.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/publish/SnapshotFileStreamer.java
```diff
@@ -0,0 +1,50 @@
...
+        jGenerator.writeStringField("source", snapshotFile.getSource().name());
+        jGenerator.writeStringField("session_id", snapshotFile.getSessionID());
+        jGenerator.writeNumberField("version", snapshotFile.getVersion());
+        jGenerator.writeArrayFieldStart("objects");
+        snapshotObjectRepository.streamSnapshots(snapshotFile.getSource()).forEach((payload) -> {
+            try {
```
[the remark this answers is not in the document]
petchells: done

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotFileGenerator.java
petchells: yes, done

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotFileGenerator.java
petchells: it now streams directly to disk -- ByteArrayOutputStream hogs Java memory so it's generating OOMErrors in the JVM.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/NrtmFileUtil.java
petchells: done

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotFileGenerator.java
maggarwal13: ByteArrayOutputStream uses the heap memory and loads all data there . can you try writing directly to the file using BufferedOutputStream

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotFileGenerator.java
petchells: update: changed this now to generate the file. reason: it takes fewer resources to calculate the hash of a file rather than doing it in memory. it takes a bit longer, though. we'll need to see how long when it's working on a prod, or prod-like machine.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/publish/PublishableDeltaFile.java
petchells: done

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmFileService.java
petchells: this code is for dealing with incoming client requests -- it checks if the file exists on disk before going to the database.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotFileGenerator.java
petchells: yes, was already in an earlier commit. still seeing OOMErrors

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotFileGenerator.java
petchells: hang on -- i don't think that was your question, was it? i thought earlier that i'd need the id for the update notification, so i'd need a new SnapshotFile with the id. doing that a different way now so changed the code -- 'save' returns void.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/SnapshotFileRepository.java
petchells: switched to use dataTimeProvider -- same as the last and history tables in whois.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/DeltaFileRepository.java
petchells: done

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/SnapshotUpdateWindow.java
petchells: class was removed -- it's just a sketch that i happened to leave in the commit.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/persist/DeltaFile.java
petchells: package renamed. not sure what you mean about the class names -- do you mean rename *Repository to *Dao? "Repository" is already used in Whois, and these classes are annotated "Repository", while DAO is architecturally on layer above the Repository. for my money "Repository" is both consistent and correct, so please confirm if you want them to be renamed to Dao.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/SnapshotUpdateWindow.java
petchells: removed this 'sketch' from the branch

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmFileStore.java
```diff
@@ -0,0 +1,52 @@
...
+import java.nio.charset.StandardCharsets;
+
+
+@Service
+public class NrtmFileStore {
+
+    private final String path;
+
+    NrtmFileStore(
+        @Value("${nrtm.file.path:/tmp}") final String path
```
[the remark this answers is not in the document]
petchells: this is the path to the dir where the files are stored

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotSynchronizer.java
```diff
@@ -1,20 +1,74 @@
...
 
 
 @Service
 public class SnapshotSynchronizer {
 
-    void synchronizeDeltasToSnapshot(final List<DeltaChange> changes) {
-        // TODO: apply changes to the snapshot
-        // if 'operation' is '1' (add) then add to snapshot table
-        // if 'operation' is '2' (del) then remove from snapshot table
+    private static final Logger LOGGER = LoggerFactory.getLogger(SnapshotSynchronizer.class);
```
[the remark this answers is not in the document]
petchells: ah, true. i've commented it out.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/DeltaFileGenerator.java
```diff
@@ -0,0 +1,88 @@
...
+        if (whoisChanges.size() < 1) {
+            LOGGER.info("No Whois changes found -- delta file generation skipped");
+            return Optional.empty();
+        }
+        final List<DeltaChange> deltas = deltaTransformer.toDeltaChange(whoisChanges);
+        if (deltas.size() < 1) {
+            LOGGER.info("Whois changes found but all were filtered");
+            return Optional.empty();
+        }
+        final int lastSerialId = whoisChanges.get(whoisChanges.size() - 1).getSerialId();
```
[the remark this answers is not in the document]
petchells: this is to get the serial ID of the last item in the list of whoisChanges. they're ordered by serial ID so the last one holds the last serial ID processed when the transaction started. is that what you mean, or have i missed sth?

whois-commons/src/main/java/net/ripe/db/whois/common/dao/jdbc/JdbcRpslObjectOperations.java
```diff
@@ -493,96 +514,174 @@ public SerialEntry mapRow(ResultSet rs, int rowNum) throws SQLException {
         }, serialId);
     }
 
+//    private static List<SerialEntry> getSerialEntryWithBlobsFromLastForNrtm4(final JdbcTemplate jdbcTemplate) {
```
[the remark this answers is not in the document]
petchells: yes it's removed now.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/DeltaFileGenerator.java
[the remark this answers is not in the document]
petchells: about the @Component on the util class -- there's a test to see that the sha256 hash is calculated correctly. however, the payload contains a session ID which is random. so i use @Component so i can mock it and return a fixed string for use in the payload, then the hash is predictable (and testable). mahesh also had an idea about how to do this. still to be resolved.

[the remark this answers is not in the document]
MiguelAHM: ok, yes it is clear now. Thank you! 
About the @Component is a bit odd, I understand what is your problem with the tests but I think that there is a way to mock static methods. I did it in the past with ImageIO

https://www.baeldung.com/mockito-mock-static-methods

So you don't need to use @Component to mock an static method. Please have a look on this.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmFileStore.java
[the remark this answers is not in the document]
MiguelAHM: yes, but I don't understand why we need to store data in a file. Maybe we talk about it before going on holidays but I don't remember now :(
Could you explain me why we need to use a file if we already have the data in the database?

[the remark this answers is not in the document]
petchells: the idea is to reduce the workload on the server in case we get a lot of requests. if we can serve a file from the file system it's quicker than going to the database for it, so we use the file system as a kind of cache. when it's working i think it's a good idea to do some load tests to see how well this works. it won't help us in the case of a DoS attack, though. maybe we need a different countermeasure to help with that (e.g. put all valid sessionID/filenames in memory? just an idea).

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/DeltaFileGenerator.java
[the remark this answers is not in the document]
petchells: ok done. though i find it odd that we have Spring but we're still making more complicated test setups so we can use static methods.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/NrtmFileUtil.java
petchells: my mistake -- it's actually an integration test. changed now to be completely static.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotInitializer.java
petchells: that's what i was doing initially, but now we're using SerialDao methods which join on both 'last' and 'history' tables -- but for the initial snapshot we only need the 'last' table. this is to make the query more efficient.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmFileService.java
petchells: split into two services and uses the 'return early' principle to avoid nesting

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/DeltaFileGenerator.java
petchells: done -- though since we're using Spring i'd normally prefer to make it injectable like pretty much every other functional class.

maggarwal13: We should not make util classes as injectable. Just because anything could be wired as a Spring bean doesn't mean everything should be wired as a Spring bean. Thanks for making this change. This keeps the code simple

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmFileStore.java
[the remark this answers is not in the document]
MiguelAHM: Yes, maybe we need to think a little bit more about this. Maybe use @Cacheable and store data in memory and build the output using the cached data? there is a way to create a programatic cache (spring boot at least) I have done some of them, very useful for Collections maybe we should keep an eye on this.

PR #1141, opened 2022-12-06
Db 4325 fix typos

MiguelAHM:


whois-query/src/test/java/net/ripe/db/whois/query/integration/TemplateTestIntegration.java
```diff
@@ -292,7 +292,7 @@ public void verbose_description() {
                 "     letter or a digit.  The following words are reserved by\n" +
                 "     RPSL, and they can not be used as names:\n" +
                 "     \n" +
-                "      any as-any rs-any peeras and or not atomic from to at\n" +
+                "      any as-any rs-any peerAS and or not atomic from to at\n" +
```
eshryane, line 295: peerAS is a keyword and needs to be lowercase. See https://www.rfc-editor.org/rfc/rfc2622#section-2

PR #1147, opened 2023-01-02
feat: as-set more explainatory

MiguelAHM:


whois-update/src/main/java/net/ripe/db/whois/update/domain/UpdateMessages.java
```diff
@@ -191,7 +191,8 @@ public static Message cantRemoveOrgAttribute() {
     }
 
     public static Message cantCreateShortFormatAsName() {
-        return new Message(Type.ERROR, "Cannot create AS-SET object with a short format name.");
+        return new Message(Type.ERROR, "Cannot create AS-SET object with a short format name. Only hierarchical " +
+                "AS-SET creation is allowed. Refer to the documentation for more information");
```
eshryane, line 195: I suggest not writing `Refer to the documentation...` because the error message should be self-explanatory. Maybe explain as `Only hierarchical AS-SET creation is allowed, i.e. at least one ASN must be referenced`

whois-rpsl/src/main/java/net/ripe/db/whois/common/rpsl/AttributeSyntax.java
```diff
@@ -69,7 +69,9 @@ public interface AttributeSyntax extends Documented {
             "\n" +
             "An as-set name can also be hierarchical.  A hierarchical set\n" +
             "name is a sequence of set names and AS numbers separated by\n" +
-            "colons \":\".  At least one component of such a name must be\n" +
+            "colons \":\". The first element of the name must be an ASN\n" +
+            "followed by colons and finishing with as-set (example: AS15562:AS-SNIJDERS)." +
```
eshryane, line 73: We should always use generic examples (i.e. don't reference a specific person or company). How about AS3333:AS-TEST as an example instead?

whois-rpsl/src/main/java/net/ripe/db/whois/common/rpsl/AttributeSyntax.java
```diff
@@ -69,7 +69,9 @@ public interface AttributeSyntax extends Documented {
             "\n" +
             "An as-set name can also be hierarchical.  A hierarchical set\n" +
             "name is a sequence of set names and AS numbers separated by\n" +
-            "colons \":\".  At least one component of such a name must be\n" +
+            "colons \":\". The first element of the name must be an ASN\n" +
```
eshryane, line 72: Replace "ASN" with "AS number" and "finishing" with "ending".

Re-phrase as, `The first element of the name must be an AS number followed a colon and ending with a name`

MiguelAHM: AS number followed **by** a colon?

eshryane: Ah yes thanks! Correct

eshryane (approved):

PR #1156, opened 2023-01-06
feat: nrtm4 log class set up

MiguelAHM:


whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmLog.java
```diff
@@ -0,0 +1,29 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import java.net.InetAddress;
+
+@Component
+public class NrtmLog {
+
```
maggarwal13, line 11: can we rename it to Nrtm4Log to avoid confusion with NrtmLog?

tools/log4j2.xml
```diff
@@ -69,6 +69,19 @@
...
 
+        <RollingFile name="NRTM4LOG"
+                     append="true"
+                     filePattern="var/log/nrtm4/nrtmlog.%d{yyyyMMdd}"
+                     ignoreExceptions="false">
+            <ThresholdFilter level="debug"/>
+            <PatternLayout>
+                <Pattern>%d{yyyyMMdd HH:mm:ss} %m%n</Pattern>
+            </PatternLayout>
+            <Policies>
```
maggarwal13, line 80: I think one of these log4j2.cml is not being used. Can we remove one that is nit used for tests? We manually copy the log4 xml from db-operational to deployment directory

maggarwal13 (approved):
Make sure you edit the log4j2.xml for salt in db-operational and delete the non salt version

PR #1157, opened 2023-01-10
Db 4381 nrtm4 log changes

MiguelAHM:


whois-commons/src/main/java/net/ripe/db/whois/common/dao/SerialDao.java
```diff
@@ -4,6 +4,7 @@
 import net.ripe.db.whois.common.domain.serials.SerialRange;
 
 import java.util.List;
+import java.util.stream.Stream;
```
maggarwal13, line 7: Can you raise a merge request from Etch branch so we see only your changes to review

MiguelAHM: oh sorry, I was talking with Etch about this changes I will create a pull request just with those changes. I only changed the parameter of nrtm4log and add a dependency

MiguelAHM: https://github.com/RIPE-NCC/whois/pull/1158/files
But the artefact of the dependency is <artifactId>jetty-rewrite</artifactId> instead of jetty-servlets

PR #1163, opened 2023-01-17
Replace junit4 imports (dependency still exists thanks to testcontainers)

eshryane:


whois-api/src/test/java/net/ripe/db/whois/api/elasticsearch/ElasticIndexServiceIntegrationTest.java
```diff
@@ -26,63 +27,63 @@ public class ElasticIndexServiceIntegrationTest extends AbstractElasticSearchInt
...
         whoisDocCount = elasticIndexService.getWhoisDocCount();
         // no document in index after deleting
-        assertEquals(whoisDocCount, 0);
+        assertThat(whoisDocCount, is(0));
     }
 
     @Test
     public void isEnabledWhenIndicesExist() {
-        assertTrue(elasticIndexService.isEnabled());
+        assertThat(elasticIndexService.isEnabled(), is(true));
```
MiguelAHM, line 62: I read in Mahesh PR that we are using assertTrue in almost everywhere. Maybe we should keep using it and if in the end we decided to change it we should apply the change in one task because of traceability. What do you think? Because if we change it in different task if going to be difficult to track it

eshryane: On the one hand, it's better to follow whatever precedent is already set in the code. On the other hand, we're using Hamcrest for writing tests, and `assertTrue` is inconsistent with this style (where you first write `assertThat` then the test, e.g. see the tutorial: https://hamcrest.org/JavaHamcrest/tutorial
I checked the code and I see a mix between `assertTrue` and `assertThat`, then if we have a choice I prefer `assertThat` to be consistent with Hamcrest and I think it's more readable. Let's discuss separately.

eshryane: After discussion, agreed to move `assertTrue` refactor to a separate change

MiguelAHM (approved):

eshryane:
Integration tests have passed

PR #1167, opened 2023-01-19
Replace JUnit 5 assertions with Hamcrest assertThat(...) for consistency

eshryane:


eshryane:
Integration tests have passed

PR #1168, opened 2023-01-26
Elasticsearch hhtps

maggarwal13:


whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticSearchInstance.java
```diff
@@ -0,0 +1,54 @@
...
+
+import javax.annotation.Nullable;
+import java.util.List;
+
+@Profile({WhoisProfile.DEPLOYED})
+@Primary
+@Component
+public class ElasticSearchInstance implements ElasticRestHighlevelClient {
+    private final RestHighLevelClient client;
+
```
MiguelAHM, line 27: Should we add the logger here as well?
private static final Logger LOGGER = getLogger(ElasticIndexService.class);

maggarwal13: Do we need to log something here?

whois-api/src/test/java/net/ripe/db/whois/api/elasticsearch/ElasticSearchTestInstance.java
```diff
@@ -0,0 +1,47 @@
...
+import javax.annotation.Nullable;
+import java.util.List;
+
+import static org.slf4j.LoggerFactory.getLogger;
+
+@Profile({WhoisProfile.TEST})
+@Component
+public class ElasticSearchTestInstance implements ElasticRestHighlevelClient {
+
+    private static final Logger LOGGER = getLogger(ElasticIndexService.class);
```
eshryane, line 24: Change LOGGER class to `ElasticSearchTestInstance`

whois-api/src/test/java/net/ripe/db/whois/api/elasticsearch/ElasticSearchTestInstance.java
```diff
@@ -0,0 +1,47 @@
...
+    private final RestHighLevelClient client;
+
+    @Autowired
+    public ElasticSearchTestInstance(@Value("#{'${elastic.host:}'.split(',')}") final List<String> elasticHosts) {
+        this.client = getEsClient(elasticHosts);
+    }
+
+    @Nullable
+    private RestHighLevelClient getEsClient(final List<String> elasticHosts) {
```
eshryane, line 33: Can you name the class according to what it does? It looks like this is the "non auth" instance vs. `ElasticSearchInstance` which requires authentication.

whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticSearchInstance.java
maggarwal13: Actually I should catch all exception here and log as warning.

whois-api/src/test/java/net/ripe/db/whois/api/elasticsearch/ElasticSearchTestInstance.java
maggarwal13: May be ElasticSearchTestNonAuthInstance? It is used only in test and it is in test package so I named it as ElasticSearchTestInstance

maggarwal13: Renamed to ElasticSearchNonAuthInstance

eshryane: Thanks @maggarwal13 I think this name is clearer as it describes what it does (but still stays in `src/test/`)

eshryane (approved):

PR #1169, opened 2023-01-27
feat: change source of as-set when parent belong to another irr

MiguelAHM:


whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceTestIntegration.java
```diff
@@ -1835,6 +1835,67 @@ public void create_succeeds() {
         assertThat(whoisResources.getTermsAndConditions().getHref(), is(WhoisResources.TERMS_AND_CONDITIONS));
     }
 
+
+
+    @Test
+    public void create_asSet_non_auth_aut_num_succeeds() {
+        final RpslObject TEST_AS_SET = RpslObject.parse("""
+                as-set:     AS3333:AS-TEST
```
maggarwal13, line 1843: Can we add test case that for flat hierarchy it does not change the source?

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetAttributeTransformer.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public AsSetAttributeTransformer(final RpslObjectDao rpslObjectDao){
+        this.rpslObjectDao = rpslObjectDao;
+    }
+
+    @Override
+    public RpslObject transform(RpslObject rpslObject, Update update, UpdateContext updateContext, Action action) {
+        String asSetKey = rpslObject.getKey().toString();
+        if (asSetKey.contains(":")){
+            String autnumKey = asSetKey.substring(0, asSetKey.indexOf(":"));
```
maggarwal13, line 35: other object  key can contains :, this will be executed for intenum too?

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceTestIntegration.java
```diff
@@ -1835,6 +1835,67 @@ public void create_succeeds() {
         assertThat(whoisResources.getTermsAndConditions().getHref(), is(WhoisResources.TERMS_AND_CONDITIONS));
     }
 
+
```
eshryane, line 1838: Add a test case for a child AS-SET (e.g. AS3333:AS-TEST:AS12654)

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetAttributeTransformer.java
```diff
@@ -0,0 +1,49 @@
...
+    @Value("${whois.nonauth.source}")
+    private String nonAuthRipeSource;
+
+    public AsSetAttributeTransformer(final RpslObjectDao rpslObjectDao){
+        this.rpslObjectDao = rpslObjectDao;
+    }
+
+    @Override
+    public RpslObject transform(RpslObject rpslObject, Update update, UpdateContext updateContext, Action action) {
+        String asSetKey = rpslObject.getKey().toString();
```
maggarwal13, line 33: I would say return if action is delete or object type is not as-set

eshryane, line 33: Don't perform this action for non-AS-SET objects, also not on delete. For example,
```
        if (DELETE == action ||
            (rpslObject.getType() != ObjectType.AS_SET)) {
            return rpslObject;
        }
```

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetAttributeTransformer.java
```diff
@@ -0,0 +1,49 @@
...
+        if (asSetKey.contains(":")){
+            String autnumKey = asSetKey.substring(0, asSetKey.indexOf(":"));
+            RpslObject autnumObject = rpslObjectDao.getByKeyOrNull(ObjectType.AUT_NUM, autnumKey);
+
+            if (autnumObject != null && nonAuthRipeSource.equals(Objects.requireNonNull(autnumObject.getValueOrNullForAttribute(AttributeType.SOURCE)).toString())){
+                final Map<RpslAttribute, RpslAttribute> replace = Maps.newHashMap();
```
maggarwal13, line 39: why you need Objects.requireNonNull? Source is mandatory so use getValueForAttribute. Also use CISTring equals method it takes care of string as well.

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetAttributeTransformer.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Override
+    public RpslObject transform(RpslObject rpslObject, Update update, UpdateContext updateContext, Action action) {
+        String asSetKey = rpslObject.getKey().toString();
+        if (asSetKey.contains(":")){
+            String autnumKey = asSetKey.substring(0, asSetKey.indexOf(":"));
+            RpslObject autnumObject = rpslObjectDao.getByKeyOrNull(ObjectType.AUT_NUM, autnumKey);
+
+            if (autnumObject != null && nonAuthRipeSource.equals(Objects.requireNonNull(autnumObject.getValueOrNullForAttribute(AttributeType.SOURCE)).toString())){
```
eshryane, line 38: Replace `Objects.requireNonNull(...)` with `autnumObject.getValueForAttribute()` which will throw `IllegalArgumentException` if `SOURCE` is not present exactly once.

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetAttributeTransformer.java
maggarwal13: Also if user has specified source as non-auth there is no need to perform a check? Return from here only?

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetAttributeTransformer.java
```diff
@@ -0,0 +1,49 @@
...
+                final Map<RpslAttribute, RpslAttribute> replace = Maps.newHashMap();
+                replace.put(rpslObject.findAttribute(AttributeType.SOURCE), new RpslAttribute(AttributeType.SOURCE, nonAuthRipeSource));
+
+                RpslObjectBuilder builder = new RpslObjectBuilder(rpslObject);
+                builder.replaceAttributes(replace);
+                return builder.get();
```
maggarwal13, line 44: also use replaceAttribute not replaceAttributes? You can also simplify it with new RpslObjectBuilder(rpslObject).replaceAttribute(<oldAttribute>,<New Attribute>).get();

whois-update/src/main/java/net/ripe/db/whois/update/handler/validator/common/SourceValidator.java
```diff
@@ -28,7 +29,7 @@ public class SourceValidator implements BusinessRuleValidator {
     private static final ImmutableList<Action> ACTIONS = ImmutableList.of(Action.CREATE, Action.MODIFY);
     private static final ImmutableList<ObjectType> TYPES = ImmutableList.copyOf(ObjectType.values());
 
-    private static final Set<ObjectType> NON_AUTH_OBJECT_TYPES = ImmutableSet.of(AUT_NUM, ROUTE, ROUTE6);
+    private static final Set<ObjectType> NON_AUTH_OBJECT_TYPES = ImmutableSet.of(AS_SET, AUT_NUM, ROUTE, ROUTE6);
```
eshryane, line 32: This will allow the user to set `source: RIPE-NONAUTH` on *any* AS-SET object, we must prevent that unless the AUT-NUM is `RIPE-NONAUTH` by adding validation to the Transformer class

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetAttributeTransformer.java
```diff
@@ -0,0 +1,49 @@
...
+            }
+        }
+        return rpslObject;
+    }
```
maggarwal13, line 48: Add a UpdateMessages as warning ?? Check MntRoutesAttributeTransformer as an example

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetAttributeTransformer.java
eshryane, line 38: Add `else` case to require source: matches main source otherwise

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceTestIntegration.java
eshryane, line 1838: Add a test case for UPDATE flat (non-hierarchical) AS-SET with `source: TEST-NONAUTH`

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceTestIntegration.java
```diff
@@ -1835,6 +1835,67 @@ public void create_succeeds() {
...
+        final WhoisResources whoisResources = RestTest.target(getPort(), "whois/test/as-set?password=test")
+                .request()
+                .post(Entity.entity(map(TEST_AS_SET), MediaType.APPLICATION_XML), WhoisResources.class);
+
+        assertThat(whoisResources.getLink().getHref(), is(String.format("http://localhost:%s/test/as-set", getPort())));
+        assertThat(whoisResources.getErrorMessages(), is(empty()));
+        final WhoisObject object = whoisResources.getWhoisObjects().get(0);
+
+        assertThat(object.getAttributes(), containsInAnyOrder(
```
eshryane, line 1868: Use `contains()` rather than `containsInAnyOrder` since the order is predictable

eshryane: Can also just test the value of the "source" attribute as it's clearer (we don't care about the other values)

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceTestIntegration.java
```diff
@@ -1835,6 +1835,67 @@ public void create_succeeds() {
...
+                new Attribute("as-set", "AS3333:AS-TEST"),
+                new Attribute("tech-c", "TP1-TEST", null, "person", Link.create("http://rest-test.db.ripe.net/test/person/TP1-TEST"), null),
+                new Attribute("admin-c", "TP1-TEST", null, "person", Link.create("http://rest-test.db.ripe.net/test/person/TP1-TEST"), null),
+                new Attribute("mnt-by", "OWNER-MNT", null, "mntner", Link.create("http://rest-test.db.ripe.net/test/mntner/OWNER-MNT"), null),
+                new Attribute("created", "2001-02-04T17:00:00Z"),
+                new Attribute("last-modified", "2001-02-04T17:00:00Z"),
+                new Attribute("source", "TEST-NONAUTH")));
+
+        assertThat(whoisResources.getTermsAndConditions().getHref(), is(WhoisResources.TERMS_AND_CONDITIONS));
```
eshryane, line 1877: No need for this test here

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceTestIntegration.java
```diff
@@ -1835,6 +1835,67 @@ public void create_succeeds() {
...
+    }
+
+
+    @Test
+    public void create_asSet_non_existing_aut_num_bad_request() {
```
eshryane, line 1882: This may happen if the AUT-NUM is deleted after the AS-SET is created, I don't think this should fail.

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetAttributeTransformer.java
```diff
@@ -0,0 +1,49 @@
...
+import net.ripe.db.whois.update.domain.Update;
+import net.ripe.db.whois.update.domain.UpdateContext;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+import java.util.Map;
+import java.util.Objects;
+
+@Component
+public class AsSetAttributeTransformer implements Transformer{
```
eshryane, line 20: Rename to `AsSetSourceAttributeTransformer` as it's the SOURCE attribute not the AS-SET attribute being updated

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetAttributeTransformer.java
```diff
@@ -0,0 +1,49 @@
...
+
+    private final RpslObjectDao rpslObjectDao;
+
+    @Value("${whois.nonauth.source}")
+    private String nonAuthRipeSource;
```
eshryane, line 25: Rename to `nonAuthSource` (may not be RIPE source)

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetAttributeTransformer.java
MiguelAHM: yes I will check the type and the actions

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetAttributeTransformer.java
MiguelAHM: yes I know, IntelliJ suggestion, but we can get rid of it yes (I don't like to see highlighted lines of code by IntelliJ)

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetAttributeTransformer.java
eshryane: I mean, there's no need to use `Objects.requireNonNull()` it's taken care of by RpslObject.

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetAttributeTransformer.java
maggarwal13: Yes use CIString equals instead of converting CIString first to string. Intellij won't complain then

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetAttributeTransformer.java
MiguelAHM: mmm...we are creating that warning for deprecated fields, here I think that we don't need a warning. We need the error message when the user is trying to create/update as-set with nonauth source when the aut-num's source is ripe

maggarwal13: Not sure, if we change anything in object we should send back warning or inform user . check UpdateMessaages class. For example: valueChangedDueToLatin1Conversion.   May be discuss with @eshryane  about this

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceTestIntegration.java
MiguelAHM: So if we remove the aut-num and we try to update the related as-set, we shouldn't receive an error? I think that we should delete the as-set as well when we delete the related aut-num. A cascade delete I mean.

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetAttributeTransformer.java
MiguelAHM: ah yes, you are right

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceTestIntegration.java
eshryane: Yes you won't receive an error. We don't automatically remove the as-set because they are used for different purposes., although maybe we should put a constraint not to allow deleting the aut-num object if there's a reference from an as-set.

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetAttributeTransformer.java
eshryane: Yes we should include a warning on the AS-SET object (add to UpdateMessages.java): 
```"The "source:" attribute value has been updated from X to Y to match the referenced AUT-NUM Z```
I'll update the requirements.

File: whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceTestIntegration.java (hunk 1 above)
Comment (MiguelAHM) [reply whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceTestIntegration.java:1875]: if you try to create a flat as-set you will receive an error and if you try to update and you are updating the source you will receive an error as well

File: whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetAttributeTransformer.java (hunk 4 above)
Comment (MiguelAHM) [reply whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetAttributeTransformer.java]: we need to check that the aut-num is non-auth as well. If both, aut-num and as-set has the same source I do a return now.

File: whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceTestIntegration.java (hunk 3 above)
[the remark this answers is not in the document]
Comment (MiguelAHM) [reply whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceTestIntegration.java:1847]: done, we receive an error in that case

File: whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetSourceAttributeTransformer.java [hunk 15]
```diff
@@ -0,0 +1,69 @@
...
+        if (Action.DELETE.equals(action) || !ObjectType.AS_SET.equals(rpslObject.getType())) {
+            return rpslObject;
+        }
+        final String asSetKey = rpslObject.getKey().toString();
+
+        boolean flatAsSet = !asSetKey.contains(":");
+        final CIString asSetSource = rpslObject.getValueForAttribute(AttributeType.SOURCE);
+
+        if (flatAsSet){
+            updateContext.addMessage(update, UpdateMessages.flatModelNotAllowSourceModifications(asSetSource.toString()));
```
eshryane: This message is added for *all* create or update to a flat AS-SET, regardless of source?

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetAttributeTransformer.java
MiguelAHM: if the user doesn't specify the source we will throw an illegal exception from the builder.

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetSourceAttributeTransformer.java
```diff
@@ -0,0 +1,69 @@
...
+        if (autnumObject == null){
+            return rpslObject;
+        }
+
+        final CIString autnumSource = autnumObject.getValueForAttribute(AttributeType.SOURCE);
+        if (asSetSource.equals(autnumSource)){
+            return rpslObject;
+        }
+
+        if (nonAuthSource.contentEquals(autnumSource)){
```
eshryane, line 60: Reverse the comparison, i.e. `autnumSource.equals(nonAuthSource)` as CIString equals() is case-insensitive

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetSourceAttributeTransformer.java
```diff
@@ -0,0 +1,69 @@
...
+            updateContext.addMessage(update, UpdateMessages.sourceChanged(asSetSource, autnumSource, autnumKey));
+
+            return new RpslObjectBuilder(rpslObject).replaceAttribute(rpslObject.findAttribute(AttributeType.SOURCE),
+                    new RpslAttribute(AttributeType.SOURCE, nonAuthSource)).get();
+        } else if (nonAuthSource.contentEquals(asSetSource)) {
```
eshryane, line 65: Reverrse the comparison, i.e. `asSetSource.equals(nonAuthSource)`

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/Transformer.java
```diff
@@ -7,7 +7,7 @@
 
 public interface Transformer {
 
-    RpslObject transform(final RpslObject rpslObject,
+    RpslObject transform(final RpslObject updatedObject,
```
eshryane, line 10: The *original* object is passed to `transform()` and the *updated* object is returned from it

MiguelAHM: mmmm...here we have the updated object, not the original I mean we don't have here the error from the database. We have the body of the request

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceTestIntegration.java
eshryane: We need a case-insensitive comparison

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/Transformer.java
eshryane: No I mean the original (as in the unmodified by *this* method) object is passed to the method, it's updated if necessary and returned. Using `updatedObject` before it's updated is misleading.

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetSourceAttributeTransformer.java
MiguelAHM: I put this in that order because I usually put the CONST.equals(variable) to avoid null pointers. On the other hand, IntelliJ doesn't like to compare CIString with an String because of different types and appears in yellow the equals :S 
Anyway if you are ok with it I don't mind. I will update

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetSourceAttributeTransformer.java
MiguelAHM: I am reviewing this, I will try to add a warning and change to the original source if the user try to update the source. I don't have the original object so I don't know the original message source...and I don't like to pass another parameter to the transformer

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetSourceAttributeTransformer.java
```diff
@@ -0,0 +1,71 @@
...
+            return rpslObject;
+        }
+        final String asSetKey = rpslObject.getKey().toString();
+
+        boolean flatAsSet = !asSetKey.contains(":");
+        final CIString asSetSource = rpslObject.getValueForAttribute(AttributeType.SOURCE);
+
+        if (flatAsSet){
+            if (asSetSource != null) {
+                updateContext.addMessage(update, UpdateMessages.flatModelNotAllowSourceModifications(asSetSource.toString()));
```
maggarwal13, line 43: asSetSource can neve be null it is mandatory. Also you are using getValueForAttribute which will throw exception

whois-update/src/main/java/net/ripe/db/whois/update/domain/UpdateMessages.java
```diff
@@ -656,6 +656,18 @@ public static Message multipleUserMntBy(final Collection<CIString> userMntners)
         return new Message(Type.ERROR, "Multiple user-'mnt-by:' are not allowed, found are: '%s'", Joiner.on(", ").join(userMntners));
     }
 
+    public static Message notValidSource() {
+        return new Message(Type.ERROR, "Can not set NONAUTH source when aut-num is not NONAUTH");
+    }
+
+    public static Message flatModelNotAllowSourceModifications(final String source){
+        return new Message(Type.ERROR, "Can not set '%s' source when as-set in not hierarchical", source);
```
maggarwal13, line 664: Do we need a check for flat as-set?

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetSourceAttributeTransformer.java
MiguelAHM: I am changing this

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetSourceAttributeTransformer.java
maggarwal13: CIString checks for the null value as well

whois-update/src/main/java/net/ripe/db/whois/update/domain/UpdateMessages.java
MiguelAHM: I am changing the error message, now it will be a warning and it will explain that the update request will change because we are not allow to update the source, we will take the database's source (this warning can only be thrown for update)

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetSourceAttributeTransformer.java
```diff
@@ -0,0 +1,71 @@
...
+    @Override
+    public RpslObject transform(final RpslObject rpslObject, final Update update,
+                                final UpdateContext updateContext,
+                                final Action action) {
+        if (Action.DELETE.equals(action) || !ObjectType.AS_SET.equals(rpslObject.getType())) {
+            return rpslObject;
+        }
+        final String asSetKey = rpslObject.getKey().toString();
+
+        boolean flatAsSet = !asSetKey.contains(":");
```
maggarwal13, line 38: use final?

whois-update/src/main/java/net/ripe/db/whois/update/domain/UpdateMessages.java
maggarwal13: But the check or warning is only for hierarchal for the flat there should not be any change in source?

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetAttributeTransformer.java
eshryane: You needed an `else` to check the main (RIPE) source, you were only checking the nonauth source. The code has changed since this comment, I'll resolve it.

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetSourceAttributeTransformer.java
eshryane: Yes const first is a good practice, but you need a case-insensitive comparison. Either make the `nonAuthSource` field a CIString and put that first, or add a null-check on the source to be sure.

MiguelAHM: ok, nice to know!

whois-update/src/main/java/net/ripe/db/whois/update/handler/transform/AsSetSourceAttributeTransformer.java
MiguelAHM: now this message will be triggered when the user is updating an flat as-set and the source is different. We will notify the user that the update request has changed to match the database source

whois-update/src/main/java/net/ripe/db/whois/update/domain/UpdateMessages.java
MiguelAHM: in the jira task: 
Do not update the source of flat AS-SET names (i.e. without an AUT-NUM reference).

We can not update the source, so if the user updates the source we need to notify about the change I guess

eshryane: There's no AUT-NUM reference in flat AS-SET names (by definition) so no need to add any warning, just do nothing.

whois-update/src/main/java/net/ripe/db/whois/update/domain/UpdateMessages.java
```diff
@@ -656,6 +656,19 @@ public static Message multipleUserMntBy(final Collection<CIString> userMntners)
         return new Message(Type.ERROR, "Multiple user-'mnt-by:' are not allowed, found are: '%s'", Joiner.on(", ").join(userMntners));
     }
 
+    public static Message notValidSource() {
+        return new Message(Type.ERROR, "Can not set NONAUTH source when aut-num is not NONAUTH");
+    }
```
maggarwal13, line 661: If you add an error instead of changing something, it should be in validator not transfer. Also check SingleUpdateHandler, we look for update error message only in businessValidator rule, transfer should not throw validation error , it should just only transform object.

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceTestIntegration.java
MiguelAHM: yes, I allows you to update the object. Test:
update_as_set_when_aut_num_deleted_succeeds()

Update an hierarchical as-set without autumn

whois-update/src/main/java/net/ripe/db/whois/update/domain/UpdateMessages.java
MiguelAHM: ok, I will move this

MiguelAHM: removed this, as we talk in by zoom

whois-update/src/main/java/net/ripe/db/whois/update/generator/SourceGenerator.java
```diff
@@ -21,30 +28,69 @@ public class SourceGenerator extends AttributeGenerator {
...
+            return updatedObject;
+        }
+        final String asSetKey = updatedObject.getKey().toString();
+
+        final boolean flatAsSet = !asSetKey.contains(":");
+        final CIString asSetSource = updatedObject.getValueForAttribute(AttributeType.SOURCE);
+
+        if (flatAsSet){
+            final RpslObject asSetObject = rpslObjectDao.getByKeyOrNull(ObjectType.AS_SET, asSetKey);
+            if (asSetObject == null){
```
maggarwal13, line 66: Original object is saved in the updateContext, you do not need to hit the database?

whois-update/src/main/java/net/ripe/db/whois/update/generator/SourceGenerator.java
```diff
@@ -21,30 +28,69 @@ public class SourceGenerator extends AttributeGenerator {
...
                 return updatedObject;
+            }
+            final CIString databaseAsSetSource = asSetObject.getValueForAttribute(AttributeType.SOURCE);
+            if(!asSetSource.equals(databaseAsSetSource)) {
+                updateContext.addMessage(update, ValidationMessages.suppliedAttributeReplacedWithGeneratedValue(AttributeType.SOURCE));
```
maggarwal13, line 71: if it is a flatAsSet and source are equal it should return from here? Right now it will continue executing ?

whois-update/src/main/java/net/ripe/db/whois/update/generator/SourceGenerator.java
MiguelAHM: mmmm...I only see the object of the update request I don't see the one from database

whois-update/src/main/java/net/ripe/db/whois/update/generator/SourceGenerator.java
MiguelAHM: added a return inside the if (flatAsSet)

whois-update/src/main/java/net/ripe/db/whois/update/generator/SourceGenerator.java
```diff
@@ -21,30 +28,69 @@ public class SourceGenerator extends AttributeGenerator {
...
     public RpslObject generateAttributes(RpslObject originalObject, RpslObject updatedObject, Update update, UpdateContext updateContext) {
-        switch (updatedObject.getType()) {
-            case AUT_NUM:
-            case ROUTE:
-            case ROUTE6:
-                return generateSource(updatedObject, update, updateContext);
-            default:
+        return switch (updatedObject.getType()) {
+            case AS_SET -> generateAsSetSource(updatedObject, update, updateContext);
+            case AUT_NUM, ROUTE, ROUTE6 -> generateSource(updatedObject, update, updateContext);
```
maggarwal13, line 50: we need a break statement?

MiguelAHM: no because we are doing a return previously

whois-update/src/main/java/net/ripe/db/whois/update/generator/SourceGenerator.java
MiguelAHM: done

maggarwal13 (approved):

whois-update/src/main/java/net/ripe/db/whois/update/generator/SourceGenerator.java
```diff
@@ -21,30 +28,73 @@ public class SourceGenerator extends AttributeGenerator {
...
+        if (update.getOperation() == Operation.DELETE) {
+            return updatedObject;
+        }
+        final String asSetKey = updatedObject.getKey().toString();
+
+        final boolean flatAsSet = !asSetKey.contains(":");
+        final CIString asSetSource = updatedObject.getValueForAttribute(AttributeType.SOURCE);
+
+        if (flatAsSet){
+            final RpslObject asSetObject = updateContext.getPreparedUpdate(update).getReferenceObject();
```
eshryane, line 65: Use `originalObject` instead, it's passed in the parent method

maggarwal13: Yes attributeGenrator does have original Object , sorry I suggested to get it from updateContext instead of database. I missed this

MiguelAHM: ohh totally blind! thank you!

eshryane (approved):

MiguelAHM:
https://gitlab.ripe.net/swe-database-team/whois/-/pipelines/304606 -> Tests

PR #1171, opened 2023-01-31
Separate sources

petchells:


whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotFileGenerator.java
```diff
@@ -45,51 +48,56 @@ public SnapshotFileGenerator(
         this.nrtmFileStore = nrtmFileStore;
     }
 
-    public Optional<PublishableSnapshotFile> createSnapshot(final NrtmSource source) throws IOException {
+    public List<PublishableSnapshotFile> createSnapshots() throws IOException {
         // Get last version from database.
-        final Optional<NrtmVersionInfo> lastVersion = nrtmVersionInfoRepository.findLastVersion(source);
-        NrtmVersionInfo version;
+        final Optional<NrtmVersionInfo> lastVersion = nrtmVersionInfoRepository.findLastVersion();
+        final List<NrtmVersionInfo> versions = new ArrayList<>();
```
MiguelAHM, line 54: Could we change the name of these variable, to specify that we are storing a version per source? maybe sourcesVersion instead of version? or something like this?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotFileGenerator.java
```diff
@@ -45,51 +48,56 @@ public SnapshotFileGenerator(
...
-            LOGGER.info("Generated snapshot file in {}", stopwatch);
-            return Optional.of(snapshotFile);
-        } catch (final IOException e) {
-            LOGGER.error("Exception thrown when calculating hash of snapshot file", e);
-            throw e;
+        final List<PublishableSnapshotFile> snapshotFiles = new ArrayList<>();
+        for (final NrtmVersionInfo version : versions) {
+            LOGGER.info("{} at version: {}", version.getSource().source(), version);
+            if (version.getVersion() > 1) {
+                LOGGER.debug("Sync Whois changes to snapshot here (not implemented)");
```
MiguelAHM, line 78: I don't understand this step, could you explain me what do we have to do here?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotFileGenerator.java
```diff
@@ -45,51 +48,56 @@ public SnapshotFileGenerator(
...
+            try (final OutputStream out = nrtmFileStore.getFileOutputStream(snapshotFile.getSessionID(), fileName)) {
+                snapshotFileSerializer.writeSnapshotAsJson(snapshotFile, out);
+            }
+            LOGGER.info("Wrote JSON for {} in {}", snapshotFile.getSource().name(), stopwatch);
+            try {
+                stopwatch = Stopwatch.createStarted();
+                final String sha256hex = DigestUtils.sha256Hex(nrtmFileStore.getFileInputStream(snapshotFile.getSessionID(), fileName));
+                snapshotFile.setFileName(fileName);
+                snapshotFile.setHash(sha256hex);
+                snapshotFileRepository.insert(snapshotFile);
```
MiguelAHM, line 92: Here you are inserting one file per source in the end. However, I see that you are not using the source to identify that file, are we going to use the hash to identify the file of that source, right? there is going to be a match between the source and the hash

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotObjectSynchronizer.java
```diff
@@ -31,31 +34,33 @@ public class SnapshotObjectSynchronizer {
...
+        this.sourceRepository = sourceRepository;
         this.whoisObjectRepository = whoisObjectRepository;
     }
 
-    NrtmVersionInfo initializeSnapshotObjects(final NrtmSource source) {
-        final String method = "initializeSnapshotObjects";
+    InitialSnapshotState initializeSnapshotObjects() {
+        LOGGER.info("initializeSnapshotObjects entered");
+        final Map<CIString, NrtmSourceModel> sourceMap = new HashMap<>();
+        for (final NrtmSource source : NrtmSourceHolder.getAllSources()) {
```
MiguelAHM, line 56: instead of doing the loop on this list twice (NrtmSourceHolder.getAllSource())
you could do it once in the SnapshotFileGenerator and pass the Map<CIString, NrtmSourceModel> as a parameter. I think that here you could avoid one loop

petchells: they're used in slightly different ways, so i'd still need to loop or refactor the code to make both uses compatible. that's ok except the code needs changes for the next story so i'd prefer to keep it like this for now. there's no I/O or memory implications so it will be really fast anyway.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotFileGenerator.java
petchells: done

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotFileGenerator.java
MiguelAHM: ok, thank you!

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/NrtmVersionInfoRepository.java
```diff
@@ -57,7 +58,7 @@ public NrtmVersionInfoRepository(
     public Optional<NrtmVersionInfo> findLastVersion(final NrtmSource source) {
         try {
```
maggarwal13, line 59: If we are building initial snapshot files  for all sources at once, just checking if this table is empty or not is suffice , there is no need to join with sources table and get last record (last record could be any source)? If we want to do this way , then we need to check if lastVersion is  available for all sources and build for the sources for which there is no initial version available but we cannot do that so best to check just for empty or not?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/domain/NrtmSourceModel.java
```diff
@@ -0,0 +1,5 @@
+package net.ripe.db.nrtm4.domain;
+
+public record NrtmSourceModel(long id, NrtmSource source) {
```
maggarwal13, line 3: What do you think of this? 
Use NrtmSource itself and add id there, we can put jsonIgnore for it. Source will always have id so I am not sure why we have three domain classes for sources.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotObjectSynchronizer.java
```diff
@@ -31,31 +34,33 @@ public class SnapshotObjectSynchronizer {
...
+            final NrtmSourceModel sourceModel = sourceRepository.createSource(source);
+            sourceMap.put(CIString.ciString(source.name()), sourceModel);
```
maggarwal13, line 58: NrtmSourceHolder is a component. I think it should deal with all logic related to source. What you think if that class has a function createAllSources? And NrtmSourceHolder should also have function to give us id based on source instead of creating a local map here? This will simplify this class to deal with only snapshotObject logic not nrtmSource? Also may be rename NrtmSourceHolder to NrtmSourceProcessor or something similar?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotObjectSynchronizer.java
```diff
@@ -68,14 +73,19 @@ NrtmVersionInfo initializeSnapshotObjects(final NrtmSource source) {
                             continue;
                         }
                         final RpslObject dummyRpsl = dummifierNrtm.dummify(NRTM_VERSION, rpslObject);
-                        batch.add(new SnapshotObject(0, version.getId(), object.objectId(), object.sequenceId(), dummyRpsl));
+                        final NrtmSourceModel source = sourceMap.get(dummyRpsl.getValueForAttribute(AttributeType.SOURCE));
+                        if (source == null) {
+                            final String msg = "RPSL object declares an unknown source attribute";
+                            LOGGER.error(msg + " " + dummyRpsl.getValueForAttribute(AttributeType.SOURCE));
```
maggarwal13, line 79: This whole logic should be in nrtmSource component, you should just pass the source name and it gives you id.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/NrtmVersionInfoRepository.java
```diff
@@ -113,12 +129,12 @@ INSERT INTO version_info (source_id, version, session_id, type, last_serial_id,
                 return pst;
             }, keyHolder
         );
-        return new NrtmVersionInfo(keyHolder.getKeyAs(Long.class), source, version, sessionID, type, lastSerialId, now);
+        return new NrtmVersionInfo(keyHolder.getKeyAs(Long.class), sourceID, source, version, sessionID, type, lastSerialId, now);
     }
 
     public NrtmVersionInfo findLastSnapshotVersion(final NrtmSource source) {
         final String sql = """
```
maggarwal13, line 136: If we are creating snapshot file for all sources at once, do we really need this method?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/SnapshotFileRepository.java
```diff
@@ -34,20 +35,16 @@ public SnapshotFileRepository(
...
         jdbcTemplate.update(connection -> {
                 final String sql = "" +
                     "INSERT INTO snapshot_file (version_id, name, hash) " +
                     "VALUES (?, ?, ?)";
                 final PreparedStatement pst = connection.prepareStatement(sql, Statement.RETURN_GENERATED_KEYS);
-                pst.setLong(1, versionId);
-                pst.setString(2, name);
-                pst.setString(3, hash);
+                pst.setLong(1, snapshotFile.getVersionId());
+                pst.setString(2, snapshotFile.getFileName());
```
maggarwal13, line 46: directly call update(String sql, @Nullable Object... args) method, it creates PreparedStatement automatically. Keeps it simpler?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/domain/NrtmSourceHolder.java
```diff
@@ -3,6 +3,7 @@
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
```
maggarwal13, line 8: This is component should not be in domain package?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/SourceRepository.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public NrtmSourceModel createSource(final NrtmSource source) {
+        final KeyHolder keyHolder = new GeneratedKeyHolder();
+        jdbcTemplate.update(connection -> {
+            final String sql = """
+                INSERT INTO source (name)
+                VALUES (?)
+                """;
+            final PreparedStatement pst = connection.prepareStatement(sql, Statement.RETURN_GENERATED_KEYS);
+            pst.setString(1, source.name());
```
maggarwal13, line 35: Do you really need to pass on Statement.RETURN_GENERATED_KEYS? It should be automatically set to 1?

whois-nrtm4/src/test/java/net/ripe/db/nrtm4/SnapshotFileGeneratorIntegrationTest.java
```diff
@@ -45,9 +45,9 @@ public void initial_snapshot_file_is_generated_and_written_to_disk() throws IOEx
         loadScripts(whoisTemplate, "nrtm_sample_sm.sql");
         final String sessionID;
```
maggarwal13, line 46: Can we add a testcase or test dataq for mutiple sources not just TEST source?

maggarwal13: Can we try to use DatabaseHelper in setup class, if we are not dealing with lot of objects? Its easier to navigate and no need to add manual entries in serials and history tables

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotFileGenerator.java
```diff
@@ -45,51 +48,56 @@ public SnapshotFileGenerator(
...
-                LOGGER.info("Not generating snapshot file yet");
-                return Optional.empty();
-            } else {
-                LOGGER.info("Not generating snapshot file since no deltas have been published since v{} with serialID {}",
-                    version.getVersion(), version.getLastSerialId());
-                return Optional.empty();
+            final InitialSnapshotState state = snapshotObjectSynchronizer.initializeSnapshotObjects();
+            for (final NrtmSource source : NrtmSourceHolder.getAllSources()) {
+                final NrtmVersionInfo version = nrtmVersionInfoRepository.createInitialVersion(source, state.serialId());
+                nrtmFileStore.createNrtmSessionDirectory(version.getSessionID());
```
maggarwal13, line 60: is this needs o be inside if condition?  We always have to create a version even if its not initial snapshot? So this logic remains same as for publishable file generation?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/NrtmVersionInfoRepository.java
```diff
@@ -130,7 +146,7 @@ public NrtmVersionInfo findLastSnapshotVersion(final NrtmSource source) {
 
     public Optional<NrtmVersionInfo> findById(final long versionId) {
         final String sql = """
-            SELECT v.id, src.name, v.version, v.session_id, v.type, v.last_serial_id, v.created
+            SELECT v.id, src.id, src.name, v.version, v.session_id, v.type, v.last_serial_id, v.created
             FROM version_info v
```
maggarwal13, line 150: This will throw an emptyResultException optional.nullable has no effect?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/NrtmVersionInfoRepository.java
petchells: actually this method is already unused except by tests. removed

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotObjectSynchronizer.java
petchells: agree. i'll try it out.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/SnapshotFileRepository.java
maggarwal13: But this function is returning void

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/NrtmVersionInfoRepository.java
maggarwal13: But this being uses in createSnapshot to check if we need to initialize the snapshot or not

whois-nrtm4/src/test/java/net/ripe/db/nrtm4/SnapshotFileGeneratorIntegrationTest.java
maggarwal13: It's ok if you prefer sql dump and if we have been using it other places. Its just for me its easier to read other way but its personal choice :)

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/NrtmVersionInfoRepository.java
maggarwal13: In that case catch the empty result exception and throw IllegalStateException or InternalException?My point is optional.empty will never be a case here as the method will throw exception, so no point in using optional actually

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/domain/NrtmSourceModel.java
maggarwal13: But properties are same for every machine as we copy from single location. Also we use jmx for creating initial source in tables so it can never happen that it will be created by two instances at same time? and we should have unique constraint on name in table schema?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotFileGenerator.java
petchells: Latest commit uses only NrtmSourceModel to represent a source and all functions are in SourceRepository. Don't worry about how the code in this section is now -- it will change a lot in the next story when we won't have a snapshot_object table (so initializeSnapshotObjects will be removed)

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotObjectSynchronizer.java
petchells: still in place for now, but the next story doesn't use the db for snapshot objects so this class will be removed.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/NrtmVersionInfoRepository.java
petchells: separated now into checking for sources and checking for versions with a source. no constraints or check on whether number of versions == number of sources, maybe should be added?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/NrtmVersionInfoRepository.java
petchells: done

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotObjectSynchronizer.java
petchells: done

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/NrtmVersionInfoRepository.java
MiguelAHM: Here I feel more comfortable with try catch...you never know what could happen in the future

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmFileProcessor.java
```diff
@@ -21,38 +21,38 @@ public class NrtmFileProcessor {
...
     public void updateNrtmFilesAndPublishNotification() throws IOException {
-        LOGGER.info("runWrite() called");
-        final NrtmSource source = nrtmSourceHolder.getSource();
-        final Optional<SnapshotFile> lastSnapshot = snapshotFileGenerator.getLastSnapshot(source);
-        Optional<PublishableSnapshotFile> publishableSnapshotFile = Optional.empty();
+        LOGGER.info("updateNrtmFilesAndPublishNotification() called");
+        final List<NrtmSourceModel> sourceList = sourceRepository.getAllSources();
+        if (sourceList.isEmpty()) {
+            sourceRepository.createSources();
+        }
```
maggarwal13, line 44: Do we really want to do this? If source is empty it means no initial snapshot exists and it should start creating in jmx enabled. So why create separately?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmFileProcessor.java
```diff
@@ -21,38 +21,38 @@ public class NrtmFileProcessor {
...
+        final Optional<SnapshotFile> lastSnapshot = snapshotFileGenerator.getLastSnapshot(sourceRepository.getSource().orElseThrow());
         if (lastSnapshot.isEmpty()) {
```
maggarwal13, line 46: Is this checking for whois.source only? What will happen if non auth source does not exists?  So either you check both sources to decide if initialisation is needed or not or you check if just the tables is empty or not? But I would prefer you just check for empty table as your initialisation logic expects an empty database . correct?

MiguelAHM: as far as I understand we create everything for all the sources at same time. So if we check one of the source and that source contains the snapshot it means that the rest of them contains the snapshot as well. This check should work correctly and if it doesn't is because be have a bug. Therefore I really don't mind to check if the source has the snapshot or if this table is empty or not (regardless the source)

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmFileProcessor.java
petchells: done

MiguelAHM (approved):

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/SourceRepository.java
```diff
@@ -0,0 +1,74 @@
...
+            INSERT INTO source (name)
+            VALUES (?)
+            """;
+        for (final String source : sources) {
+            jdbcTemplate.update(sql, source);
+        }
+    }
+
+    public Optional<NrtmSourceModel> getSource() {
+        final String sql = """
```
maggarwal13, line 47: May be rename to getMainSource or getWhoisSource as by default we are sending whois source?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmFileWriterScheduledTask.java
```diff
@@ -15,15 +16,23 @@ public class NrtmFileWriterScheduledTask implements DailyScheduledTask {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(NrtmFileWriterScheduledTask.class);
     private final NrtmFileProcessor nrtmFileProcessor;
+    private final Boolean disableNrtm4Feature;
 
-    NrtmFileWriterScheduledTask(final NrtmFileProcessor nrtmFileProcessor) {
+    NrtmFileWriterScheduledTask(
+        final NrtmFileProcessor nrtmFileProcessor,
+        @Value("${nrtm.disable.nrtm4.feature:false}") final Boolean disableNrtm4Feature
```
maggarwal13, line 23: rename to nrtm4.enabled?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/domain/NrtmSourceModel.java
```diff
@@ -0,0 +1,26 @@
...
+        this.id = id;
+        this.source = source;
+    }
+
+    public long getId() {
+        return id;
+    }
+
+    @JsonValue
+    public CIString getSource() {
```
maggarwal13, line 22: are we serializing source anywhere?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/SourceRepository.java
petchells: only used by tests so better to remove it from production code. renamed for now.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/domain/NrtmSourceModel.java
petchells: yes, explicitly in SnapshotFileSerializer, so the annotation isn't strictly necessary. i'll remove it.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmFileWriterScheduledTask.java
petchells: ok

maggarwal13 (approved):

PR #1174, opened 2023-02-02
feat: upgrade the version to 1.106

MiguelAHM:


whois-commons/src/main/resources/whois.version.properties
```diff
@@ -3,7 +3,7 @@
 #
 
 # public release version (tagged in repo)
-application.version=1.105-SNAPSHOT
+application.version=1.106-SNAPSHOT
```
eshryane, line 6: You need to set the version to "1.106" then merge it, and create a tagged release. The release is not a "SNAPSHOT".

Afterwards, create another PR and set the version to "1.107-SNAPSHOT" (i.e. for developing snapshots for the *next* release). Then merge it and we'll use that for development.

eshryane: Also in this PR it's a good idea to update changes.txt, include the major changes since 1.105 (that a user would care about).

MiguelAHM: ok, I got confused because now we have 1.1.0.5-SNAPSHOT we should have 1.1.0.6-SNAPSHOT, right? 
About the changes.txt totally agree I just noticed once I created the PR I think that I added a comment in Mattermost

eshryane (approved):

PR #1175, opened 2023-02-02
feat: set snapshot version

MiguelAHM:


whois-commons/src/main/resources/whois.version.properties
```diff
@@ -3,7 +3,7 @@
 #
 
 # public release version (tagged in repo)
-application.version=1.106
+application.version=1.107-SNAPSHOT
```
maggarwal13, line 7: Comment say only for public release version. If we keep on updating snapshot version like this every commit in master needs to have a new snapshot version or your branch . I do not see use of this. Also we never deploy snapshot in any other environment apart from dev and predev

MiguelAHM: yes, seems that we forgot to set the 1.106-SNAPSHOT previously:
we had 1.0.5-SNAPSHOT
then I made the change to 1.106 (because it is the release that we want to deploy)
so then I have to change to 1.0.6-SNAPSHOT

Seems that when we deployed the 1.0.5 we didn't update this file adding the 1.0.6-SNAPSHOT so not we have 1.0.5-SNAPSHOT and 1.0.7-SNAPSHOT

MiguelAHM: so do you mean that we should not do this?

maggarwal13: No I understood  now. If tag is ready we can merge this

MiguelAHM: exactly, we can not merge because there is a problem with the NRTM PR. We are checking it, so I need to create another tag after fixing...anyways I will merge after creating that new tag

PR #1179, opened 2023-02-09
feat: get rid of default maintainer history tables

MiguelAHM:


whois-commons/src/main/resources/patch/default-maintainer-history.sql
```diff
@@ -0,0 +1,2 @@
+DROP TABLE IF EXISTS `default_maintainer_history`;
```
eshryane, line 1: I think we need to version patch files against the current or next version of Whois, also update the `version` table to match - check the other patch files for examples

whois-commons/src/main/resources/patch/whois-1.106.sql
```diff
@@ -0,0 +1,7 @@
+-- delete default maintainer history tables. Not longer in use.
+
+DROP TABLE IF EXISTS `default_maintainer_history`;
+DROP TABLE IF EXISTS `default_maintainer_sync_history`;
+
```
maggarwal13, line 5: It should be Internals patch file?

maggarwal13: check internals-1.102.sql as an example

MiguelAHM: what is the difference here? how could I know that this is an internals?

maggarwal13: By which schema table exists in.  If tables exists in whois schema it should be in whois patch file, if table exists in internals schema it should be in internals schema

MiguelAHM: ok, makes sense

maggarwal13 (approved):

PR #1183, opened 2023-02-15
Use new default mntner table

maggarwal13:


MiguelAHM:
Should we include this change in the current release or should we create a bug fix and create a new release?

maggarwal13:
> Should we include this change in the current release or should we create a bug fix and create a new release?

maggarwal13:
> > Should we include this change in the current release or should we create a bug fix and create a new release?

I am not sure, we need to discuss with Ed but in any case this need to be merged to master

MiguelAHM:
is it ok to call to the internal database from whois, instead of using whois-internal? (I though that we were accessing internal schema's databases just from whois-internal

maggarwal13:
> is it ok to call to the internal database from whois, instead of using whois-internal? (I though that we were accessing internal schema's databases just from whois-internal

Not true we do lot of places in whois codebase. Search for usage if internalsTemplate. I would not put a dependency of whois-internal service being available for whois to run properly.

MiguelAHM (approved):

PR #1190, opened 2023-02-27
Gzip snapshots to db

petchells:


whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotFileGenerator.java
```diff
@@ -3,92 +3,148 @@
...
+                LOGGER.info("NRTM writer thread interrupted", e);
+                Thread.currentThread().interrupt();
             }
         }
-        return snapshotFiles;
+        LOGGER.info("NRTM generation complete {}", stopwatchRoot);
+        return publishedFiles;
     }
 
     public Optional<SnapshotFile> getLastSnapshot(final NrtmSourceModel source) {
```
MiguelAHM, line 150: We are not using it

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotFileGenerator.java
```diff
@@ -3,92 +3,148 @@
...
         this.snapshotFileRepository = snapshotFileRepository;
-        this.snapshotObjectSynchronizer = snapshotObjectSynchronizer;
+        this.snapshotFileSerializer = snapshotFileSerializer;
         this.sourceRepository = sourceRepository;
-        this.nrtmFileStore = nrtmFileStore;
+        this.whoisObjectRepository = whoisObjectRepository;
     }
 
-    public List<PublishableSnapshotFile> createSnapshots() throws IOException {
+    public Set<PublishableNrtmFile> createSnapshots() {
```
MiguelAHM, line 70: this side of code is a bit complicated to understand.
Could we split into modules? which a descriptive name per each module?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/RpslObjectEnqueuer.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.atomic.AtomicInteger;
+
+
+@Service
+public class RpslObjectEnqueuer {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(RpslObjectEnqueuer.class);
+    private static final int BATCH_SIZE = 100;
+    public static final RpslObjectData POISON_PILL = new RpslObjectData(0, 0, null);
```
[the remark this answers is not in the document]
MiguelAHM: We are using this to do a kind of separation between object when we are writing the json, right?
Maybe OBJECT_DELIMITER or something like this?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/domain/ObjectData.java
```diff
@@ -0,0 +1,5 @@
+package net.ripe.db.nrtm4.domain;
+
+public record ObjectData(int objectId, int sequenceId) {
```
MiguelAHM, line 3: LastWhoisObjects?
ObjectData is too ambiguous I think

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/NrtmFileUtil.java
```diff
@@ -17,6 +17,11 @@ public static String newFileName(final PublishableNrtmDocument file) {
         return String.format("%s.%d.%s.json", prefix, file.getVersion(), randomHexString());
```
MiguelAHM, line 17: newFileName visibility could be private, right? we are just using it from newGzFileName

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/domain/ObjectData.java
petchells: they're not always 'last', that's the thing. it's possible to do updates while this is running so an object might actually be deleted between getting List<ObjectData> and then getting the RPSL (which will then be in 'history'). "WhoisObjectMetaData"? "WhoisObjectState"? if so then what should RpslObjectData become?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/NrtmFileService.java
```diff
@@ -1,54 +1,126 @@
...
+        try (final FileInputStream fis = NrtmFileUtil.getFileInputStream(path, sessionId, name)) {
+            fis.transferTo(out);
+        }
     }
 
     void syncNrtmFileToFileSystem(final String sessionId, final String name) throws IOException {
         if (!name.startsWith(NrtmDocumentType.DELTA.getFileNamePrefix()) &&
-                !name.startsWith(NrtmDocumentType.SNAPSHOT.getFileNamePrefix())) {
+            !name.startsWith(NrtmDocumentType.SNAPSHOT.getFileNamePrefix())) {
             throw new IllegalArgumentException("Not an NRTM file name: " + name);
```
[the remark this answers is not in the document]
MiguelAHM: Agree that it is cleaner on this way

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/domain/ObjectData.java
MiguelAHM: Maybe merge both? In the end is the same just with a new attribute 
Then we can call it WhoisObjectData?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotFileGenerator.java
```diff
@@ -3,92 +3,148 @@
...
+                    Thread.currentThread().interrupt();
+                    throw new RuntimeException(e);
+                }
+            });
+            queueReader.start();
+            queueReaders.add(queueReader);
+        }
+        final Thread queueWriter = new Thread(() -> {
+            try {
+                final int total = state.objectData().size();
```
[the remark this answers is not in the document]
MiguelAHM: I saw that you already changed this. Thank you!

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotFileGenerator.java
MiguelAHM: Thank you! It is easier to read now

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotFileGenerator.java
```diff
@@ -3,92 +3,148 @@
...
+        this.whoisObjectRepository = whoisObjectRepository;
     }
 
-    public List<PublishableSnapshotFile> createSnapshots() throws IOException {
+    public Set<PublishableNrtmFile> createSnapshots() {
+        final Stopwatch stopwatchRoot = Stopwatch.createStarted();
         // Get last version from database.
         final List<NrtmVersionInfo> sourceVersions = nrtmVersionInfoRepository.findLastVersionPerSource();
+        final SnapshotState state = whoisObjectRepository.getSnapshotState();
+        LOGGER.info("NRTM found {} objects in {}", state.objectData().size(), stopwatchRoot);
```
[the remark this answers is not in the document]
MiguelAHM: ah ok! Leave it if you think that it is easier to track the logs, in the end this is the final purpose.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotFileGenerator.java
```diff
@@ -3,96 +3,134 @@
...
-        return snapshotFiles;
+        LOGGER.info("Generation complete {}", stopwatchRoot);
+        return publishedFiles;
     }
 
     public Optional<SnapshotFile> getLastSnapshot(final NrtmSourceModel source) {
         return snapshotFileRepository.getLastSnapshot(source);
     }
 
+    private record RunnableFileGenerator(NrtmFileService nrtmFileService, SnapshotFileRepository snapshotFileRepository,
```
MiguelAHM, line 108: these parameters should be final, right?

petchells: no -- a record is immutable by default so 'final' is implied.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/domain/ObjectData.java
petchells: at the moment ObjectData and RpslObjectData are defined as records. records are immutable, so you can't do 'setRpslObject(...)'; all values must be passed in the constructor. i could use a normal POJO and leave the rpslObject field null until it gets filled but that's hacky -- better to keep the clean separation in definition and in the code, for my money. I'll make them WhoisObjectData and RpslObjectData for now.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/RpslObjectEnqueuer.java
```diff
@@ -0,0 +1,127 @@
...
+    ) {
+        return new RpslObjectQueueRunner(
+            whoisObjectRepository,
+            snapshotState,
+            queueMap,
+            CIString.ciString(whoisSource)
+        );
+    }
+
+    private static class RpslObjectQueueRunner implements Runnable {
```
MiguelAHM, line 56: This could be a record, right?

petchells: yes -- i just noticed the notice from intellij. but do you think it's a good idea to have a 'runnable record' class? there are no getters for these constructor vars and i don't think there should be any, so it's not /really/ a record, nor it used like one (i.e. it implements functionality, not a data class like a pojo).

MiguelAHM: so we only will use record for pojo.
ok!

maggarwal13 (approved):

PR #1193, opened 2023-03-02
Handle expected servlet exceptions in RDAP exception mapper

eshryane:
Don't log an ERROR for expected exceptions from the servlet container (Jetty).

We *do* still need to log any unexpected errors, such as the unhandled "multiple mnt-by found" (which I created a separate test case for, needs to be fixed still).


whois-commons/src/test/resources/whois.properties
```diff
@@ -142,3 +142,5 @@ rdap.inaccuracy_notice.description=If you see inaccuracies in the results, pleas
 rdap.inaccuracy_notice.linkhref=https://www.ripe.net/contact-form?topic=ripe_dbm&show_form=true
 
 shutdown.pause.sec=0
+
+instance.name=localhost
```
maggarwal13, line 146: why we need this? I remeber we did because of aws migration but we do not need this anymore

maggarwal13 (approved):

whois-commons/src/test/resources/whois.properties
eshryane: Maybe we'll need it again once we start to run Whois in a container? The reason I defined it is because of an IllegalStateException is thrown in `whois-rpsl/src/main/java/net/ripe/db/whois/common/domain/Hosts.java` if neither the host or the instance is defined, which is the case on my laptop, and all the exceptions are distracting if running multiple tests.

MiguelAHM (approved):

PR #1196, opened 2023-03-14
Update notification

petchells:


whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotGenerationWindow.java
```diff
@@ -0,0 +1,53 @@
...
+import org.springframework.stereotype.Service;
+
+import java.time.LocalTime;
+import java.time.ZoneOffset;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+
+@Service
+public class SnapshotGenerationWindow {
```
eshryane, line 15: Window doesn't have to be configurable, we can hard-code a daily time to create a snapshot

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotGenerationWindow.java
```diff
@@ -0,0 +1,53 @@
...
+        to = LocalTime.of(Integer.parseInt(matcher.group(3)), Integer.parseInt(matcher.group(4)));
+    }
+
+    public boolean hasVersionExpired(final NrtmVersionInfo snapshotVersion) {
+        final long fromMs = from.toEpochSecond(dateTimeProvider.getCurrentDate(), ZoneOffset.UTC);
+        final long toMs = to.isBefore(from) ?
+            to.toEpochSecond(dateTimeProvider.getCurrentDate().plusDays(1), ZoneOffset.UTC) :
+            to.toEpochSecond(dateTimeProvider.getCurrentDate(), ZoneOffset.UTC);
+        final long expireTime = snapshotVersion.created() + toMs - fromMs;
+        return expireTime < dateTimeProvider.getCurrentDateTime().toEpochSecond(ZoneOffset.UTC);
```
eshryane, line 41: use `isBefore` two timestamps rather than calculating differences in milliseconds

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/SnapshotFileRepository.java
```diff
@@ -42,6 +42,20 @@ INSERT INTO snapshot_file (version_id, name, hash, payload)
...
     }
 
+    public Optional<SnapshotFile> getByVersionID(final long versionID) {
+        final String sql = """
+            SELECT sf.id, sf.version_id, sf.name, sf.hash
+            FROM snapshot_file sf
+            WHERE sf.version_id = ?
+            """;
+        try {
+            return Optional.ofNullable(jdbcTemplate.queryForObject(sql, rowMapper, versionID));
```
eshryane, line 52: Why use `ofNullable` rather than `of`? Can the value ever be `null`?

whois-commons/src/main/resources/nrtm_schema.sql
```diff
@@ -77,9 +77,10 @@ CREATE TABLE `delta_file`
 
 CREATE TABLE `notification_file`
 (
-    `id`         int unsigned NOT NULL AUTO_INCREMENT,
-    `version_id` int unsigned NOT NULL,
-    `payload`    longtext     NOT NULL,
+    `id`         int unsigned    NOT NULL AUTO_INCREMENT,
```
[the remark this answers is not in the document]
petchells: for debugging it's been handy to see the changes as new files are created. this behaviour can be changed once we know we're generating them consistently. for now i'd recommend keeping them.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotGenerationWindow.java
petchells: the RFC says: "3.2.2 The mirror server MUST generate a new Snapshot File between once per hour and once per day, if there have been changes to the IRR objects." so the idea of this 'window' is to prevent snapshots less than 1 hour old from being overwritten, in the case that the NRTM db is reset and the server rebooted just before midnight (say) and the snapshot timer ticks at midnight.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/SnapshotFileRepository.java
petchells: technically, yes (and intellij warns about it) but in practice an exception will be thrown and empty() will be returned. so it's really to stop intellij nagging -- it makes no difference for our code.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotGenerationWindow.java
eshryane: Can you run a scheduled job more frequently then and generate a snapshot if an existing one is >24h old? Then there's no need for a window.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/SnapshotFileRepository.java
eshryane: OK

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotGenerationWindow.java
petchells: not really, because then the time since the last snapshot will be 24h 14m since it takes 14m to generate (at least, it does on dev & prepdev) and we won't be compliant with the RFC. if we assume that it takes 14m and look for an expiry date at at 23h 47m, we must assume that 14m is consistent across all production machines every day, which is unlikely to be the case. so then the exact time of generation will slide every day, forwards or backwards depending on which machine picks up the job and how busy it happens to be. some days we might not generate a file in 14m and then we'll be non-compliant again. using the window gives us the best chance of remaining compliant while not ever needing manual intervention to reset the time of snapshots if the time slips too much in one direction. it's also an easy change to define two windows in one day, for example, to make very sure we'll always be compliant and hit the 'at least one snapshot every 24h' requirement. if you still want to generate a snapshot at a fixed time then of course that's fine, but i think it's important to understand why i wrote the code this way, and the implications of the fixed time / fixed expiry approaches.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotFileGenerator.java
```diff
@@ -8,6 +8,7 @@
 import net.ripe.db.nrtm4.domain.NrtmSourceModel;
```
MiguelAHM, line 8: you are not using the output of createSnapshots anywhere 
createSnapshotsForVersions visibility could be private (you are just using this method in this class)

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/NrtmVersionInfoRepository.java
```diff
@@ -22,26 +22,28 @@
...
 
 @Repository
 public class NrtmVersionInfoRepository {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(NrtmVersionInfoRepository.class);
-    private final JdbcTemplate jdbcTemplate;
-    private final DateTimeProvider dateTimeProvider;
-    private final RowMapper<NrtmVersionInfo> rowMapper = (rs, rowNum) -> {
-        final NrtmSourceModel source = new NrtmSourceModel(rs.getLong(2), CIString.ciString(rs.getString(3)));
+    static final Function<Integer, RowMapper<NrtmVersionInfo>> rowMapperWithOffset = (offset) -> (rs, rowNum) -> {
```
MiguelAHM, line 32: why do we need this offset?

petchells: i use it in two places. with a zero it will create an NrtmVersionInfo object using all the rows returned from the query. it's also used to get DeltaVersion, where a DeltaFile object and a NrtmVersionInfo object are fetched at the same time, so then i use the offset to point to the fields for the version object.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/SnapshotGenerationWindow.java
petchells: done

PR #1211, opened 2023-03-30
Seperate job nrtm notification

maggarwal13:


whois-nrtm4/src/main/java/net/ripe/db/nrtm4/Nrtmv4Condition.java
```diff
@@ -0,0 +1,16 @@
...
+
+import org.springframework.context.annotation.Condition;
+import org.springframework.context.annotation.ConditionContext;
+import org.springframework.core.type.AnnotatedTypeMetadata;
+
+public class Nrtmv4Condition implements Condition {
+
+    @Override
+    public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata) {
+        if( !context.getEnvironment().containsProperty("nrtm4.enabled")) {
```
MiguelAHM, line 11: if the property nrtm4.enabled is false and therefore we do not complies the condition what is going to happen?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/DeltaFileDao.java
```diff
@@ -90,6 +91,50 @@ public List<DeltaFileVersionInfo> getDeltasForNotification(final NrtmVersionInfo
...
+        try {
+            return jdbcTemplate.query(sql, rowMapperWithVersion, source.getId(), sinceTimestamp);
+        } catch (final DataAccessException e) {
+            LOGGER.warn("Exception in getDeltasForNotification", e);
+            return List.of();
+        }
+    }
+
+
+    public List<DeltaFileVersionInfo> getAllDeltas(final NrtmVersionInfo sinceVersion, final LocalDateTime since) {
```
MiguelAHM, line 116: I think that you are not using this method, aren't you?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/NotificationFileDao.java
```diff
@@ -38,19 +39,28 @@ INSERT INTO notification_file (version_id, created, payload)
...
                 JOIN version_info vi ON vi.id = nf.version_id
                 WHERE vi.source_id = ?
-                ORDER BY nf.created DESC LIMIT 1
                 """;
             return Optional.ofNullable(jdbcTemplate.queryForObject(sql, rowMapper, source.getId()));
         } catch (final EmptyResultDataAccessException e) {
             return Optional.empty();
+        } catch (final IncorrectResultSizeDataAccessException exception) {
+            LOGGER.error("More than one entry found for a source {} in update notification file", source.getName());
+            throw new IllegalStateException("More than one entry found for a source {} in update notification file");
```
MiguelAHM, line 63: Don't you need to parse the {} here? the final detailed message will keep {} instead of the source

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/NrtmVersionInfoRepository.java
```diff
@@ -70,7 +76,12 @@ public List<NrtmVersionInfo> findLastVersionPerSource() {
         }
     }
 
-    public NrtmVersionInfo findLastSnapshotVersionForSource(final NrtmSourceModel source) {
+    public Optional<NrtmVersionInfo> findLastVersion(final NrtmSource source) {
+       return findLastVersionPerSource().stream().filter( (versionInfo) ->  versionInfo.source().getName().equals(source.getName())).findFirst();
```
MiguelAHM, line 80: Here we are calling findLastVersionPerSource and returning the data per each source, when we just want the data for one of them. is it not better to perform the query just adding the condition of the source there?

whois-nrtm4/src/test/java/net/ripe/db/nrtm4/NrtmFileProcessorIntegrationTest.java
```diff
@@ -39,6 +40,7 @@
 
 
 @Tag("IntegrationTest")
+@Ignore
 public class NrtmFileProcessorIntegrationTest extends AbstractNrtm4IntegrationBase {
```
MiguelAHM, line 44: Why are we ignoring this test?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/Nrtmv4Condition.java
MiguelAHM: the bean is not going to be generated so the job is not going to run

eshryane (approved):

eshryane:
To be merged after the upcoming Whois release

maggarwal13:
Merged in https://github.com/RIPE-NCC/whois/pull/1219

PR #1218, opened 2023-04-11
feat: elasticsearch[ES] allow generate words part in email words delimiter

MiguelAHM:


maggarwal13:
Please change the oneshot script accordingly , rebuild the indexes in prepdev first and test it out after that. There should not be any need for whois release. Once all goes well , we can do same for production. There should not be any downtime for this change

whois-api/src/test/java/net/ripe/db/whois/api/ElasticSearchHelper.java
```diff
@@ -113,6 +112,11 @@ private XContentBuilder getSettings() throws IOException {
                             .field("preserve_original", true)
                             .field("split_on_case_change", true)
                         .endObject()
+                        .startObject("my_word_email_delimiter_graph")
```
MiguelAHM, line 115: By default the values of the rest of fields are
generate_word_parts -> true
catenate_words -> false
catenate_numbers -> false

I could add those fields for readability instead of skip them. The one that is fixing the problem is generate_word_parts -> true. This will generate all the combinations of the world as far as I understand.

whois-api/src/test/java/net/ripe/db/whois/api/ElasticSearchHelper.java
```diff
@@ -97,7 +96,7 @@ private XContentBuilder getSettings() throws IOException {
                         .startObject("my_email_analyzer")
                             .field("type", "custom")
                             .field("tokenizer", "uax_url_email")
-                            .field("filter", new String[]{"stop", "lowercase"})
+                            .field("filter", new String[]{"my_word_email_delimiter_graph", "lowercase", "english_stop"})
```
MiguelAHM, line 99: by default stop uses _english_ so here english_stop is the same as stop

whois-api/src/test/java/net/ripe/db/whois/api/fulltextsearch/ElasticFullTextSearchTestIntegration.java
```diff
@@ -1189,6 +1189,185 @@ public void dont_include_null_elements() {
...
+                        "org-type:     OTHER\n" +
+                        "descr:        test org\n" +
+                        "address:      street 1\n" +
+                        "e-mail:       test@domain1.domain2.nl\n" +
+                        "mnt-ref:      OWNER-MNT\n" +
+                        "mnt-by:       OWNER-MNT\n" +
+                        "source:       RIPE\n"));
+        rebuildIndex();
+
+        assertThat(numFound(query("q=test")), is(1L));
```
eshryane, line 1249: This is ambiguous, as the word `test` appears multiple times in the object. Can you make the local part of each email distinct?

eshryane:
Can you add "elasticsearch" or "full text search" to the PR title to make clear which area of the code is changing?

whois-api/src/test/java/net/ripe/db/whois/api/fulltextsearch/ElasticFullTextSearchTestIntegration.java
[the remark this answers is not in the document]
maggarwal13: It does matched but since all the matched term is in one object, it will return one object. May be add highlight tests? to show matched attributes?

MiguelAHM: change by testemail

eshryane:
> Please change the oneshot script accordingly , rebuild the indexes in prepdev first and test it out after that. There should not be any need for whois release. Once all goes well , we can do same for production. There should not be any downtime for this change. Later on we should also shift the job to whois but there will be a seperate story for this

Except if there's any updated objects with an email attribute, they will be indexed using the "old" method

whois-api/src/test/java/net/ripe/db/whois/api/fulltextsearch/ElasticFullTextSearchTestIntegration.java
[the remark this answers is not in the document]
eshryane: Ah I see. But, we can't tell then on which word the match was. I suggest removing duplicate words from the object, just use one distinct word to test with.

maggarwal13:
> > Please change the oneshot script accordingly , rebuild the indexes in prepdev first and test it out after that. There should not be any need for whois release. Once all goes well , we can do same for production. There should not be any downtime for this change. Later on we should also shift the job to whois but there will be a seperate story for this
> 
> Except if there's any updated objects with an email attribute, they will be indexed using the "old" method

No , It will save the last serial Id and as soon as we change the pointer to point to new index, it will update those objects as there is scheduled job that runs every minute

whois-api/src/test/java/net/ripe/db/whois/api/ElasticSearchHelper.java
eshryane: Why change `stop` to `english_stop` then?

MiguelAHM: I have been talking with Mahesh about it. We agree that for readability is better to keep english_stop

whois-api/src/test/java/net/ripe/db/whois/api/fulltextsearch/ElasticFullTextSearchTestIntegration.java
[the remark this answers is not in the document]
MiguelAHM: I have already removed the duplicated words as well, I am using testemail@domain1.domain2.nl instead of test@domain1...etc

eshryane:
> > > Please change the oneshot script accordingly , rebuild the indexes in prepdev first and test it out after that. There should not be any need for whois release. Once all goes well , we can do same for production. There should not be any downtime for this change. Later on we should also shift the job to whois but there will be a seperate story for this
> > 
> > 
> > Except if there's any updated objects with an email attribute, they will be indexed using the "old" method
> 
> No , It will save the last serial Id when it build a new indexes and as soon as we change the pointer to point to new index, it will update those objects as there is scheduled job that runs every minute

Understood, thanks! The new indexing code is deployed in a oneshot jar and used directly by elasticsearch

MiguelAHM:
> Please change the oneshot script accordingly , rebuild the indexes in prepdev first and test it out after that. There should not be any need for whois release. Once all goes well , we can do same for production. There should not be any downtime for this change. Later on we should also shift the job to whois but there will be a seperate story for this

please review https://github.com/RIPE-NCC/whois-internal-oneshot/pull/140 with the changes

MiguelAHM:
Change applied to one-shot scrip, tested in PREPDEV and working fine attached screen shots in Jira task

PR #1220, opened 2023-04-12
Limit resource lookups in RDAP organisation entity requests

eshryane:
Don't lookup every referenced network resource for RDAP organisation entity requests.

Instead find all top-level resources based on object key, and query for a limited amount.


MiguelAHM:
now we are performing a query and using the rest call, are we planning to stop using the rest call and use the database instead in a future refactor? (I can't see any related task in the backlog)

MiguelAHM:
> > now we are performing a query and using the rest call, are we planning to stop using the rest call and use the database instead in a future refactor? (I can't see any related task in the backlog)
> 
> I just added a task to do this

Thanks!

PR #1222, opened 2023-04-13
feat: add handling and write a better integration test

MiguelAHM:


whois-api/src/main/java/net/ripe/db/whois/api/httpserver/RequestParamRegexRule.java
```diff
@@ -39,6 +39,7 @@ public String matchAndApply(String target, HttpServletRequest request, HttpServl
     private String apply(final String target, final HttpServletResponse response) throws IOException {
         if (HttpStatus.isClientError(responseCode) || HttpStatus.isServerError(responseCode)) {
             response.sendError(responseCode);
+            this.setHandling(true);
```
eshryane, line 42: magic!

eshryane (approved):

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/RequestParamRegexRule.java
MiguelAHM: yes, found another one. In the rules the default one is returning a 400 if it is not able to process the path of the request. In that case the error message was a 400 Not Found as well :S 
I fixed that one as well.

eshryane:
Thanks for checking 👍

PR #1226, opened 2023-04-18
Fix RDAP language notices test

eshryane:


eshryane:
Integration tests passed

PR #1229, opened 2023-04-21
feat: add special character matcher and a generic test

MiguelAHM:


MiguelAHM:
Person rol and organisation consist on ASCII alphanumeric only. So it is not required to allow all the latin-1 table (just the ASCII)

PR #1230, opened 2023-04-26
Db 4604 filters inconsistence

MiguelAHM:


MiguelAHM:
added .custom search to cover all the fields. Take a look at this website for more information
https://www.elastic.co/guide/en/elasticsearch/reference/5.6/multi-fields.html

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/FulltextSearch.java
```diff
@@ -3,9 +3,12 @@
...
 import net.ripe.db.whois.api.rest.domain.Version;
 import net.ripe.db.whois.common.ApplicationVersion;
+import net.ripe.db.whois.common.rpsl.AttributeType;
 
 import java.io.IOException;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public abstract class FulltextSearch {
```
maggarwal13, line 14: If these functions are being used only in one implementation , any reason ts in abstract class?

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/FulltextSearch.java
```diff
@@ -54,6 +57,29 @@ protected SearchResponse.Lst createVersion() {
         return result;
     }
 
+    protected String customizeIfFiltering(final String value){
+        final Matcher matcher = Pattern.compile("\\((.*?\\))").matcher(value);
+        if (!matcher.find()) {
+            return value;
+        }
+        final String fullAttributeValue = matcher.group(1);
+        final String attributeType = fullAttributeValue.replace("(", "").split(":")[0];
```
maggarwal13, line 66: I think you are assuming too many things here. query can have multiple attribute, multiple AND OR operators. This might work for single attribute value. We need to look at first some setting or mappings in ES rather than manually modifying query like this

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/FulltextSearch.java
MiguelAHM: Didn't realise that this was an abstract class used by Lucene implementation as well. I will move the method. Thank you!

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/FulltextSearch.java
MiguelAHM: See the test cases. There is test cases for multiple attributes, or, and...multiple object types...etc
This is working for multiple values as well. (I take all this examples from db-web-ui calls)

I tried to find some settings or mappings but the bug is in the query...we should use all the filters that we are specifying, not just one

PR #1231, opened 2023-04-28
Db 4604 filters inconsistence

MiguelAHM:


whois-api/src/test/java/net/ripe/db/whois/api/ElasticSearchHelper.java
```diff
@@ -150,11 +152,17 @@ private XContentBuilder getMappings() throws IOException {
              if(type.getSyntax() == AttributeSyntax.EMAIL_SYNTAX) {
                  mappings.startObject(type.getName())
                              .field("type", "text")
+                             .field("analyzer", "fulltext_analyzer")
```
MiguelAHM, line 155: We will use this analyser as well in the e-mail, this is why is working now.
However, if we move this behind "custom" it doesn't work...we need to figure out why!

MiguelAHM: Here by default we are telling to use the full text analyser by default and the my_email_analyzer as a custom one. https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-fields.html

So it will try to find the match in the fulltext_analyzer and then in the custom one if don't find the solution.

MiguelAHM:
test ok -> https://gitlab.ripe.net/swe-database-team/whois/-/pipelines/340981
I have had to change one contains for containsInAny order to make it works. Please take a look!

PR #1232, opened 2023-05-01
Db 4577 return 426 when http

MiguelAHM:


whois-api/src/main/java/net/ripe/db/whois/api/nrtm4/NrtmClientService.java
```diff
@@ -103,6 +106,12 @@ private void validateSource(final String source, final String fileName) {
         }
     }
 
+    private void validateHttps(final HttpServletRequest httpServletRequest) {
+        boolean isHttps = Objects.equals(httpServletRequest.getHeader("X-Forwarded-Proto"), "HTTPS");
```
maggarwal13, line 110: But does the application handles this? I thought we do not allow http requests anyways. F5 does the redirect. And even if applocation needs to handle it should be a requestFilter?

MiguelAHM: In jira ticket:

Implementation Details
All requests come through the F5 LB which adds a X-Forwarded-Proto header. Check the value of this header, if it's set to "http" then return a 426 response.

maggarwal13: This should be job of  request filter. Can you create a filter instead and add it to WhoisNrtmServletDeployer ?

MiguelAHM: ok, let me check this

whois-api/src/test/java/net/ripe/db/whois/api/nrtmv4/NrtmClientServiceTestIntegration.java
```diff
@@ -349,12 +349,56 @@ public void should_throw_exception_invalid_source_notification_file()  {
     @Test
     public void should_throw_exception_delta_file_not_found()  {
         final Response response = createResource("TEST-NONAUTH/nrtm-delta.1.TEST-NONAUTH.60b9e8c4e4891411be.json")
-                .request(MediaType.APPLICATION_JSON)
+                .request(MediaType.APPLICATION_JSON).header("X-Forwarded-Proto", "HTTPS")
                 .get(Response.class);
```
maggarwal13, line 353: Can you create a abstract method that adds the header instead. You don't have to add in every call

whois-api/src/main/java/net/ripe/db/whois/api/exceptions/UpgradeException.java
```diff
@@ -0,0 +1,12 @@
+package net.ripe.db.whois.api.exceptions;
+
+import javax.ws.rs.WebApplicationException;
+
+public class UpgradeException extends WebApplicationException {
+
+    private static final long serialVersionUID = -6362082162191216040L;
+
```
maggarwal13, line 8: Are you sure it will return 429? There is NrtmExceptionMapper and it returns InternalServer error. Can you add a test case just to be sure?

maggarwal13: Also I don't think you need a custom exception for this. Throwing SC_UPGRADE_REQUIRED status should be enough

MiguelAHM: there are 2 test case in which we are returning 426 I added this exception in the mapper as well

whois-api/src/test/java/net/ripe/db/whois/api/AbstractNrtmIntegrationTest.java
```diff
@@ -189,21 +189,21 @@ protected String getSnapshotNameFromUpdateNotification(final PublishableNotifica
 
     protected Response getSnapshotFromUpdateNotificationBySource(final String sourceName) throws JsonProcessingException {
         final Response updateNotificationResponse = createResource(sourceName + "/update-notification-file.json")
-                .request(MediaType.APPLICATION_JSON)
+                .request(MediaType.APPLICATION_JSON).header("X-Forwarded-Proto", "HTTPS")
                 .get(Response.class);
```
maggarwal13, line 193: use HttpHeader.X_FORWARDED_PROTO

whois-api/src/main/java/net/ripe/db/whois/api/exceptions/UpgradeException.java
MiguelAHM: I added the exception because I think that it is more clear...we are using notfoundexception, badrequestexception from WebApplicationException...so create a new one which is not already implemented by the library is the correct way (to follow the same behaviour)

whois-api/src/main/java/net/ripe/db/whois/api/exceptions/UpgradeException.java
```diff
@@ -0,0 +1,12 @@
+package net.ripe.db.whois.api.exceptions;
+
+import javax.ws.rs.WebApplicationException;
+
+public class UpgradeException extends WebApplicationException {
+
```
maggarwal13, line 6: do you think UpgradeRequiredException is better name?

whois-api/src/test/java/net/ripe/db/whois/api/AbstractNrtmIntegrationTest.java
MiguelAHM: Done!

whois-api/src/main/java/net/ripe/db/whois/api/exceptions/UpgradeException.java
MiguelAHM: I deleted the exception, we are returning directly the error

whois-api/src/test/java/net/ripe/db/whois/api/nrtmv4/NrtmClientServiceTestIntegration.java
MiguelAHM: done

whois-api/src/main/java/net/ripe/db/whois/api/nrtm4/NrtmHttpControl.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.Objects;
+
+@Component
+public class NrtmHttpControl implements Filter {
+
+    @Override
+    public void doFilter(final ServletRequest request, final ServletResponse response, final FilterChain chain) throws IOException, ServletException {
+        final HttpServletRequest httpRequest = (HttpServletRequest) request;
+        if (isHttps(httpRequest)){
+            chain.doFilter(request, response);
```
maggarwal13, line 26: HttpScheme.HTTPS.is(request.getHeader(HttpHeader.X_FORWARDED_PROTO.asString())

whois-api/src/main/java/net/ripe/db/whois/api/nrtm4/NrtmHttpControl.java
```diff
@@ -0,0 +1,45 @@
...
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.util.Objects;
+
+@Component
+public class NrtmHttpControl implements Filter {
+
```
maggarwal13, line 21: NrtmHttpSchemeFilter

MiguelAHM: Done

whois-api/src/main/java/net/ripe/db/whois/api/nrtm4/NrtmHttpControl.java
MiguelAHM: Ah I see that "is" is ignoring the case, perfect!

MiguelAHM: Updated!

MiguelAHM:
Test ok -> https://gitlab.ripe.net/swe-database-team/whois/-/pipelines/340771

maggarwal13 (approved):

PR #1234, opened 2023-05-03
feat: skip lookup n hightlights

MiguelAHM:


whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
```diff
@@ -167,8 +169,12 @@ private SearchResponse.Lst createHighlights(final SearchHit hit) {
         final SearchResponse.Lst documentLst = new SearchResponse.Lst(hit.getId());
         final List<SearchResponse.Arr> documentArrs = Lists.newArrayList();
 
-
-        hit.getHighlightFields().forEach((attribute, highlightField) -> {
+        for (Map.Entry<String, HighlightField> highlightFieldMap : hit.getHighlightFields().entrySet()) {
```
eshryane, line 172: Why drop the `forEach`? I think it's more readable and does the same thing as the for loop and variables

MiguelAHM: I am using a foreach but in different way, in the other way you can not use the continue and I wanted to avoid and if/else that covers all of this logic.

MiguelAHM: I prefer my approach that this one:


        hit.getHighlightFields().forEach((attribute, highlightField) -> {
            if (!("lookup-key".equals(attribute) || "lookup-key.custom".equals(attribute))){
                if(attribute.contains(".custom")) {
                    final SearchResponse.Arr arr = new SearchResponse.Arr(StringUtils.substringBefore(highlightField.name(), ".custom"));
                    arr.setStr(new SearchResponse.Str(null, StringUtils.join(highlightField.getFragments(), ",")));
                    documentArrs.add(arr);

                    //Somehow if searched term contains "." highlight field custom has no vlue for it.
                } else if(!hit.getHighlightFields().containsKey(attribute + ".custom"))  {
                    final SearchResponse.Arr arr = new SearchResponse.Arr(highlightField.name());
                    arr.setStr(new SearchResponse.Str(null, StringUtils.join(highlightField.getFragments(), ",")));
                    documentArrs.add(arr);
                }
            }
        });

eshryane: Use `return` rather than `continue` in a `forEach` loop

MiguelAHM: yes, this should work. Ok, thank you! I will update

eshryane:
As discussed I suggest looking deeper at why `lookup-key` is in the index, if we don't need it then remove it altogether

maggarwal13:
> As discussed I suggest looking deeper at why `lookup-key` is in the index, if we don't need it then remove it altogether
We need the look-up key to use this for sorting and aggregations . Technically we  can rename it to primary-key to be consistent with RpslObject. But I just let it be lookup-key as Lucene code also uses the same name. I can explain on standup how its being used

MiguelAHM:
We are already storing the primary-key
https://dexter.ripe.net:9200/whois/_doc/17651681

nic-hdl: MD24115-RIPE
lookup-key: MD24115-RIPE

maggarwal13:
> We are already storing the primary-key https://dexter.ripe.net:9200/whois/_doc/17651681
> 
> nic-hdl: MD24115-RIPE lookup-key: MD24115-RIPE

nic-hdl can be different from primary key in most cases it is different use and is auto generated

MiguelAHM:
> > We are already storing the primary-key https://dexter.ripe.net:9200/whois/_doc/17651681
> > nic-hdl: MD24115-RIPE lookup-key: MD24115-RIPE
> 
> nic-hdl can be different from primary key in most cases it is different use and is auto generated

It is an example, I see that in the code we have 

        // primary key
        document.add(new IntPoint(PRIMARY_KEY_FIELD_NAME, rpslObject.getObjectId()));
        document.add(new StoredField(PRIMARY_KEY_FIELD_NAME, rpslObject.getObjectId()));

        // lookup key
        document.add(new SortedDocValuesField(LOOKUP_KEY_FIELD_NAME, new BytesRef(rpslObject.getKey().toString())));
        document.add(new StringField(LOOKUP_KEY_FIELD_NAME, rpslObject.getKey().toString(), Field.Store.YES));

maggarwal13:
> > > We are already storing the primary-key https://dexter.ripe.net:9200/whois/_doc/17651681
> > > nic-hdl: MD24115-RIPE lookup-key: MD24115-RIPE
> > 
> > 
> > nic-hdl can be different from primary key in most cases it is different use and is auto generated
> 
> It is an example, I see that in the code we have
> 
> ```
>     // primary key
>     document.add(new IntPoint(PRIMARY_KEY_FIELD_NAME, rpslObject.getObjectId()));
>     document.add(new StoredField(PRIMARY_KEY_FIELD_NAME, rpslObject.getObjectId()));
> 
>     // lookup key
>     document.add(new SortedDocValuesField(LOOKUP_KEY_FIELD_NAME, new BytesRef(rpslObject.getKey().toString())));
>     document.add(new StringField(LOOKUP_KEY_FIELD_NAME, rpslObject.getKey().toString(), Field.Store.YES));
> ```

Yes primary-key is object id which is different then rpslObject.getKey(). If it confusing for you we can rename the field but it will require a wois release and it will become different then Lucene , I suggest till we do not remove lucene keep it consistent

eshryane (approved):

MiguelAHM:
test ok 
https://gitlab.ripe.net/swe-database-team/whois/-/pipelines/342304

PR #1235, opened 2023-05-05
feat: truncate long fields

MiguelAHM:
When the there is a long field, that long field will be truncated. This is the suggestion given by ES to fix this error:
https://www.elastic.co/guide/en/elasticsearch/reference/current/highlighting.html

I could move the configuration related to highlights  that we have in whois (when we perform the query) here, so we don't need to deploy whois if we need to change the highlights configuration.

Another way to ignore big values is https://www.elastic.co/guide/en/elasticsearch/reference/current/ignore-above.html 
for example here they are just indexing those values with less than 20 characters. We are using 10922 value, but they suggest to use 8191 for non-ASCCI characters  (discarded this last way because it is not available to "custom" or "text" analyser)

### DIFF
--- a/whois-api/src/test/java/net/ripe/db/whois/api/ElasticSearchHelper.java
+++ b/whois-api/src/test/java/net/ripe/db/whois/api/ElasticSearchHelper.java
@@ -87,6 +87,11 @@ public class ElasticSearchHelper {
 
         final XContentBuilder indexSettings =  XContentFactory.jsonBuilder();
         indexSettings.startObject()
+                .startObject("index")
+                    .startObject("highlight")
+                        .field("max_analyzed_offset", "500000")
+                    .endObject()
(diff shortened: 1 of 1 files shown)


MiguelAHM:
This doesn't work, we need to update the query parameter. Updating this configuration we get:

whois.20230505:Caused by: org.elasticsearch.ElasticsearchException: Elasticsearch exception [type=illegal_argument_exception, reason=The length [1572863] of field [remarks.custom] in doc[140578]/index[whois-20230505-095216] exceeds the [index.highlight.max_analyzed_offset] limit [500000]. To avoid this error, set the query parameter [max_analyzed_offset] to a value less than index setting [500000] and this will tolerate long field values by truncating them.]

So we should leave the default 1.000.000 and add this value to the query parameter

PR #1241, opened 2023-05-22
Nrtm update notification signing

maggarwal13:


whois-api/src/test/java/net/ripe/db/whois/api/nrtmv4/NrtmClientServiceTestIntegration.java
```diff
@@ -261,10 +272,57 @@ public void should_get_update_notification_file() {
...
+        verifier.init(false, generateKeyPair().getPublic());
+        verifier.update(notificationFile.getBytes(), 0, notificationFile.getBytes().length);
+        assertThat(verifier.verifySignature(Hex.decodeHex(signature)), is(Boolean.FALSE));
+    }
+
+    @Test
+    public void should_throw_exeption_no_key_exists() {
+        insertUpdateNotificationFile();
+
+        final Response response = getResponseFromHttpsRequest("TEST/update-notification-file.json.sig", MediaType.APPLICATION_JSON);
```
eshryane, line 318: Use a String entity instead of response and use an `assertThrows` an `InternalErrorException` instead?

whois-commons/src/main/resources/nrtm_schema.sql
```diff
@@ -85,6 +86,19 @@ CREATE TABLE `notification_file`
 ) ENGINE = InnoDB
   DEFAULT CHARSET = utf8;
 
+
+CREATE TABLE `key_pair`
+(
+    `id`          int unsigned    NOT NULL AUTO_INCREMENT,
+    `private_key` VARBINARY(3000) NOT NULL,
+    `public_key`  VARBINARY(3000) NOT NULL,
+    `created`     bigint unsigned NOT NULL,
```
eshryane, line 95: Does it help to add an `expires` column rather than having to extract it from the key?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/Ed25519Util.java
```diff
@@ -0,0 +1,39 @@
...
+    public static String signWithEd25519(final byte[] payload, final byte[] privateKey)  {
+        try {
+            final Signer signer = new Ed25519Signer();
+            signer.init(true, new Ed25519PrivateKeyParameters(privateKey, 0));
+            signer.update(payload, 0, payload.length);
+            byte[] signature = signer.generateSignature();
+            return Hex.toHexString(signature);
+        } catch (CryptoException ex) {
+            LOGGER.error("failed to sign payload {}", ex.getMessage());
+            throw new RuntimeException("failed to sign contents of file");
```
eshryane, line 30: Use `IllegalStateException` instead as it's more specific

whois-api/src/test/java/net/ripe/db/whois/api/nrtmv4/NrtmClientServiceTestIntegration.java
```diff
@@ -420,4 +478,20 @@ INSERT INTO version_info (id, source_id, version, session_id, type, last_serial_
         final String payloadNonAuth = "{\"nrtm_version\":4,\"timestamp\":\"2023-03-13T12:31:08Z\",\"type\":\"snapshot\",\"source\":\"TEST-NONAUTH\",\"session_id\":\"7f42be43-a1fd-48e7-947f-91c852cf13f6\",\"version\":1,\"snapshot\":{\"version\":1,\"url\":\"https://nrtm.ripe.net//7f42be43-a1fd-48e7-947f-91c852cf13f6/nrtm-snapshot.1.RIPE-NONAUTH.9a6ca0a3e1b68eb65d4a2d277c3e5a96.json.gz\",\"hash\":\"8140b2d3a3ec4cb6a14c1ca19f7cbca468adcc172ad3e313aed74bd0d1036838\"}}";
         databaseHelper.getNrtmTemplate().update("INSERT INTO notification_file (version_id, created, payload) VALUES (?, ?, ?)",2, dateTimeProvider.getCurrentDateTime().toEpochSecond(ZoneOffset.UTC), payloadNonAuth);
     }
+
+    public AsymmetricCipherKeyPair generateKeyPair() {
```
eshryane, line 482: Move this code into a separate DAO? Can we call it automatically at startup if nrtmv4 is enabled and no key pair is found already?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/Ed25519Util.java
```diff
@@ -0,0 +1,39 @@
...
+import org.bouncycastle.crypto.params.Ed25519PrivateKeyParameters;
+import org.bouncycastle.crypto.signers.Ed25519Signer;
+import org.bouncycastle.util.encoders.Hex;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.security.SecureRandom;
+import java.util.Base64;
+
+public class Ed25519Util {
```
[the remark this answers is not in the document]
maggarwal13: But this will be used only in test class so better to seperate? But anyways does not matter much. I will shift it to util

whois-api/src/test/java/net/ripe/db/whois/api/nrtmv4/NrtmClientServiceTestIntegration.java
maggarwal13: Ok, in that case we should also call this method in case of generating new key pair while rotating? If its generated once a year I thought better to have it added manually.But I do not mind either way

whois-commons/src/main/resources/nrtm_schema.sql
eshryane: I wouldn't assume a 1 year expiry either, we could expire more frequently in future (e.g. Let's Encrypt certificates expire after 90 days). The expiry column is a shortcut so you don't need to extract the expiry from the key itself, and allows you to query by expiry date (also useful for a monitoring check).

whois-api/src/test/java/net/ripe/db/whois/api/nrtmv4/NrtmClientServiceTestIntegration.java
eshryane: Yes also call this when rotating the key. We should automate it so we don't forget to do it manually

whois-commons/src/main/resources/nrtm_schema.sql
MiguelAHM: we could create a new column called "end_date", which means that the key is not longer valid. 
From a job, we could calculate if the time ran out and modify that "end_date" and generate a new pair.

maggarwal13: I will add an expires column

MiguelAHM (approved):

MiguelAHM (approved):

eshryane (approved):

PR #1251, opened 2023-06-02
Include the Country Code in Public Organisation Split File

maggarwal13:
Include the Country Code in Public Organisation Split File


whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/DummifierNrtm.java
```diff
@@ -30,10 +30,9 @@ public class DummifierNrtm implements Dummifier {
             AttributeType.ZONE_C
     );
 
-    static final List<AttributeType> ATTRIBUTES_TO_KEEP = Lists.newArrayList(
-            AttributeType.ABUSE_C,
-            AttributeType.LAST_MODIFIED,
-            AttributeType.CREATED);
+    static final Map<ObjectType, List<AttributeType>> ATTRIBUTES_TO_KEEP_OBJECT_TYPE = Map.of(
```
eshryane, line 33: abuse-c is only an attribute on organisation, aut-num, inetnum, inet6num. Not on MNTNER ?

eshryane: Do we need to keep attributes by object type?

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/DummifierNrtm.java
```diff
@@ -66,14 +65,14 @@ public RpslObject dummify(final int version, final RpslObject rpslObject) {
         return new RpslObject(rpslObject, attributes);
     }
 
-    private void stripSomeNonMandatoryAttributes(List<RpslAttribute> attributes, ObjectType objectType) {
-        if (!STRIPPED_OBJECT_TYPES.contains(objectType)) {
+    private void stripSomeNonMandatoryAttributes(final List<RpslAttribute> attributes, final ObjectType objectType) {
+        if (!ATTRIBUTES_TO_KEEP_OBJECT_TYPE.containsKey(objectType)) {
```
eshryane, line 69: What happened to STRIPPED_OBJECT_TYPES ?

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/DummifierNrtm.java
eshryane: We can keep country for *all* object types and just add it to ATTRIBUTES_TO_KEEP ?

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/DummifierNrtm.java
MiguelAHM: I think that the idea is to merge two lists into one.
Anyway I think that we should remove STRIPPED_OBJECT_TYPES it not longer used with the last changes.

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/DummifierNrtm.java
MiguelAHM: If we add country to the current Attributes_to_keep it would be a bit confusing because mntner doesn't have country. So I suggest delete ABUSE_C from MNTNER and keep the map

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/DummifierNrtm.java
maggarwal13: Yes, Initially it was two variables. STRIPPED_OBJECT_TYPES and ATTRIBUTES_TO_KEEP. Since both are being used together I merged them

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/DummifierNrtm.java
maggarwal13: So basically just for ORG and MNTNER we strip all non mandatory attributes so thats why for these two types we need a list to keep optional attributes

maggarwal13: Regarding ABUSE_C It was added in old code so did not removed. We need to refactor the dummifier class after release and revisit it . So I just did minimal changes

eshryane: we want to keep created and last-modified attributes on *all* object types, so maybe change back to a `List` and allow country on all object types also (including resources)

maggarwal13: But this function is only for ORG and MNTNER, for rest of objects we do not strip any optional attributes

eshryane: OK! I see! The  field and function naming confused me. Then keep the Map, but remove ABUSE_C from MNTNER.

eshryane: Looks good thanks!

MiguelAHM (approved):

PR #1255, opened 2023-06-07
Fix number of results found returned in Full Text Search response

MiguelAHM:
- Refactoring ES to avoid using database. We will go with ES information, building the object from given attributes. This allows us to return the correct number of hits. There was an inconsistency between the total hits and the returned objects because we were deleting the objects that no longer exist from the response.
- numFound should return the total hits from ES. This value could not match with the number of returned objects. If rows is 3 and ES has 5 hits, we will return 5 as numFound and 3 records.
- Finally, we will not allow to request more than 10.000 objects, by default ES trim the output to 10k objects but we are able to modify that behaviour with rows parameter. We will return a 400 if the user set >10.000 to that parameter. This has been done because of the ES performance

### DIFF
--- a/whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
@@ -36,6 +36,7 @@ import org.springframework.stereotype.Component;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
@@ -171,7 +172,7 @@ public class ElasticFulltextSearch extends FulltextSearch {
         final Set<AttributeType> templateAttributes = ObjectTemplate.getTemplate(objectType).getAllAttributes();
 
 
-        for (final RpslAttribute rpslAttribute : fullTextIndex.filterRpslAttributes(templateAttributes, hit.getSourceAsMap())) {
+        for (final RpslAttribute rpslAttribute : filterRpslAttributes(templateAttributes, hit.getSourceAsMap())) {
             if(rpslAttribute.getType().equals(AttributeType.ABUSE_MAILBOX)){
                 abuseMailBoxAttributes.add(rpslAttribute);
             }
@@ -180,7 +181,30 @@ public class ElasticFulltextSearch extends FulltextSearch {
         return attributes;
     }
 
+    private List<RpslAttribute> filterRpslAttributes(final Set<AttributeType> attributeTypes, final Map<String, Object> hitAttributes) {
 
+        List<RpslAttribute> attributes = Lists.newArrayList();
+
+        for (final AttributeType attribute : attributeTypes) {
+            if (FullTextIndex.SKIPPED_ATTRIBUTES.contains(attribute)) {
+                continue;
+            }
+            final Object attributeValues = hitAttributes.get(attribute.getName());
+            if (attributeValues == null){
+                continue;
+            }
+            if (attributeValues instanceof List) {
+                for (final String attributeValue: (List<String>) attributeValues) {
+                    attributes.add(new RpslAttribute(attribute, fullTextIndex.filterRpslAttribute(attribute,
+                            attributeValue)));
+                }
+            } else {
+                attributes.add(new RpslAttribute(attribute, fullTextIndex.filterRpslAttribute(attribute,
+                        (String) attributeValues)));
+            }
+        }
+        return attributes;
+    }
     private String getHighlightTag(final String format, final String highlightPost) {
         return SearchRequest.XML_FORMAT.equals(format) ? escape(highlightPost) : highlightPost;
     }
--- a/whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
@@ -6,7 +6,6 @@ import net.ripe.db.whois.api.autocomplete.ElasticSearchCondition;
 import net.ripe.db.whois.api.elasticsearch.ElasticIndexService;
 import net.ripe.db.whois.api.elasticsearch.ElasticSearchAccountingCallback;
 import net.ripe.db.whois.common.ApplicationVersion;
-import net.ripe.db.whois.common.dao.RpslObjectDao;
 import net.ripe.db.whois.common.rpsl.AttributeType;
 import net.ripe.db.whois.common.rpsl.ObjectTemplate;
 import net.ripe.db.whois.common.rpsl.ObjectType;
@@ -30,7 +29,6 @@ import org.elasticsearch.search.sort.SortBuilders;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Conditional;
 import org.springframework.stereotype.Component;
@@ -56,7 +54,6 @@ public class ElasticFulltextSearch extends FulltextSearch {
     private final ElasticIndexService elasticIndexService;
 
     private final Source source;
-    private final RpslObjectDao objectDao;
 
     //Truncate after 100k of characters
     private static final int HIGHLIGHT_OFFSET_SIZE = 100000;
@@ -66,7 +63,6 @@ public class ElasticFulltextSearch extends FulltextSearch {
     @Autowired
     public ElasticFulltextSearch(final FullTextIndex fullTextIndex,
                                  final ElasticIndexService elasticIndexService,
-                                 @Qualifier("jdbcRpslObjectSlaveDao") final RpslObjectDao rpslObjectDao,
                                  final AccessControlListManager accessControlListManager,
                                  final SourceContext sourceContext,
                                  final ApplicationVersion applicationVersion,
@@ -76,7 +72,6 @@ public class ElasticFulltextSearch extends FulltextSearch {
         this.fullTextIndex = fullTextIndex;
         this.accessControlListManager = accessControlListManager;
         this.source = sourceContext.getCurrentSource();
-        this.objectDao = rpslObjectDao;
         this.maxResultSize = maxResultSize;
         this.elasticIndexService = elasticIndexService;
     }
@@ -89,6 +84,9 @@ public class ElasticFulltextSearch extends FulltextSearch {
             throw new IllegalArgumentException("Too many results requested, the maximum allowed is " + maxResultSize);
         }
 
+        if (searchRequest.getRows() < searchRequest.getStart()){
+            throw new IllegalArgumentException("Start parameter can not be bigger than rows parameter");
+        }
         return new ElasticSearchAccountingCallback<SearchResponse>(accessControlListManager, remoteAddr, source) {
 
             @Override
--- a/whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
@@ -114,7 +114,7 @@ public class ElasticFulltextSearch extends FulltextSearch {
                 final org.elasticsearch.action.search.SearchResponse fulltextResponse = elasticIndexService.getClient().search(fulltextRequest, RequestOptions.DEFAULT);
                 final SearchHit[] hits = fulltextResponse.getHits().getHits();
 
-                LOGGER.info("ElasticSearch {} hits for the query: {}", hits.length, searchRequest.getQuery());
+                LOGGER.debug("ElasticSearch {} hits for the query: {}", hits.length, searchRequest.getQuery());
                 return prepareResponse(fulltextResponse, hits, searchRequest, stopwatch);
             }
         }.search();
--- a/whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
@@ -35,7 +35,6 @@ import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Conditional;
 import org.springframework.stereotype.Component;
 
-import javax.ws.rs.BadRequestException;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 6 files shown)


whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
```diff
@@ -111,51 +114,70 @@ protected SearchResponse doSearch() throws IOException {
                 final org.elasticsearch.action.search.SearchResponse fulltextResponse = elasticIndexService.getClient().search(fulltextRequest, RequestOptions.DEFAULT);
                 final SearchHit[] hits = fulltextResponse.getHits().getHits();
 
-                LOGGER.debug("ElasticSearch {} hits for the query: {}", hits.length, searchRequest.getQuery());
-                final List<RpslObject> results = Lists.newArrayList();
-                int resultSize = Math.min(maxResultSize,Long.valueOf(fulltextResponse.getHits().getTotalHits().value).intValue());
+                LOGGER.info("ElasticSearch {} hits for the query: {}", hits.length, searchRequest.getQuery());
```
eshryane, line 117: Make this DEBUG or we'll be seeing it constantly in the console log

whois-api/src/test/java/net/ripe/db/whois/api/fulltextsearch/ElasticFullTextSearchTestIntegration.java
```diff
@@ -122,26 +123,25 @@ public void search_single_result() {
...
-         rebuildIndex();
+        rebuildIndex();
         databaseHelper.deleteObject(mntner);
 
         final QueryResponse queryResponse = query("q=DEV-MNT");
 
         assertThat(queryResponse.getStatus(), is(0));
-        assertThat(queryResponse.getResults().getNumFound(), is(0L));
-        assertThat(queryResponse.getResults(), hasSize(0));
+        assertThat(queryResponse.getResults().getNumFound(), is(1L));
```
eshryane, line 143: I can't believe we we testing for zero when there was one result! 🤦

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
MiguelAHM: oh yes sorry, forgot to revert the change!

whois-api/src/test/java/net/ripe/db/whois/api/fulltextsearch/ElasticFullTextSearchTestIntegration.java
MiguelAHM: yes, because previously we were checking the object in database...
so this is why the results was 0...because we were comparing what we had in ES and database and that was an inconsistence between both. 
Now we are using just the ES output so that record will be there

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
```diff
@@ -112,50 +115,69 @@ protected SearchResponse doSearch() throws IOException {
...
+        }.search();
+    }
+
+    private SearchResponse prepareResponse(final org.elasticsearch.action.search.SearchResponse fulltextResponse,
+                                           final SearchHit[] hits,
+                                           final SearchRequest searchRequest,
+                                           final Stopwatch stopwatch) {
+        final List<SearchResponse.Lst> highlightDocs = Lists.newArrayList();
+        final List<SearchResponse.Result.Doc> resultDocumentList = Lists.newArrayList();
+        fillHighlightsAndResults(hits, highlightDocs, resultDocumentList);
```
maggarwal13, line 130: Can you separate the highlights and results? Anyways you are calling  createHighlights inside the function

MiguelAHM: I spent some time thinking on this, if I do that I need to perform another loop, exactly the same look for hits.
Finally I choose to do everything in the same loop and putting a good name to the method. I could split in two but I am not fancy of that idea

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
```diff
@@ -112,50 +115,69 @@ protected SearchResponse doSearch() throws IOException {
...
 
-                for (final SearchHit hit : hits) {
-                    try {
-                        final RpslObject rpslObject = objectDao.getById(Integer.parseInt(hit.getId()));
-                        results.add(rpslObject);
-                        account(rpslObject);
+        final List<SearchResponse.Lst> responseLstList = Lists.newArrayList();
 
-                        highlightDocs.add(createHighlights(hit));
+        responseLstList.add(getResponseHeader(searchRequest, stopwatch.elapsed(TimeUnit.MILLISECONDS)));
```
maggarwal13, line 140: Why response header function needs a stopwatch as parameter?

MiguelAHM: yes, it is weird. Thank you for pointing on this

MiguelAHM: It seems that is not something new look master 
FulltextSearch.getResponseHeader
it is already like that in master...let me think how to refactor this...

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
```diff
@@ -112,50 +115,69 @@ protected SearchResponse doSearch() throws IOException {
...
+
+        final SearchResponse searchResponse = new SearchResponse();
+        searchResponse.setResult(result);
 
-                for (final SearchHit hit : hits) {
-                    try {
-                        final RpslObject rpslObject = objectDao.getById(Integer.parseInt(hit.getId()));
-                        results.add(rpslObject);
-                        account(rpslObject);
+        final List<SearchResponse.Lst> responseLstList = Lists.newArrayList();
```
maggarwal13, line 137: You removed the account method this method counts the personal objects limit and throws exception in case user breaches the limit. Where are you accounting for the personal objects now?  I think there is a test case also if not ca you add one?

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
MiguelAHM: we can discuss abut it. 
For getting the search response we need to loop hits
For getting the highlight response the same....we need to loop hits

I don't want to loop hits twice

whois-api/src/test/java/net/ripe/db/whois/api/fulltextsearch/ElasticFullTextSearchTestIntegration.java
```diff
@@ -249,11 +248,12 @@ public void search_multiple_results_with_search_limit() {
 
          rebuildIndex();
 
-        final QueryResponse queryResponse = query("q=remark&facet=true");
+        final QueryResponse queryResponse = query("q=remark&facet=true&rows=3");
 
-        //search limit to 3
+        //search limit to 3, however the total that ES is able to find is 5
         assertThat(queryResponse.getStatus(), is(0));
-        assertThat(queryResponse.getResults().getNumFound(), is(3L));
```
maggarwal13, line 256: If you set the system property limit as 10 instead of 3 this will anyways return 5 objects? I think the behaviour is changed now it checking the row number and object instead of limit. Can you rename the method?

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/FullTextIndex.java
```diff
@@ -307,7 +307,31 @@ public RpslObject filterRpslObject(final RpslObject rpslObject) {
 
         return new RpslObject(rpslObject.getObjectId(), attributes);
     }
-    
+
+    public List<RpslAttribute> filterRpslAttributes(final Set<AttributeType> attributeType, final Map<String, Object> hitAttributes) {
+
+        List<RpslAttribute> attributes = Lists.newArrayList();
```
maggarwal13, line 313: FulltextIndex is also being used by lucene . If its specific to elastic can you shift this code in that class itself and remove the reference to fulltextIndex?

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/FullTextIndex.java
```diff
@@ -307,7 +307,31 @@ public RpslObject filterRpslObject(final RpslObject rpslObject) {
 
         return new RpslObject(rpslObject.getObjectId(), attributes);
     }
-    
+
+    public List<RpslAttribute> filterRpslAttributes(final Set<AttributeType> attributeType, final Map<String, Object> hitAttributes) {
```
maggarwal13, line 311: rename atributeType to attributeTypes  and attribute to attributeType?

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/FullTextIndex.java
```diff
@@ -307,7 +307,31 @@ public RpslObject filterRpslObject(final RpslObject rpslObject) {
...
+
+        List<RpslAttribute> attributes = Lists.newArrayList();
+
+        for (final AttributeType attribute : attributeType) {
+            if (SKIPPED_ATTRIBUTES.contains(attribute)) {
+                continue;
```
maggarwal13, line 317: Can you remove the SKIPPED_ATTRIBUTES before iterating from attributeType? Also this whole logic is already in ElasticAutocomplete Can you reuse it and put this in some helper class?

whois-api/src/test/java/net/ripe/db/whois/api/fulltextsearch/ElasticFullTextSearchTestIntegration.java
```diff
@@ -153,15 +153,13 @@ public void search_single_result_json() {
...
 
         assertThat(searchResponse.getResult().getDocs(), hasSize(1));
-        assertThat(searchResponse.getResult().getDocs().get(0).getStrs(), hasSize(4));
-        assertThat(searchResponse.getResult().getDocs().get(0).getStrs().get(0).getName(), is("primary-key"));
-        assertThat(searchResponse.getResult().getDocs().get(0).getStrs().get(0).getValue(), is("1"));
-        assertThat(searchResponse.getResult().getDocs().get(0).getStrs().get(1).getName(), is("object-type"));
-        assertThat(searchResponse.getResult().getDocs().get(0).getStrs().get(1).getValue(), is("mntner"));
-        assertThat(searchResponse.getResult().getDocs().get(0).getStrs().get(2).getName(), is("lookup-key"));
+        assertThat(searchResponse.getResult().getDocs().get(0).getStrs(), hasSize(3));
+        assertThat(searchResponse.getResult().getDocs().get(0).getStrs().get(0).getName(), is("object-type"));
```
maggarwal13, line 157: why the result size decreased from 4 to 3? What happen to primary-key?

whois-api/src/test/java/net/ripe/db/whois/api/fulltextsearch/ElasticFullTextSearchTestIntegration.java
```diff
@@ -1745,7 +1739,7 @@ public void search_organisation_with_exact_match_multiple_object_type() {
         rebuildIndex();
         final QueryResponse queryResponse = query("facet=true&format=xml&hl=true&q=(remarks:(secondDomain%20nl))" +
                 "+AND+" +
-                "(object-type:organisation+OR+object-type:person)&start=0&wt=json");
+                "(object-type:organisation+OR+object-type:person)&start=0&wt=json&rows=3");
         assertThat(queryResponse.getResults().getNumFound(), is(1L));
```
maggarwal13, line 1743: why you need to add rows as 3?

MiguelAHM: I was testing the limit, then I see that there was already a test for that and I forgot to delete the rows from the parameter.
So I will delete that rows

whois-api/src/test/java/net/ripe/db/whois/api/fulltextsearch/ElasticFullTextSearchTestIntegration.java
MiguelAHM: - If you set the system property limit as 10 instead of 3 this will anyways return all the objects? Yes, It would return all of them (5)
- I think the behaviour is changed now it checking the row number and object instead of limit. -> Yes, we are not checking the limit I will change the name

maggarwal13: Well, we are limiting the output right? Not in this test case limit never came into picture here as there are just  5 objects and limit is 10 so it will never breach it

whois-api/src/test/java/net/ripe/db/whois/api/fulltextsearch/ElasticFullTextSearchTestIntegration.java
```diff
@@ -2130,6 +2124,15 @@ public void query_returns_maximum_results_and_mixed_objects_sorted_by_score_look
         assertThat(queryResponse.getResults().get(9).get("lookup-key"), is("83.92.220.64 - 83.92.220.71"));
     }
 
+    @Test
+    public void request_more_than_allowed_rows_bad_request() {
+        final BadRequestException badRequestException = assertThrows(BadRequestException.class, () -> {
```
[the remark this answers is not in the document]
MiguelAHM: if the total hits is 5 (0, 1, 2, 3, 4): 

- rows 3, start 1 => will get 1,2,3,
- rows 3, start 2 => will get 2,3,4
- rows 3, start 3 => will get 3,4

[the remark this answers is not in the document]
MiguelAHM: somehow ES is returning one value for this, the third value. I will return a badRequest for this

whois-api/src/test/java/net/ripe/db/whois/api/fulltextsearch/ElasticFullTextSearchTestIntegration.java
MiguelAHM: I renamed by search_multiple_results_with_first_three_records()

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
MiguelAHM: pushed using the account

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/FullTextIndex.java
MiguelAHM: " and attribute to attributeType"
I the second statement of your comment. About what attribute are you talking about?

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/FullTextIndex.java
MiguelAHM: the output in ElasticAutocomplete is a list of maps and here we have just a list. I could create a generic type and transforming that type but I think that it will make the logic more complicated

MiguelAHM: about to move SKIPPED_ATTRIBUTES we need to iterate the list of attributes to skip those attributes that we don't want.

MiguelAHM:
Moved to this branch refcator_ES

PR #1257, opened 2023-06-09
Fix the number of results found in Elasticsearch

MiguelAHM:
- Refactoring ES to avoid using database. We will go with ES information, building the object from given attributes. This allows us to return the correct number of hits. There was an inconsistency between the total hits and the returned objects because we were deleting the objects that no longer exist from the response.
- numFound should return the total hits from ES. This value could not match with the number of returned objects. If rows is 3 and ES has 5 hits, we will return 5 as numFound and 3 records.
- Finally, we will not allow to request more than 10.000 objects, by default ES trim the output to 10k objects but we are able to modify that behaviour with rows parameter. We will return a 400 if the user set >10.000 to that parameter. This has been done because of the ES performance
- Requested Rows could be smaller than start rows. Imagine that we have 50k results, and we want to start from 10k and return just 900 rows. So the output will be numFound 50k and the number of objects returned 900.

### DIFF
--- a/whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
@@ -36,6 +36,7 @@ import org.springframework.stereotype.Component;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
@@ -171,7 +172,7 @@ public class ElasticFulltextSearch extends FulltextSearch {
         final Set<AttributeType> templateAttributes = ObjectTemplate.getTemplate(objectType).getAllAttributes();
 
 
-        for (final RpslAttribute rpslAttribute : fullTextIndex.filterRpslAttributes(templateAttributes, hit.getSourceAsMap())) {
+        for (final RpslAttribute rpslAttribute : filterRpslAttributes(templateAttributes, hit.getSourceAsMap())) {
             if(rpslAttribute.getType().equals(AttributeType.ABUSE_MAILBOX)){
                 abuseMailBoxAttributes.add(rpslAttribute);
             }
@@ -180,7 +181,30 @@ public class ElasticFulltextSearch extends FulltextSearch {
         return attributes;
     }
 
+    private List<RpslAttribute> filterRpslAttributes(final Set<AttributeType> attributeTypes, final Map<String, Object> hitAttributes) {
 
+        List<RpslAttribute> attributes = Lists.newArrayList();
+
+        for (final AttributeType attribute : attributeTypes) {
+            if (FullTextIndex.SKIPPED_ATTRIBUTES.contains(attribute)) {
+                continue;
+            }
+            final Object attributeValues = hitAttributes.get(attribute.getName());
+            if (attributeValues == null){
+                continue;
+            }
+            if (attributeValues instanceof List) {
+                for (final String attributeValue: (List<String>) attributeValues) {
+                    attributes.add(new RpslAttribute(attribute, fullTextIndex.filterRpslAttribute(attribute,
+                            attributeValue)));
+                }
+            } else {
+                attributes.add(new RpslAttribute(attribute, fullTextIndex.filterRpslAttribute(attribute,
+                        (String) attributeValues)));
+            }
+        }
+        return attributes;
+    }
     private String getHighlightTag(final String format, final String highlightPost) {
         return SearchRequest.XML_FORMAT.equals(format) ? escape(highlightPost) : highlightPost;
     }
--- a/whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
@@ -6,7 +6,6 @@ import net.ripe.db.whois.api.autocomplete.ElasticSearchCondition;
 import net.ripe.db.whois.api.elasticsearch.ElasticIndexService;
 import net.ripe.db.whois.api.elasticsearch.ElasticSearchAccountingCallback;
 import net.ripe.db.whois.common.ApplicationVersion;
-import net.ripe.db.whois.common.dao.RpslObjectDao;
 import net.ripe.db.whois.common.rpsl.AttributeType;
 import net.ripe.db.whois.common.rpsl.ObjectTemplate;
 import net.ripe.db.whois.common.rpsl.ObjectType;
@@ -30,7 +29,6 @@ import org.elasticsearch.search.sort.SortBuilders;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Conditional;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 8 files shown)


eshryane:
Can you explain how the objects in the response are affected by the change to read from the Elasticsearch index instead of from the RIPE database itself? Is there *any* change? Which attributes will not be returned?

MiguelAHM:
> Can you explain how the objects in the response are affected by the change to read from the Elasticsearch index instead of from the RIPE database itself? Is there _any_ change? Which attributes will not be returned?

In the there is not going to be any change, the output will be the same...we are storing the debase internal id in ElasticSearch, so we are able to show that identifier as we were doing until now.

eshryane:
> > Can you explain how the objects in the response are affected by the change to read from the Elasticsearch index instead of from the RIPE database itself? Is there _any_ change? Which attributes will not be returned?
> 
> In the end there is not going to be any change in the output, it will be the same...we are storing the database object's internal id in ElasticSearch, so we are able to show that identifier as we were doing until now.

Maybe we should stop returning the object_id as `"primary-key"` as it can't be used as a primary key in the API, and it's internal only.

We *do* return RPSL attributes as `"str"` elements, are these currently read from the database or from the ES index?

MiguelAHM:
> > > Can you explain how the objects in the response are affected by the change to read from the Elasticsearch index instead of from the RIPE database itself? Is there _any_ change? Which attributes will not be returned?
> > 
> > 
> > In the end there is not going to be any change in the output, it will be the same...we are storing the database object's internal id in ElasticSearch, so we are able to show that identifier as we were doing until now.
> 
> Maybe we should stop returning the object_id as `"primary-key"` as it can't be used as a primary key in the API, and it's internal only.
> 
> We _do_ return RPSL attributes as `"str"` elements, are these currently read from the database or from the ES index?

Always from the index, the only one that we were reading from the database was the objectId. After this PR is merged we will not use the database for nothing, just ElasticSearch

eshryane:
> > > > Can you explain how the objects in the response are affected by the change to read from the Elasticsearch index instead of from the RIPE database itself? Is there _any_ change? Which attributes will not be returned?
> > > 
> > > 
> > > In the end there is not going to be any change in the output, it will be the same...we are storing the database object's internal id in ElasticSearch, so we are able to show that identifier as we were doing until now.
> > 
> > 
> > Maybe we should stop returning the object_id as `"primary-key"` as it can't be used as a primary key in the API, and it's internal only.
> > We _do_ return RPSL attributes as `"str"` elements, are these currently read from the database or from the ES index?
> 
> Always from the index, the only one that we were reading from the database was the objectId. After this PR is merged we will not use the database for nothing, just ElasticSearch

Understood thanks! Then I have no issue with this. We can drop the primary-key some other time.

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
```diff
@@ -82,80 +80,135 @@ public ElasticFulltextSearch(final FullTextIndex fullTextIndex,
...
         final Stopwatch stopwatch = Stopwatch.createStarted();
 
-        if (searchRequest.getStart() > maxResultSize) {
-            throw new IllegalArgumentException("Too many rows");
+        if (searchRequest.getRows() > maxResultSize) {
+            throw new IllegalArgumentException("Too many results requested, the maximum allowed is " + maxResultSize);
         }
 
+        if (searchRequest.getRows() < searchRequest.getStart()){
+            throw new IllegalArgumentException("Start parameter can not be bigger than rows parameter");
```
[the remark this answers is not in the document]
eshryane: I think this is a useful sanity check, that's not there at the moment. The client should not be allowed to `start` after the last result. 

For example, if you search for "SHRYANE" you get 9 results, but if you specify `start=10` you get an OK response with no results. You should get a bad request instead.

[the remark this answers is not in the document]
maggarwal13: No, I think we have different undertsanding how rows and start works

whois-api/src/test/java/net/ripe/db/whois/api/fulltextsearch/ElasticFullTextSearchTestIntegration.java
```diff
@@ -223,7 +223,8 @@ public void search_multiple_results_with_facet() {
     }
 
     @Test
-    public void search_multiple_results_with_search_limit() {
+    public void search_multiple_results_with_first_three_records() {
+
```
maggarwal13, line 227: Can you add a testcase where personal accounting is increased when we did a fulltext search?

MiguelAHM: not easy :S This is why I think that we don't have that test case already
I have already tried and I could perform that test but I need to change the constant of max people to a constant

MiguelAHM: I tried using it but this works when you perform the search before processing the ES output. I have already debugged that. But yes, lets talk separately

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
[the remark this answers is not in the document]
MiguelAHM: Yes, you are right, we shouldn't be checking this....Just read this now :S 
I added this use case in the description to avoid confusions

whois-api/src/test/java/net/ripe/db/whois/api/fulltextsearch/ElasticFullTextSearchTestIntegration.java
MiguelAHM: ok, done as discussed via Mattermost. I am adding another test for the counting error that I said you

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/FullTextIndex.java
```diff
@@ -70,7 +70,7 @@ public class FullTextIndex extends RebuildableIndex {
 
     static final String[] FIELD_NAMES;
 
-    private static final Set<AttributeType> SKIPPED_ATTRIBUTES = Sets.newEnumSet(Sets.newHashSet(AttributeType.CERTIF, AttributeType.CHANGED, AttributeType.SOURCE), AttributeType.class);
+    static final Set<AttributeType> SKIPPED_ATTRIBUTES = Sets.newEnumSet(Sets.newHashSet(AttributeType.CERTIF, AttributeType.CHANGED, AttributeType.SOURCE), AttributeType.class);
```
maggarwal13, line 73: Can you instead use the FILTERED_ATTRIBUTES and SKIPPED_ATTRIBUTES from ElasticIndexService as you already have that dependency in ElasticFulltextSearch ? This class is only for Lucene

whois-api/src/test/java/net/ripe/db/whois/api/fulltextsearch/ElasticFullTextSearchTestIntegration.java
```diff
@@ -912,6 +914,40 @@ public void permanent_block() {
...
+            query("q=john%20mcdonald");
+            fail("request should have been blocked");
+        } catch (ClientErrorException cee) {
+            assertThat(cee.getResponse().getStatus(), is(429));
+        }
+    }
+
+    @Test
+    public void check_personal_Accounting() {
+        testPersonalObjectAccounting.accountPersonalObject(Inet4Address.getLoopbackAddress(), 1);
```
maggarwal13, line 937: rename to should_account_for_personal_objects?

maggarwal13 (approved):

MiguelAHM:
Test ok -> https://gitlab.ripe.net/swe-database-team/whois/-/pipelines/359087

PR #1261, opened 2023-06-15
Return 400 Bad Request if Full Text Search client exceeds maximum documents

MiguelAHM:
If the user tries to fetch doc contained in > 100k position we should throw a 400.

ES by default has a limitation of 10k documents retrieved by ES, but we extended that limitation to 100k so we can fetch until that position.

### DIFF
--- a/whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
@@ -60,6 +60,7 @@ public class ElasticFulltextSearch extends FulltextSearch {
     //Truncate after 100k of characters
     private static final int HIGHLIGHT_OFFSET_SIZE = 100000;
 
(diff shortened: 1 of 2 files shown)


whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
```diff
@@ -84,6 +85,10 @@ public SearchResponse performSearch(final SearchRequest searchRequest, final Str
             throw new IllegalArgumentException("Too many results requested, the maximum allowed is " + maxResultSize);
         }
 
+        if (searchRequest.getStart() + searchRequest.getRows() > MAX_ROW_LIMIT_SIZE) {
+            throw new IllegalArgumentException("The maximum doc position to fetch is " + MAX_ROW_LIMIT_SIZE);
```
eshryane, line 89: Reword to `"Exceeded maximum " + MAX_ROW_LIMIT_SIZE + " documents"`

eshryane (approved):

PR #1263, opened 2023-06-20
BadRequest 400 when ES doesn't fit the query

MiguelAHM:
We are returning an internal error when ES is not accepting the keywords (AND/OR/NOT)

### DIFF
--- a/whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
@@ -138,16 +138,16 @@ public class ElasticFulltextSearch extends FulltextSearch {
     }
 
     private org.elasticsearch.action.search.SearchResponse performFulltextSearch(final SearchRequest searchRequest) throws IOException {
-        org.elasticsearch.action.search.SearchResponse response;
         try {
-            response = elasticIndexService.getClient().search(getFulltextRequest(searchRequest), RequestOptions.DEFAULT);
+            return elasticIndexService.getClient().search(getFulltextRequest(searchRequest), RequestOptions.DEFAULT);
         } catch (ElasticsearchStatusException ex){
             if (ex.status().equals(RestStatus.BAD_REQUEST)){
-                throw new IllegalArgumentException("The query syntax doesn't fit ES standards, please see the documentation for more info");
+                LOGGER.info("ElasticFullTextSearch fails due to the query: " + ex.getMessage());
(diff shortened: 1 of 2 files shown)


whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
```diff
@@ -136,7 +138,16 @@ protected SearchResponse doSearch() throws IOException {
     }
 
     private org.elasticsearch.action.search.SearchResponse performFulltextSearch(final SearchRequest searchRequest) throws IOException {
-        return elasticIndexService.getClient().search(getFulltextRequest(searchRequest), RequestOptions.DEFAULT);
+        org.elasticsearch.action.search.SearchResponse response;
+        try {
+            response = elasticIndexService.getClient().search(getFulltextRequest(searchRequest), RequestOptions.DEFAULT);
```
eshryane, line 143: You can still return from inside the try block, no need for a `response` variable (it's never used)

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
```diff
@@ -136,7 +138,16 @@ protected SearchResponse doSearch() throws IOException {
...
+        } catch (ElasticsearchStatusException ex){
+            if (ex.status().equals(RestStatus.BAD_REQUEST)){
+                throw new IllegalArgumentException("The query syntax doesn't fit ES standards, please see the documentation for more info");
```
eshryane, line 146: Simplify to `invalid query syntax`. 

Is there something in the exception `message` that could help the user? If so then append `ex.getMessage()`.

Also please log at `INFO` in an else block so we can tell how else the request could fail.

MiguelAHM: Simplify to invalid query syntax. -> OK

Is there something in the exception message that could help the user? If so then append ex.getMessage(). -> The message is not so pretty, I would prefer to link to our documentation instead of forward the ES message. However, I am ok putting that message in our logs

Also please log at INFO in an else block so we can tell how else the request could fail. -> Ok

eshryane: Thanks just log the ES message for now then, we can investigate further how to improve the message once we have more examples.

eshryane (approved):

PR #1268, opened 2023-06-28
RDAP: New exception for ASN errors

MiguelAHM:
Add flat conformance tag to aut-num errors


whois-api/src/main/java/net/ripe/db/whois/api/rdap/AutnumException.java
```diff
@@ -0,0 +1,28 @@
+package net.ripe.db.whois.api.rdap;
+
+public class AutnumException extends RuntimeException {
```
eshryane, line 3: Can you make `AutnumException` extend `RdapException` ?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapExceptionMapper.java
```diff
@@ -67,12 +73,14 @@ public Response toResponse(final Exception exception) {
 
     private Response createErrorResponse(final int status, final String errorTitle, final String ... errorMessage) {
         return Response.status(status)
-                .entity(createErrorEntity(status, errorTitle, errorMessage))
+                .entity(rdapObjectMapper.mapError(status, errorTitle, Lists.newArrayList(errorMessage)))
```
eshryane, line 76: Can you keep the `createErrorEntity` as it's more readable?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/AutnumException.java
MiguelAHM: I was thinking of doing that but the subclass will be the same as the super class, so it will just extend RdapException but it will be empty.
I will do that because for me it is more clear (AutnumException is an RdapException, they are related)

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -206,6 +206,12 @@ public RdapObject mapError(final int errorCode, final String errorTitle, final L
         return rdapObject;
     }
 
+    public RdapObject mapAutnumError(final int errorCode, final String errorTitle, final List<String> errorDescriptions){
+        final RdapObject rdapObject = mapError(errorCode, errorTitle, errorDescriptions);
+        rdapObject.getRdapConformance().add(RdapConformance.FLAT_MODEL.getValue());
+
```
maggarwal13, line 212: do we need to add. FLAT_MODEL conformance only incase of error?

maggarwal13: I mean when we create an Autnum response we are already adding RdapConformance.FLAT_MODEL.getValue(). If we add it before performing the query any exception will have this value in that case we do not need to create a autnumexception?

maggarwal13: I have checked we do not have that information available in mapper class , it creates an empty bject incase of error

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapExceptionMapper.java
```diff
@@ -42,6 +42,13 @@ public Response toResponse(final Exception exception) {
             return createErrorResponse(HttpServletResponse.SC_BAD_REQUEST, "400 Bad Request",
                     "unknown " + parameterName);
```
maggarwal13, line 43: Ideally if we are throwing a custom exception then we should catch all exception in service layer and turn it into an RdapException, I am not sure why we still have individual exception here but feel free to leave it as is it if it is a complicated or big change

whois-api/src/main/java/net/ripe/db/whois/api/rdap/WhoisRdapService.java
```diff
@@ -215,14 +215,18 @@ private Response lookupWithRedirectUrl(final HttpServletRequest request, final S
     }
 
     private Response lookupForAutNum(final HttpServletRequest request, final String key) {
-        if (isRedirect(AUT_NUM, key) && !rdapRequestValidator.isReservedAsNumber(key)) {
-            return redirect(getRequestPath(request), AUT_NUM, key);
-        }
+        try {
+            if (isRedirect(AUT_NUM, key) && !rdapRequestValidator.isReservedAsNumber(key)) {
+                return redirect(getRequestPath(request), AUT_NUM, key);
+            }
```
maggarwal13, line 221: Incase of a redirect do we need to throw autnum exception? It will add an flat model which does not make sense as it is redirect to some other url and we are not sure which hierarchy they implemented it

whois-api/src/main/java/net/ripe/db/whois/api/rdap/WhoisRdapService.java
```diff
@@ -215,14 +215,18 @@ private Response lookupWithRedirectUrl(final HttpServletRequest request, final S
...
 
-        final Query query = getQueryObject(ImmutableSet.of(AUT_NUM), key);
-        List<RpslObject> result = rdapQueryHandler.handleAutNumQuery(query, request);
+            final Query query = getQueryObject(ImmutableSet.of(AUT_NUM), key);
+            List<RpslObject> result = rdapQueryHandler.handleAutNumQuery(query, request);
 
-        return getResponse(request, result);
+            return getResponse(request, result);
+        } catch (RdapException ex){
+            throw new AutnumException(ex.getErrorTitle(), ex.getErrorDescription(), ex.getErrorCode());
```
maggarwal13, line 228: Are you sure onl Rdapexception are being thrown , I see mapping for lot other exception like paramException which are not being turned into RdapException so on those situations we are still not sending autnum exception? I think we should turn all exceptions into rdapExceptions to make it work properly

whois-api/src/main/java/net/ripe/db/whois/api/rdap/WhoisRdapService.java
MiguelAHM: either a 200 or a 404 we need to set that conformance:

We should receive "nro_rdap_profile_asn_flat_0" because it is undelegated and we are using flat model (just for 404 on ASN) 

To signify use of the flat model, the string literal 
   "nro_rdap_profile_asn_flat_0" MUST be included as an rdapConformance 
   value in each response.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/WhoisRdapService.java
MiguelAHM: paramException is thrown by jetty I think.
We don't have control on all the exception that we can thrown

whois-api/src/main/java/net/ripe/db/whois/api/rdap/WhoisRdapService.java
MiguelAHM: lookup_autnum_not_found_still_flat_conformance -> 404 with the conformance (for Autum)
lookup_org_error_correct_conformance -> 404 without conformance (without Autnum)

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapExceptionMapper.java
MiguelAHM: discussed via Mattermost
There are some exception that we can not control or take over with RdapException

eshryane (approved):

PR #1270, opened 2023-06-29
RDAP: Vcard in nested objects including kind

MiguelAHM:


whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -466,21 +467,31 @@ private static List<Entity> createContactEntities(final RpslObject rpslObject) {
         for (final Map.Entry<CIString, Set<AttributeType>> entry : contacts.entrySet()) {
             final Entity entity = new Entity();
             entity.setHandle(entry.getKey().toString());
+            final Set<ObjectType> objectPossibleTypes = new HashSet<>();
```
eshryane, line 470: Use `Sets.newHashSet()` for consistency

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -466,21 +467,31 @@ private static List<Entity> createContactEntities(final RpslObject rpslObject) {
...
             for (final AttributeType attributeType : entry.getValue()) {
+                objectPossibleTypes.addAll(attributeType.getReferences());
```
eshryane, line 472: Move this logic into a separate method? e.g. `@Nullable RpslObject lookup(final ObjectType objectType, final CIString primaryKey)`

Is is possible that there can ever be more than 1 reference?

MiguelAHM: Done the refactor

eshryane (approved):

PR #1273, opened 2023-06-30
RDAP: Document delegated placeholders problems and add some IT to cover this use case

MiguelAHM:
Review task last comment of DB-4740 for more info

MariaDB [INTERNALS_RIPE]> SELECT resource FROM authoritative_resource WHERE resource like '0%';
Empty set (0.003 sec)

We are not taking in account 0/0 or ::0/0
So there is confusing code.

If we query by this we will get a 404 because it will perform a redirection


README.RDAP.md
```diff
@@ -61,3 +61,10 @@ The RIPE database doesn't contain any forward domain objects, consequently a nam
 Only "mnt-by:" Maintainers are Listed as Registrants
 -----------------------------------------------------
 Only maintainers referenced in "mnt-by:" attributes will be listed as Registrants in responses.
+
+IANA delegated objects will return a 404
```
MiguelAHM: Objects with "administrative" status are not returned

README.RDAP.md
```diff
@@ -61,3 +61,10 @@ The RIPE database doesn't contain any forward domain objects, consequently a nam
...
+-----------------------------------------------------
+There must be a placeholder with "administrative" status for the IANA-delegated object.
```
MiguelAHM: If the prefix is either delegated but unallocated or only partially delegated to the RIPE region, then a 404 is returned. An object with "administrative" status is never returned.

README.RDAP.md
```diff
@@ -61,3 +61,10 @@ The RIPE database doesn't contain any forward domain objects, consequently a nam
...
+
+Since there are no IANA-delegated files in the RIPE database, a Not Found response will be given.
```
MiguelAHM: Refer to [NRO RDAP Profile](https://bitbucket.org/nroecg/nro-rdap-profile/raw/v1/nro-rdap-profile.txt) section 4.5.  "Status"

eshryane (approved):

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapRedirectTestIntegration.java
```diff
@@ -185,6 +190,96 @@ public void inetnum_outside_range() {
...
+                    ".net/ip/217.180.0.0/16"));
+        }
+    }
+    
+    @Test
+    public void inetnum_inside_range_not_found() {
+        try {
+            RestTest.target(getPort(), String.format("rdap/%s", "ip/192.0.0.1"))
+                .request(MediaType.APPLICATION_JSON_TYPE)
+                .get(String.class);
```
eshryane, line 274: Thanks for fixing the test! Can you please add a `fail()` here also in case no exception is thrown

MiguelAHM: Done

PR #1278, opened 2023-07-10
RDAP: Add remarks to remark property of RDAP

MiguelAHM:
Apart from description we need to add the remarks to remarks attribute in RDAP

### DIFF
--- a/whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/Remark.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/Remark.java
@@ -16,8 +16,11 @@ import java.util.List;
 public class Remark implements Serializable {
     protected List<String> description;
 
-    public Remark(final List<String> description) {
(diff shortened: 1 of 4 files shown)


whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/Remark.java
```diff
@@ -16,8 +16,11 @@
 public class Remark implements Serializable {
     protected List<String> description;
 
-    public Remark(final List<String> description) {
+    protected List<String> remarks;
```
eshryane, line 19: Don't add remarks under remarks, that's not a valid RDAP response (see RFC 9083). Any "remarks:" attribute values need to be added as additional *description* elements.

MiguelAHM: Done

eshryane (approved):

PR #1280, opened 2023-07-11
Fix Domain tests

MiguelAHM:
- The DOMAIN objects still are working with IP, so could be the case that we try to search for the parent of 0/0 in that case we will thrown a 500 and we should just skip the parent of that 0/0.

So we need to keep the check of the parent

- Move the test "inetnum_inside_range_not_found" to the correct Integration test class
- Perform the next code in the RdapMapper instead of the service to avoid duplicated code:
        if (isIANABlock(rpslObject)){
            throw new RdapException("404 Not Found", "Requested object not found", HttpStatus.NOT_FOUND_404);
        }
- Finally "lookup_inet6num_not_found" fix this, we need to return a 404 and it was returning a 200

Test ok:
https://gitlab.ripe.net/swe-database-team/whois/-/pipelines/371714


eshryane (dismissed):

whois-api/src/main/java/net/ripe/db/whois/api/rdap/WhoisRdapService.java
```diff
@@ -370,10 +370,6 @@ private Response getResponse(HttpServletRequest request, Iterable<RpslObject> re
 
         final RpslObject resultObject = rpslIterator.next();
 
-        if (isIANABlock(resultObject)){
```
eshryane, line 373: Why move the check from the service class? Are there object lookups done elsewhere (I can't find the duplication)

eshryane (commented):
Why move the throws into the map() method

whois-api/src/main/java/net/ripe/db/whois/api/rdap/WhoisRdapService.java
MiguelAHM: Hi!
I didn't want to duplicate isIANABlock method in both
I updated setting the method as public and using it from the service instead of throwing the error in the mapping

maggarwal13: Just a suggestion, make this a util function ? We need to use this in lot of places. Feel free to ignore as it was like this earlier as well

PR #1290, opened 2023-08-14
Add env check

MiguelAHM:
After talking with Ed we will skip the idea of passing the ENV and trying to figure out the ENV using Global variables or some object from database that only exist in PROD

### DIFF
--- a/whois-commons/src/main/java/net/ripe/db/whois/common/DatabaseDummifierJmx.java
+++ b/whois-commons/src/main/java/net/ripe/db/whois/common/DatabaseDummifierJmx.java
@@ -81,10 +81,12 @@ public class DatabaseDummifierJmx extends JmxBase {
         return invokeOperation("dummify", jdbcUrl, new Callable<String>() {
             @Override
             public String call() {
-                validateEnvironment(env);
                 final SimpleDataSourceFactory simpleDataSourceFactory = new SimpleDataSourceFactory("org.mariadb.jdbc.Driver");
                 final DataSource dataSource = simpleDataSourceFactory.createDataSource(jdbcUrl, user, pass);
                 jdbcTemplate = new JdbcTemplate(dataSource);
+
+                final String dbEnvironment = jdbcTemplate.queryForObject("SELECT name FROM environment LIMIT 1", String.class);
+                validateEnvironment(env, dbEnvironment);
 
                 final DataSourceTransactionManager transactionManager = new DataSourceTransactionManager(dataSource);
                 transactionTemplate = new TransactionTemplate(transactionManager);
@@ -120,7 +122,13 @@ public class DatabaseDummifierJmx extends JmxBase {
         });
     }
 
-    private void validateEnvironment(final EnvironmentEnum env) {
+    private void validateEnvironment(final EnvironmentEnum env, final String dbEnvironment) {
+        if (dbEnvironment == null){
+            throw new IllegalStateException("Environment no specified  in the schema");
+        }
+        if (!env.name().equalsIgnoreCase(dbEnvironment)){
+            throw new IllegalArgumentException("Requested environment and database environment doesn't match");
+        }
         if (EnvironmentEnum.PROD.equals(env)) {
             throw new IllegalArgumentException("dummifier runs on non-production environments only");
         }
--- a/whois-commons/src/main/java/net/ripe/db/whois/common/DatabaseDummifierJmx.java
+++ b/whois-commons/src/main/java/net/ripe/db/whois/common/DatabaseDummifierJmx.java
@@ -84,9 +84,7 @@ public class DatabaseDummifierJmx extends JmxBase {
                 final SimpleDataSourceFactory simpleDataSourceFactory = new SimpleDataSourceFactory("org.mariadb.jdbc.Driver");
                 final DataSource dataSource = simpleDataSourceFactory.createDataSource(jdbcUrl, user, pass);
                 jdbcTemplate = new JdbcTemplate(dataSource);
-
-                final String dbEnvironment = jdbcTemplate.queryForObject("SELECT name FROM environment LIMIT 1", String.class);
-                validateEnvironment(env, dbEnvironment);
+                validateEnvironment(env);
 
                 final DataSourceTransactionManager transactionManager = new DataSourceTransactionManager(dataSource);
                 transactionTemplate = new TransactionTemplate(transactionManager);
@@ -122,9 +120,10 @@ public class DatabaseDummifierJmx extends JmxBase {
         });
     }
 
-    private void validateEnvironment(final EnvironmentEnum env, final String dbEnvironment) {
+    private void validateEnvironment(final EnvironmentEnum env) {
+        final String dbEnvironment = jdbcTemplate.queryForObject("SELECT name FROM environment LIMIT 1", String.class);
         if (dbEnvironment == null){
-            throw new IllegalStateException("Environment no specified  in the schema");
+            throw new IllegalStateException("Environment not specified in the schema");
         }
         if (!env.name().equalsIgnoreCase(dbEnvironment)){
             throw new IllegalArgumentException("Requested environment and database environment doesn't match");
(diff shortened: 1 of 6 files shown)


whois-commons/src/main/java/net/ripe/db/whois/common/DatabaseDummifierJmx.java
```diff
@@ -64,6 +68,9 @@ public class DatabaseDummifierJmx extends JmxBase {
     private static final AtomicInteger jobsAdded = new AtomicInteger();
     private static final AtomicInteger jobsDone = new AtomicInteger();
 
+    private static final Set<String> AVAILABLE_ENVIRONMENTS = ImmutableSet.of("dev", "prepdev", "test", "rc",
+            "training");
+
```
maggarwal13, line 73: If you are introucing Environment , can you create an enum for it?

whois-commons/src/main/java/net/ripe/db/whois/common/DatabaseDummifierJmx.java
```diff
@@ -117,16 +125,12 @@ public String call() {
         });
     }
 
-    private void validateJdbcUrl(final String user, final String password) {
-        if (!user.equals(password) && !user.equals(reverse(password))) {
-            throw new IllegalArgumentException("dummifier runs on non-production environments only (user==password)");
+    private void validateEnvironment(final String env) {
+        if (!AVAILABLE_ENVIRONMENTS.contains(env)) {
+            throw new IllegalArgumentException("dummifier runs on non-production environments only");
```
maggarwal13, line 130: if it will throw error only in production , can you flip the condition, if it is production throw error?

whois-commons/src/main/java/net/ripe/db/whois/common/DatabaseDummifierJmx.java
maggarwal13: There is also sentry.environment , can you replace it with this new property?

whois-commons/src/main/java/net/ripe/db/whois/common/DatabaseDummifierJmx.java
```diff
@@ -32,6 +33,7 @@
 import java.sql.ResultSet;
 import java.sql.SQLException;
```
maggarwal13, line 34: Can you change the comment on top of class as well

whois-commons/src/main/java/net/ripe/db/whois/common/DatabaseDummifierJmx.java
MiguelAHM: If I check that the user is just adding PROD is less secure that if I check that the user is specifying exactly the correct env.
For example: env=pro -> This will give True if we are just checking env.equals(prod)

maggarwal13: Can you introduce an enum for environment? And use enum equals? You need to first convert the user input to enum.

MiguelAHM: Ok, good idea, this should work yes.

whois-commons/src/main/java/net/ripe/db/whois/common/DatabaseDummifierJmx.java
MiguelAHM: as we talked in standup this is a jmx script, we are specifying the properties in the input. We are not taking the properties from the whois property file. Look the database info input.

whois-commons/src/main/java/net/ripe/db/whois/common/DatabaseDummifierJmx.java
MiguelAHM: Done

whois-commons/src/main/java/net/ripe/db/whois/common/DatabaseDummifierJmx.java
MiguelAHM: Done

whois-commons/src/main/java/net/ripe/db/whois/common/DatabaseDummifierJmx.java
```diff
@@ -242,6 +252,14 @@ public static void main(String[] argv) {
         String jdbcUrl = options.valueOf(ARG_JDBCURL).toString();
         String user = options.valueOf(ARG_USER).toString();
         String pass = options.valueOf(ARG_PASS).toString();
-        new DatabaseDummifierJmx().dummify(jdbcUrl, user, pass);
+        Environment env;
+        try {
+            env = Environment.valueOf(options.valueOf(ARG_ENV).toString().toUpperCase());
+        } catch (IllegalArgumentException ex){
```
maggarwal13, line 258: you can also use EnumUtils.getEnumIgnoreCase, feel free to ignore

whois-commons/src/main/java/net/ripe/db/whois/common/DatabaseDummifierJmx.java
```diff
@@ -64,6 +66,16 @@ public class DatabaseDummifierJmx extends JmxBase {
     private static final AtomicInteger jobsAdded = new AtomicInteger();
     private static final AtomicInteger jobsDone = new AtomicInteger();
 
+    private enum Environment {
+        DEV,
+        PREPDEV,
```
maggarwal13, line 71: We use environment property in sentry too, can you shift to a separate  Enum class so we can use this enum there as well?

whois-commons/src/main/java/net/ripe/db/whois/common/DatabaseDummifierJmx.java
```diff
@@ -72,13 +84,14 @@ public DatabaseDummifierJmx() {
     @ManagedOperationParameters({
             @ManagedOperationParameter(name = "jdbcUrl", description = "jdbc url"),
             @ManagedOperationParameter(name = "user", description = "jdbc username"),
-            @ManagedOperationParameter(name = "pass", description = "jdbc password")
```
maggarwal13, line 75: Can you change the comment too to include the environment? /**
 * in jmxterm, run with:
 *      run dummify jdbc:mariadb://<host>/<db> <user> <pass>
 *
 * in console, run with
 *      java -Xmx1G -cp whois.jar net.ripe.db.whois.common.DatabaseDummifierJmx --jdbc-url jdbc:mariadb://localhost/BLAH --user XXX --pass XXX
 *
 */

whois-commons/src/main/java/net/ripe/db/whois/common/DatabaseDummifierJmx.java
MiguelAHM: I can see that method in EnumUtils :S

whois-commons/src/main/java/net/ripe/db/whois/common/DatabaseDummifierJmx.java
MiguelAHM: Done

whois-commons/src/main/java/net/ripe/db/whois/common/DatabaseDummifierJmx.java
MiguelAHM: Done

MiguelAHM:
> Isn't it still possible to copy/paste production credentials + URL but use a non-production "ENV"? Is there something "special" about the production environment that we can check instead?

as we talked, we will create a database with the env and check that database connection with the provided information

whois-commons/src/main/java/net/ripe/db/whois/common/DatabaseDummifierJmx.java
```diff
@@ -242,6 +249,14 @@ public static void main(String[] argv) {
         String jdbcUrl = options.valueOf(ARG_JDBCURL).toString();
         String user = options.valueOf(ARG_USER).toString();
         String pass = options.valueOf(ARG_PASS).toString();
-        new DatabaseDummifierJmx().dummify(jdbcUrl, user, pass);
+        EnvironmentEnum env;
```
MiguelAHM, line 252: Use final keyword

whois-commons/src/main/java/net/ripe/db/whois/common/EnvironmentEnum.java
```diff
@@ -0,0 +1,10 @@
+package net.ripe.db.whois.common;
+
+public enum EnvironmentEnum {
```
MiguelAHM, line 3: Remove Enum from class name

eshryane (approved):
Approve pending comments

PR #1299, opened 2023-08-24
Escape ES colons

MiguelAHM:
Escape colon for query input


whois-rpsl/src/main/java/net/ripe/db/whois/common/ip/Ipv6Resource.java
```diff
@@ -33,6 +39,18 @@ public class Ipv6Resource extends IpInterval<Ipv6Resource> implements Comparable
     private final long endMsb;
     private final long endLsb;
 
+    private static final Pattern IPV6_SPLIT_PATTERN = Pattern.compile("(" +
+            "[0-9A-Fa-f]{1,4}(?::[0-9A-Fa-f]{1,4}){7}" +
+            "|::" +
+            "|:(?::[0-9A-Fa-f]{1,4}){1,6}" +
```
maggarwal13, line 45: This logic should not be in  this class. This is strictly for IPv6Resource from a string or integer.

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
```diff
@@ -151,6 +152,27 @@ private org.elasticsearch.action.search.SearchResponse performFulltextSearch(fin
         }
     }
 
+    private SearchRequest escapeIPv6Colon(final SearchRequest searchRequest){
+        final List<String> ipv6Matches = Ipv6Resource.extractIPv6(searchRequest.getQuery());
+        String originalQuery = searchRequest.getQuery();
```
maggarwal13, line 157: Can you revert the condition , if emprty return original request?

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
```diff
@@ -151,6 +152,27 @@ private org.elasticsearch.action.search.SearchResponse performFulltextSearch(fin
...
+        if (!ipv6Matches.isEmpty()){
+            for (String ipv6Match:ipv6Matches) {
+                originalQuery = originalQuery.replace(ipv6Match, ipv6Match.replace(":", "\\:"));
+            }
+            return new SearchRequest.SearchRequestBuilder()
+                    .setRows(String.valueOf(searchRequest.getRows()))
+                    .setStart(String.valueOf(searchRequest.getStart()))
+                    .setQuery(originalQuery)
```
maggarwal13, line 165: Why are you making a new search request? Can you do it in service class before setting a query

whois-rpsl/src/main/java/net/ripe/db/whois/common/ip/Ipv6Resource.java
```diff
@@ -33,6 +39,18 @@ public class Ipv6Resource extends IpInterval<Ipv6Resource> implements Comparable
...
+            "|[0-9A-Fa-f]{1,4}:(?::[0-9A-Fa-f]{1,4}){1,5}" +
+            "|(?:[0-9A-Fa-f]{1,4}:){2}(?::[0-9A-Fa-f]{1,4}){1,4}" +
+            "|(?:[0-9A-Fa-f]{1,4}:){3}(?::[0-9A-Fa-f]{1,4}){1,3}" +
+            "|(?:[0-9A-Fa-f]{1,4}:){4}(?::[0-9A-Fa-f]{1,4}){1,2}" +
+            "|(?:[0-9A-Fa-f]{1,4}:){5}:[0-9A-Fa-f]{1,4}" +
+            "|(?:[0-9A-Fa-f]{1,4}:){1,6}:" +
```
maggarwal13, line 51: This is really complicated to understand. Is this the only way?

MiguelAHM: I know: not I think that this is trying to check the correct syntax of the IPv6 as well, we don't need that. I am trying to create a new regex, but it is a bit complicated

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
MiguelAHM: I am doing it because there is only a builder that allow change the SearchRequest.
I could see any setters. I don't want to overload the service class I think that this logic should be in a lower layer (service just to throw exception that makes not sense to pass to a lower layer).
I could do that or create a setter and set the value to the current searchRequest

whois-rpsl/src/main/java/net/ripe/db/whois/common/ip/Ipv6Resource.java
MiguelAHM: Created new class

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFulltextSearch.java
MiguelAHM: Done

whois-rpsl/src/main/java/net/ripe/db/whois/common/ip/Ipv6Resource.java
MiguelAHM: Please let me know if we can make this regex easier to understand

whois-rpsl/src/main/java/net/ripe/db/whois/common/ip/IpExtractor.java
```diff
@@ -0,0 +1,27 @@
...
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class IpExtractor {
+
+    private static final Pattern IPV6_SPLIT_PATTERN = Pattern.compile("" +
+            "(" +
+            "[\\dA-Fa-f]{1,4}(?::[\\dA-Fa-f]{1,4}){7}" + // basic case when we get full IPv6(2001:0000:130F:0000:0000:09C0:876A:130B)
+            "|(?:[\\dA-Fa-f]{1,4}:){1,6}(?::[\\dA-Fa-f]{1,4}){1,6}" + // case when there is a double colon in the IPv6(2a00:1f78::fffe)
```
maggarwal13, line 13: What happens if user just send 2001: ? Colon can appear in descr or remarks as well which may or may not be ipv6

MiguelAHM: we need to escape de colon anyways, right? if not we will throw an error

whois-rpsl/src/main/java/net/ripe/db/whois/common/ip/IpExtractor.java
```diff
@@ -0,0 +1,27 @@
+package net.ripe.db.whois.common.ip;
+
+import java.util.HashSet;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class IpExtractor {
+
+    private static final Pattern IPV6_SPLIT_PATTERN = Pattern.compile("" +
```
eshryane, line 10: Why not match colons anywhere in the string and iterate?

MiguelAHM:
Test ok -> https://gitlab.ripe.net/swe-database-team/whois/-/pipelines/399999

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/FullTextSearchService.java
```diff
@@ -65,6 +68,36 @@ public Response search(
         }
     }
 
+    private String escapeColon(final String query){
+        if (query==null || !query.contains(":")){
+            return query;
```
maggarwal13, line 73: Use stringUtils.isEmpty ?

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/FullTextSearchService.java
```diff
@@ -65,6 +68,36 @@ public Response search(
...
+        }
+
+        final StringBuilder sb = new StringBuilder();
+
+        final Pattern colonRegex = Pattern.compile("[^:\\n]*((:){1,2}|$)"); //take everything that finish with :
+        final Matcher wordsWithTrailingColon = colonRegex.matcher(query);
```
maggarwal13, line 79: Just split by colon ? Would be much more readable and easier ?

MiguelAHM: That was my first attempt. If you split by colon you lose a lot of information. You can have as well ::/64 for example, if you split that you will have ["", "",] you don't know what to do.

Another example 2a00:1f78::fffe/48 -> [2a00, 1f78, "", fffe]
with this information you can do anything.

So the idea here is to get everything until the colon (1 or 2 and with them included), so you can figure out if that colon is already escaped or not and then you can.

On other hand, we are building another query, not changed the one that we already have...if you split by colon you will lose the original piece of the query.

maggarwal13: Talked separately to simplify this using split

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/FullTextSearchService.java
```diff
@@ -65,6 +68,23 @@ public Response search(
...
+
+        final String[] split = query.split(":");
+        Arrays.stream(split).limit(split.length - 1).forEach(splitColon -> sb.append(splitColon).append(shouldEscapeColon(splitColon) ? "\\:" : ":"));
+        return sb.append(split[split.length - 1]).toString();
+    }
+
+    private static boolean shouldEscapeColon(final String splitColon) {
+        final String[] words = splitColon.split("[^\\w-]+"); //Split in words (including dashes)
+        return words.length == 0 || (AttributeType.getByNameOrNull(words[words.length - 1]) == null && !"object-type" .equals(words[words.length - 1]));
+    }
```
maggarwal13, line 86: Use OBJECT_TYPE_FIELD_NAME ? Also since AttributeType using ciString equals may be use equalsIgnoreIcase for object-type also or convert it to cistring? Sorry did not check while refactoring

MiguelAHM: Done

whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/FullTextSearchService.java
```diff
@@ -65,6 +70,23 @@ public Response search(
...
+        }
+
+        final StringBuilder sb = new StringBuilder();
+
+        final String[] split = query.split(":");
+        Arrays.stream(split).limit(split.length - 1).forEach(splitColon -> sb.append(splitColon).append(shouldEscapeColon(splitColon) ? "\\:" : ":"));
+        return sb.append(split[split.length - 1]).toString();
+    }
+
+    private static boolean shouldEscapeColon(final String splitColon) {
```
maggarwal13, line 85: Since we are using the same api for public queries too, can we also  test the integration with db-web-ui too ? It should not impact that.

MiguelAHM:
test ok 
https://gitlab.ripe.net/swe-database-team/whois/-/pipelines/400621

maggarwal13 (approved):

PR #1300, opened 2023-08-25
Remove role dummified object

MiguelAHM:
Nothing is referencing the dummy role object, so it is safe not to return it. We are referencing the dummy person object in all the situations.

For the exporter the dummy role is never returned because it is never use (previously we included a dummy role but the Nic handler conflicted with a real object, that is why we removed it)


eshryane (approved):

MiguelAHM:
https://gitlab.ripe.net/swe-database-team/whois/-/pipelines/399212
test ok

PR #1302, opened 2023-08-31
Use base64 instead hex

MiguelAHM:
We are using Hex when we decode/encode the signature.

We should use base64.


MiguelAHM:
Test -> https://gitlab.ripe.net/swe-database-team/whois/-/pipelines/400111

eshryane (approved):

PR #1303, opened 2023-09-04
Move the Environment table to internals database

MiguelAHM:
Due to weekly reload process we need to move the table to a different database.
Because it is related to configuration information we decided to move into internals database.

### DIFF
--- a/whois-commons/src/main/java/net/ripe/db/whois/common/DatabaseDummifierJmx.java
+++ b/whois-commons/src/main/java/net/ripe/db/whois/common/DatabaseDummifierJmx.java
@@ -61,6 +61,8 @@ public class DatabaseDummifierJmx extends JmxBase {
     private static TransactionTemplate transactionTemplate;
     private static JdbcTemplate jdbcTemplate;
 
+    private static JdbcTemplate internalTemplate;
+
     private static final DummifierRC dummifier = new DummifierRC();
 
     private static final AtomicInteger jobsAdded = new AtomicInteger();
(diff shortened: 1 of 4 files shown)


MiguelAHM:
test ok -> https://gitlab.ripe.net/swe-database-team/whois/-/pipelines/400621

whois-commons/src/main/java/net/ripe/db/whois/common/DatabaseDummifierJmx.java
```diff
@@ -84,6 +86,10 @@ public String call() {
                 final SimpleDataSourceFactory simpleDataSourceFactory = new SimpleDataSourceFactory("org.mariadb.jdbc.Driver");
                 final DataSource dataSource = simpleDataSourceFactory.createDataSource(jdbcUrl, user, pass);
                 jdbcTemplate = new JdbcTemplate(dataSource);
+
+                final DataSource internalSource = simpleDataSourceFactory.createDataSource("jdbc:mariadb://localhost/INTERNALS_RIPE", user, pass);
```
eshryane, line 90: Is host always `localhost`? Or do we need to adjust the JDBC URL ?

MiguelAHM: yes, unless that we decide to run this script in Sussex for example and to use the Internals of Dexter for example (in that case it would be dexter.ripe.net instead of localhost). But II don't know why we would do that. Even that we will run this script always in master database (we are updating the database) and master will always have that database.

I can request the internal database as parameter as we do with whois_update database.

eshryane: No it's fine! Thanks for confirming

eshryane:
Improved the PR title

eshryane (approved):

PR #1309, opened 2023-09-12
Try parallel stream to generate NRTMv4 snapshots

maggarwal13:
Use parallel stream instead of queue for generating snapshot files.
Generation took 12 min in dev environment


whois-nrtm4/src/main/java/net/ripe/db/nrtm4/generator/SnapshotFileGenerator.java
```diff
@@ -178,4 +194,43 @@ private void cleanUpOldFiles() {
...
+            }
+        });
+    }
+
+    private void printProgress(final AtomicInteger noOfBatchesProcessed, int total, Timer timer) {
+        timer.schedule(new TimerTask() {
+            @Override
+            public void run() {
+                final int done = noOfBatchesProcessed.get();
+                LOGGER.info("Processed {} Batches out of {} ({}%).", done, total, (done * 100/ total));
```
eshryane, line 232: Can you log *objects* rather than batches (i.e. multiply count by BATCH_SIZE)

eshryane (approved):

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/GzipOutStreamWriter.java
```diff
@@ -16,30 +16,40 @@ public class GzipOutStreamWriter {
...
     }
 
     public void close() {
-        LOGGER.info("closing resources");
+        LOGGER.debug("closing resources");
         try {
             gzOut.close();
         } catch (IOException e) {
-            LOGGER.error("Exception while closing gzipStream {}", e);
+            LOGGER.debug("Exception while closing gzipStream {}", e);
```
MiguelAHM, line 47: This should be logged as Error, not as debug, right?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/GzipOutStreamWriter.java
```diff
@@ -16,30 +16,40 @@ public class GzipOutStreamWriter {
...
         }
         try {
             bos.close();
         } catch (IOException e) {
-            LOGGER.error("Exception while closing outputstream {}", e);
+            LOGGER.debug("Exception while closing outputstream {}", e);
```
MiguelAHM, line 52: This should be logged as Error, not as debug, right?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/generator/SnapshotFileGenerator.java
```diff
@@ -82,47 +85,59 @@ public void createSnapshot()  {
...
+            }).flatMap(Collection::stream).forEach(rpslObject -> {
+                if(sourceResources.containsKey(rpslObject.getValueForAttribute(AttributeType.SOURCE))) {
+                    sourceResources.get(rpslObject.getValueForAttribute(AttributeType.SOURCE)).write(new SnapshotFileRecord(rpslObject));
                 }
-            }
-        });
+            });
+
+        } catch (final Exception e) {
+            LOGGER.warn("Error while writing snapshotfile", e);
```
MiguelAHM, line 133: why is this a warning and not an error?

maggarwal13: Runtime exception is thrown after that which will anyways be reported and logged. Will change to error anyways

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/GzipOutStreamWriter.java
maggarwal13: So that when it tries to close already closed stream it will not log it as error . Also you cannot read the bytes from open output stream so it will throw a runtime error which will be logged

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/GzipOutStreamWriter.java
maggarwal13: Same as below

MiguelAHM (approved):

PR #1313, opened 2023-09-20
Fix nrtm transactions

MiguelAHM:
NRTMv4 is using a different transaction manager because the datasource is different (it's NRTM-specific).

The default transaction manager is defined in whois-commons and uses the source aware datasource.

The `@Transactional` annotations in the NRTMv4 module specifies the NRTM transaction manager.

### DIFF
--- a/whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/NrtmFileRepository.java
+++ b/whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/NrtmFileRepository.java
@@ -8,6 +8,7 @@ import net.ripe.db.nrtm4.domain.NrtmSource;
 import net.ripe.db.nrtm4.domain.NrtmVersionInfo;
 import net.ripe.db.nrtm4.domain.NrtmVersionRecord;
 import net.ripe.db.nrtm4.domain.SnapshotFile;
+import net.ripe.db.nrtm4.source.NrtmSourceContext;
 import net.ripe.db.nrtm4.util.NrtmFileUtil;
 import net.ripe.db.whois.common.DateTimeProvider;
 import net.ripe.db.whois.common.dao.jdbc.JdbcRpslObjectOperations;
@@ -37,7 +38,9 @@ public class NrtmFileRepository {
     private final DateTimeProvider dateTimeProvider;
     private final NamedParameterJdbcTemplate namedParameterJdbcTemplate;
 
-    public NrtmFileRepository(@Qualifier("nrtmSourceAwareDataSource") final DataSource dataSource, final DateTimeProvider dateTimeProvider) {
+
+    public NrtmFileRepository(@Qualifier("nrtmSourceAwareDataSource") final DataSource dataSource,
+                              final DateTimeProvider dateTimeProvider) {
         this.jdbcTemplate = new JdbcTemplate(dataSource);
         this.dateTimeProvider = dateTimeProvider;
         this.namedParameterJdbcTemplate = new NamedParameterJdbcTemplate(jdbcTemplate);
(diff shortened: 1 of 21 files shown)


eshryane (approved):

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/generator/SnapshotFileGenerator.java
```diff
@@ -91,10 +96,16 @@ public void createSnapshot()  {
...
+        try {
+            nrtmSourceContext.setCurrentSourceToWhoisMaster();
+
+            saveToDatabase(sourceToNewVersion, sourceToOutputBytes);
+            LOGGER.info("Snapshot generation complete {}", stopwatch);
 
-        cleanUpOldFiles();
+            cleanUpOldFiles();
+        } finally {
+            nrtmSourceContext.setCurrentSourceToWhoisSlave();
```
maggarwal13, line 107: In my opinion, we should not use sourceaware cotext for DB write, we should always specify write only datasource instead of changing it to master and later changing to slave as you ca  never use slave source to write to DB. May be there are other settings which made yo use sourceaware?

MiguelAHM: sourceAware is doing on this way
Just copy paste the behaviour. Check singleUpdates class

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/NrtmFileRepository.java
```diff
@@ -36,32 +36,29 @@ public class NrtmFileRepository {
...
+
+    public NrtmFileRepository(@Qualifier("nrtmSourceAwareDataSource") final DataSource dataSource, final DateTimeProvider dateTimeProvider) {
         this.jdbcTemplate = new JdbcTemplate(dataSource);
         this.dateTimeProvider = dateTimeProvider;
         this.namedParameterJdbcTemplate = new NamedParameterJdbcTemplate(jdbcTemplate);
     }
 
-    @Transactional
+    @Transactional("nrtmTransactionManager")
     public void saveDeltaVersion(final NrtmVersionInfo version, final int serialIDTo, final List<DeltaFileRecord> deltas) throws JsonProcessingException {
```
maggarwal13, line 47: <bean id="nrtmTransactionManager" class="org.springframework.jdbc.datasource.DataSourceTransactionManager">
        <property name="dataSource" ref="nrtmDataSource"/>
    </bean> 

NrtmTransaction uses nrtmDataSource (which is nrtm master datasource , we should rename it to be consistent with whois naming convention ) why we need to use sourceAware for trasactions? Is it being overriden?

MiguelAHM: we don't need it, but Ed told me to keep there for being sure that it is using that transaction

MiguelAHM: Using just master datasources and change the name to the source for nrtmMasterDataSource

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/UpdateNrtmFileRepository.java
```diff
@@ -29,34 +29,35 @@
...
+
+    public UpdateNrtmFileRepository(@Qualifier("nrtmMasterDataSource") final DataSource dataSource, final DateTimeProvider dateTimeProvider) {
         this.jdbcTemplate = new JdbcTemplate(dataSource);
         this.dateTimeProvider = dateTimeProvider;
         this.namedParameterJdbcTemplate = new NamedParameterJdbcTemplate(jdbcTemplate);
     }
 
-    @Transactional
+    @Transactional("nrtmTransactionManager")
     public void saveDeltaVersion(final NrtmVersionInfo version, final int serialIDTo, final List<DeltaFileRecord> deltas) throws JsonProcessingException {
```
maggarwal13, line 47: Is it possible to put the transactional at class level?

eshryane (approved):

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/UpdateNrtmFileRepository.java
```diff
@@ -29,34 +29,35 @@
...
-       if(deltas.isEmpty()) {
-           LOGGER.info("No delta changes found for source {}", version.source().getName());
-           return;
-       }
+        if (deltas.isEmpty()) {
+            LOGGER.info("No delta changes found for source {}", version.source().getName());
+            return;
+        }
```
maggarwal13, line 52: Why there is no transactional annotation on delete?

MiguelAHM: oh my fault, it should have that annotation as well. I will add it, thanks!

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/UpdateNrtmFileRepository.java
maggarwal13: Try @Transactional(transactionManager = "nrtmTransactionManager")

maggarwal13 (approved):

PR #1314, opened 2023-09-26
Check ES docs count vs Database object count

maggarwal13:
If the ES docs count does not match with database count log it as error


whois-commons/src/main/java/net/ripe/db/whois/common/dao/SerialDao.java
```diff
@@ -13,4 +15,6 @@ public interface SerialDao {
     SerialEntry getByIdForNrtm(int serialId);
 
     Integer getAgeOfExactOrNextExistingSerial(int serialId);
+
+    Map<Integer, Integer> getMaxSerialIdWithObjectCount();
```
MiguelAHM, line 19: Ignore the comment if you prefer: maybe it is more clear to have two different calls. For me a Map is when the first attribute of the map has something to do with the second one. For example in my task the Role depends on the AttributeType, so my map is mapping the Role with the AttributeType.
However, I am on with it, I understand perfectly the logic. I will approve

MiguelAHM (approved):

whois-commons/src/main/java/net/ripe/db/whois/common/dao/SerialDao.java
maggarwal13: Different calls not possible as there is a transactional annotation required to fetch this. Else we need to have complicated logic to get number of object count for a serial id

MiguelAHM: why are we using a transaction here if we are just reading from database? we are not writing anything.

MiguelAHM: ah ok, got it...I was thinking that maybe we had that transaction because of some type of isolation level

PR #1317, opened 2023-10-03
Rdap redaction

MiguelAHM:


whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -624,39 +635,35 @@ private Domain createDomain(final RpslObject rpslObject, final String requestUrl
...
+            case IRT -> builder.addFn(rpslObject.getValueForAttribute(IRT)).addKind(GROUP);
+            default -> {
+            }
         }
-        builder.addAdr(rpslObject.getValuesForAttribute(ADDRESS))
-                .addTel(rpslObject.getValuesForAttribute(PHONE))
+        builder.addTel(rpslObject.getValuesForAttribute(PHONE))
                 .addFax(rpslObject.getValuesForAttribute(FAX_NO))
-                .addEmail(rpslObject.getValuesForAttribute(E_MAIL))
+                .addAdr(rpslObject.getValuesForAttribute(ADDRESS))
```
eshryane, line 662: Can you remove email in a separate PR? It's not related to the redacted feature

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
```diff
@@ -0,0 +1,101 @@
...
+@Component
+public class RedactionObjectMapper {
+
+    private final RpslObjectDao rpslObjectDao;
+    private static final Map<AttributeType, Redaction> UNSUPPORTED_ENTITIES = Maps.newHashMap();
+    static {
+        UNSUPPORTED_ENTITIES.put(MBRS_BY_REF, new Redaction(new Redaction.Description("Indirect population of a set"), new Redaction.Description("No registrant mntner")));
+        UNSUPPORTED_ENTITIES.put(MNT_DOMAINS, new Redaction(new Redaction.Description("Domain objects protection"), new Redaction.Description("No registrant mntner")));
+        UNSUPPORTED_ENTITIES.put(MNT_LOWER, new Redaction(new Redaction.Description("Low level objects protection"), new Redaction.Description("No registrant mntner")));
+        UNSUPPORTED_ENTITIES.put(MNT_REF, new Redaction(new Redaction.Description("Incoming references protection"), new Redaction.Description("No registrant mntner")));
```
eshryane, line 38: Authenticate incoming references?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
```diff
@@ -0,0 +1,101 @@
...
+
+@Component
+public class RedactionObjectMapper {
+
+    private final RpslObjectDao rpslObjectDao;
+    private static final Map<AttributeType, Redaction> UNSUPPORTED_ENTITIES = Maps.newHashMap();
+    static {
+        UNSUPPORTED_ENTITIES.put(MBRS_BY_REF, new Redaction(new Redaction.Description("Indirect population of a set"), new Redaction.Description("No registrant mntner")));
+        UNSUPPORTED_ENTITIES.put(MNT_DOMAINS, new Redaction(new Redaction.Description("Domain objects protection"), new Redaction.Description("No registrant mntner")));
+        UNSUPPORTED_ENTITIES.put(MNT_LOWER, new Redaction(new Redaction.Description("Low level objects protection"), new Redaction.Description("No registrant mntner")));
```
eshryane, line 37: Authenticate more specific resources

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
```diff
@@ -0,0 +1,101 @@
...
+import static net.ripe.db.whois.common.rpsl.AttributeType.MNT_ROUTES;
+
+@Component
+public class RedactionObjectMapper {
+
+    private final RpslObjectDao rpslObjectDao;
+    private static final Map<AttributeType, Redaction> UNSUPPORTED_ENTITIES = Maps.newHashMap();
+    static {
+        UNSUPPORTED_ENTITIES.put(MBRS_BY_REF, new Redaction(new Redaction.Description("Indirect population of a set"), new Redaction.Description("No registrant mntner")));
+        UNSUPPORTED_ENTITIES.put(MNT_DOMAINS, new Redaction(new Redaction.Description("Domain objects protection"), new Redaction.Description("No registrant mntner")));
```
eshryane, line 36: Authenticate domain objects

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
```diff
@@ -0,0 +1,101 @@
...
+import static net.ripe.db.whois.common.rpsl.AttributeType.MNT_REF;
+import static net.ripe.db.whois.common.rpsl.AttributeType.MNT_ROUTES;
+
+@Component
+public class RedactionObjectMapper {
+
+    private final RpslObjectDao rpslObjectDao;
+    private static final Map<AttributeType, Redaction> UNSUPPORTED_ENTITIES = Maps.newHashMap();
+    static {
+        UNSUPPORTED_ENTITIES.put(MBRS_BY_REF, new Redaction(new Redaction.Description("Indirect population of a set"), new Redaction.Description("No registrant mntner")));
```
eshryane, line 35: Authenticate members by reference

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
```diff
@@ -0,0 +1,101 @@
...
+    private void checkContactEntityLevelCreation(final List<Redaction> redactions, final RpslAttribute rpslAttribute, final AttributeType rpslAttributeType) {
+        final List<RpslObject> referencedRpslObject = getObject(rpslAttributeType.getReferences(), rpslAttribute.getCleanValues());
+        if (!referencedRpslObject.isEmpty()) {
+            for (final Map.Entry<AttributeType, Redaction> unsupportedVcard : UNSUPPORTED_VCARDS.entrySet()) {
+                for (final RpslObject rpslObject : referencedRpslObject) {
+                    if (rpslObject.containsAttribute(unsupportedVcard.getKey())) {
+                        createRedaction(unsupportedVcard.getValue(), String.format(UNSUPPORTED_VCARD_SYNTAX,
+                                        CONTACT_ATTRIBUTE_TO_ROLE_NAME.get(rpslAttributeType).name(), unsupportedVcard.getKey()),
+                                redactions);
+                        break;
```
maggarwal13, line 79: Why are you breaking here? Are you sure there cannot be multiple attribute type in rpslObject? Also same rpslObject can have notify as well as e_mail?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    private void checkContactEntityLevelCreation(final List<Redaction> redactions, final RpslAttribute rpslAttribute, final AttributeType rpslAttributeType) {
+        final List<RpslObject> referencedRpslObject = getObject(rpslAttributeType.getReferences(), rpslAttribute.getCleanValues());
+        if (!referencedRpslObject.isEmpty()) {
+            for (final Map.Entry<AttributeType, Redaction> unsupportedVcard : UNSUPPORTED_VCARDS.entrySet()) {
+                for (final RpslObject rpslObject : referencedRpslObject) {
+                    if (rpslObject.containsAttribute(unsupportedVcard.getKey())) {
+                        createRedaction(unsupportedVcard.getValue(), String.format(UNSUPPORTED_VCARD_SYNTAX,
+                                        CONTACT_ATTRIBUTE_TO_ROLE_NAME.get(rpslAttributeType).name(), unsupportedVcard.getKey()),
```
maggarwal13, line 77: seems to be it does not matter which rpslObject key it is ? just all the attributes from all rpslObjects? we are not using rpsl key  while putting redaction?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/Redaction.java
```diff
@@ -0,0 +1,80 @@
...
+        "name",
+        "reason",
+        "prePath",
+        "method"
+})
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+public class Redaction implements Serializable {
+
+    private Description name;
+
```
maggarwal13, line 24: can you include  @XmlElement(required = true) for madatory fields?

whois-api/src/test/java/net/ripe/db/whois/api/elasticsearch/WhoisRdapElasticServiceTestIntegration.java
```diff
@@ -2,6 +2,7 @@
 
 import com.google.common.collect.Lists;
 import com.google.common.util.concurrent.Uninterruptibles;
+import com.hazelcast.com.jayway.jsonpath.JsonPath;
 import jakarta.ws.rs.BadRequestException;
```
maggarwal13, line 6: wrong jsonPath import? Can you use jackson not hazelcast library?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -624,39 +635,35 @@ private Domain createDomain(final RpslObject rpslObject, final String requestUrl
...
         return domain;
     }
+    private static VCard createContactVCard(final RpslObject rpslObject) {
+        return commonVcard(rpslObject).build();
+
+    }
 
     private static VCard createVCard(final RpslObject rpslObject) {
+        return commonVcard(rpslObject).addEmail(rpslObject.getValuesForAttribute(E_MAIL)).build();
+    }
```
maggarwal13, line 645: Why we need this method separately for email?  also can you rename to addEmailVCard instead?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
MiguelAHM: I did a refactor here

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
MiguelAHM: no, the key doesn't matter. Just the role(s) and the type in case of contact entities. Just the type in case of higher level

maggarwal13: Then may be iterate over all attributes then? instead of double looping of keys and then attributes?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
MiguelAHM: as far as I understand we need to specify all of attributes that we are not specifying in the output because of some internal reason. This is why I am taking in account the email here and trying to fix the email as well. Let me know if I am wrong.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
```diff
@@ -0,0 +1,101 @@
...
+
+    private void createRedaction(final Redaction redaction, final String prePath, final List<Redaction> redactions) {
+        redaction.setPrePath(prePath);
+        if (!redactions.contains(redaction)){
+            redactions.add(redaction);
+        }
+    }
+
+    private List<RpslObject> getObject(final Set<ObjectType> possibleObjectTypes, final Set<CIString> lookupKeys){
+        final List<RpslObject> objects = Lists.newArrayList();
```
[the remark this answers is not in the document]
MiguelAHM: yes, I was already doing a refactor of this class. That method now as just one list of possibleObjectTypes. Not two.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
MiguelAHM: {
             "name": {
               "description": "Technical Email"
             },
             "prePath": "$.entities[?(@.roles[0]=='technical')].
               vcardArray[1][?(@[0]=='email')]",
             "method": "removal",
             "reason": {
               "description": "Server policy"
             }
             
             this is an example from the RFC.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
maggarwal13: In that case can you rename it to createVcardWithEmail Honestly I would prefer you have just one function createVcard and where you need to set email use entity.getVcard.addEmail.

Also can you rename commonVcard to createVcard? as it is not common to all it is based on objectType

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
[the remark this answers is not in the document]
MiguelAHM: I will update creating new instance of Redaction out of the static map

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
```diff
@@ -0,0 +1,101 @@
...
+    public List<Redaction> createContactOrEntityRedaction(final List<RpslAttribute> rpslAttributes){
+        // TODO: [MH] Part of this logic can be merge with the logic of contactEntities in RdapObjectMapper. Big refactor to make a clear merge
+        final List<Redaction> redactions = Lists.newArrayList();
+        for (final RpslAttribute rpslAttribute: rpslAttributes) {
+            final AttributeType rpslAttributeType = AttributeType.getByName(rpslAttribute.getKey());
+            if (UNSUPPORTED_ENTITIES.containsKey(rpslAttributeType)){
+                createRedaction(UNSUPPORTED_ENTITIES.get(rpslAttributeType), String.format(UNSUPPORTED_ENTITIES_SYNTAX, rpslAttributeType), redactions);
+            } else if (CONTACT_ATTRIBUTE_TO_ROLE_NAME.containsKey(rpslAttributeType)){
+                checkContactEntityLevelCreation(redactions, rpslAttribute, rpslAttributeType);
+            }
```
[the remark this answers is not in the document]
MiguelAHM: no, the rfc says that it have to be at the rdapObject level. There are some examples that you can see 

Screenshot 2023-10-04 at 09 46 03

the prePath is like a path that you can follow to figure out from where you are removing the value. This is why I have two different templates for the jsonPath (one of them if for entity level redaction and the other is for contact entities level)

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
MiguelAHM: as we talked in the standup I reverted the change of e-mail. We will tackle this in a separated task

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
MiguelAHM: Done

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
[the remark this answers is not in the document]
MiguelAHM: Done

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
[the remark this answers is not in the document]
MiguelAHM: I will continue checking the code of this class. I agree that is quite complicated to follow.

whois-api/pom.xml
```diff
@@ -98,6 +98,10 @@
             <groupId>com.fasterxml.jackson.module</groupId>
             <artifactId>jackson-module-jakarta-xmlbind-annotations</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.jayway.jsonpath</groupId>
+            <artifactId>json-path</artifactId>
+        </dependency>
```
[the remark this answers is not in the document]
MiguelAHM: In the tests. To check that prePath has the correct syntax

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    private static final Map<AttributeType, Pair> UNSUPPORTED_VCARDS = Maps.newHashMap();
+    {
+        //UNSUPPORTED_VCARDS.put(E_MAIL, new Pair("e-mail contact information", "Personal data"));
+        UNSUPPORTED_VCARDS.put(NOTIFY, new Pair("Updates notification e-mail information", "Personal data"));
+    }
+
+    public static String UNSUPPORTED_ENTITIES_SYNTAX = "$.entities[?(@.handle=='%s')]";
+
```
maggarwal13, line 51: If we are removing @.handle=='%s' will never be part of response ?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
```diff
@@ -0,0 +1,150 @@
...
+    public static String UNSUPPORTED_VCARD_SYNTAX = "$.entities[?(@.roles=='%s')].vcardArray[1][?(@[0]=='%s')]";
+
```
maggarwal13, line 53: We need to test the zeroth index

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
```diff
@@ -0,0 +1,150 @@
...
+    RedactionObjectMapper(final RpslObjectDao rpslObjectDao){
+        this.rpslObjectDao = rpslObjectDao;
```
maggarwal13, line 55: I think instead hardcoding prepath ourself we should use the library as you anyways adding the dependency.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
```diff
@@ -0,0 +1,150 @@
...
+
+    private static final Map<AttributeType, Pair> UNSUPPORTED_VCARDS = Maps.newHashMap();
+    {
+        //UNSUPPORTED_VCARDS.put(E_MAIL, new Pair("e-mail contact information", "Personal data"));
+        UNSUPPORTED_VCARDS.put(NOTIFY, new Pair("Updates notification e-mail information", "Personal data"));
+    }
+
+    public static String UNSUPPORTED_ENTITIES_SYNTAX = "$.entities[?(@.handle=='%s')]";
+
+    public static String UNSUPPORTED_VCARD_SYNTAX = "$.entities[?(@.roles=='%s')].vcardArray[1][?(@[0]=='%s')]";
```
maggarwal13, line 52: I think instead hardcoding prepath ourself we should use the library as you anyways adding the dependency.

maggarwal13: I understand it much better now. Thanks

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/RdapObject.java
```diff
@@ -117,6 +119,13 @@ public List<Autnum> getAutnums() {
         return autnums;
     }
 
+    public List<Redaction> getRedacted() {
+        if (redacted == null) {
+            redacted = Lists.newArrayList();
```
maggarwal13, line 124: return this.reacted == null ? Lists.newArrayList() : redacted;

MiguelAHM: I can't really explained but if I do this
rdapObject.getRefacted().addAll(x) 
Doesn't add X to the list. So I am leaving as it is, and in that Object we are doing the same for rest of attributes.
We are not using the short statement

pom.xml
```diff
@@ -62,6 +62,7 @@
         <hazelcast.version>4.2.8</hazelcast.version>
 
         <jackson.version>2.15.2</jackson.version>
+        <jackson.path.version>2.4.0</jackson.path.version>
         <jaxb.api.version>4.0.0</jaxb.api.version>
```
maggarwal13, line 66: json.path.version? Jackson and json path are different libraries

pom.xml
```diff
@@ -544,7 +545,11 @@
                 <artifactId>jackson-module-jakarta-xmlbind-annotations</artifactId>
                 <version>${jackson.version}</version>
             </dependency>
-
+            <dependency>
+                <groupId>com.jayway.jsonpath</groupId>
+                <artifactId>json-path</artifactId>
+                <version>${jackson.path.version}</version>
+            </dependency>
```
maggarwal13, line 552: is it possible to put scope as test for dependency?

whois-api/src/test/java/net/ripe/db/whois/api/rdap/WhoisRdapServiceTestIntegration.java
```diff
@@ -374,7 +378,7 @@ public void lookup_org_single_language_codes() {
                 "address:        Singel 258\n" +
                 "e-mail:         bitbucket@ripe.net\n" +
                 "descr:          Acme Carpet Organisation\n" +
-                "remark:         some remark\n" +
+                "remarks:         some remark\n" +
                 "phone:          +31 1234567\n" +
```
maggarwal13, line 382: is attribute remark or remarks ?

whois-api/src/test/java/net/ripe/db/whois/api/rdap/WhoisRdapServiceTestIntegration.java
```diff
@@ -1447,10 +1453,10 @@ public void lookup_autnum_abuse_contact_from_org() {
                 "[[version, {}, text, 4.0], " +
                 "[fn, {}, text, Abuse Contact], " +
                 "[kind, {}, text, group], " +
-                "[adr, {label=Singel 358}, text, [, , , , , , ]], " +
                 "[tel, {type=voice}, text, +31 6 12345678], " +
                 "[email, {type=email}, text, work@test.com], " +
                 "[email, {type=email}, text, personal@test.com], " +
+                "[adr, {label=Singel 358}, text, [, , , , , , ]], " +
                 "[email, {type=abuse}, text, abuse@test.net]]"));
```
maggarwal13, line 1460: Have we changed the order of email and addr? I know its not part of this MR but can you remove the real data from test as well ? Feel free to ignore if it turns out to be lot of chnages

whois-api/src/test/java/net/ripe/db/whois/api/rdap/WhoisRdapServiceTestIntegration.java
```diff
@@ -2283,6 +2291,74 @@ public void lookup_org_entity() {
...
+        assertThat(entity.getRedacted().get(2).getReason().getDescription(), is("No registrant mntner"));
+        assertDoesNotThrow(() -> JsonPath.compile(entity.getRedacted().get(2).getPrePath()));
+        assertThat("$.entities[?(@.handle=='OWNER-MNT')]", is(entity.getRedacted().get(2).getPrePath()));
+        assertThat(entity.getRedacted().get(2).getMethod(), is("removal"));
+
+        assertThat(entity.getRedacted().get(3).getName().getDescription(), is("Authenticate members by reference"));
+        assertThat(entity.getRedacted().get(3).getReason().getDescription(), is("No registrant mntner"));
+        assertDoesNotThrow(() -> JsonPath.compile(entity.getRedacted().get(3).getPrePath()));
+        assertThat("$.entities[?(@.handle=='OWNER-MNT,ANOTHER-MNT')]", is(entity.getRedacted().get(3).getPrePath()));
+        assertThat(entity.getRedacted().get(3).getMethod(), is("removal"));
```
maggarwal13, line 2357: Correct me if I am wrong, we have individual keys we never send comma separated multiple values in one handle in RDAP?

whois-api/src/test/java/net/ripe/db/whois/api/elasticsearch/WhoisRdapElasticServiceTestIntegration.java
```diff
@@ -182,6 +183,7 @@ public void setup() throws IOException {
                 "admin-c:       PP1-TEST\n" +
                 "e-mail:        org@test.com\n" +
                 "mnt-by:        OWNER-MNT\n" +
+                "mnt-ref:       OWNER-MNT\n" +
                 "created:         2022-08-14T11:48:28Z\n" +
```
maggarwal13, line 187: can we add other attributes too in this for testing mnt_lower, mnt_domains etc?? so we can have one object to test all attributes?

whois-api/src/test/java/net/ripe/db/whois/api/rdap/WhoisRdapServiceTestIntegration.java
```diff
@@ -71,7 +74,7 @@ public void setup() {
                 "upd-to:        noreply@ripe.net\n" +
                 "auth:          MD5-PW $1$d9fKeTr2$Si7YudNf4rUGmR71n/cqk/ #test\n" +
                 "mnt-by:        OWNER-MNT\n" +
-                "referral-by:   OWNER-MNT\n" +
+                "mbrs-by-ref:   OWNER-MNT\n" +
                 "created:         2011-07-28T00:35:42Z\n" +
```
maggarwal13, line 78: can you also add test with multiple value of same attribute? This should have multiple redaction with different handle for same attrbute , if I am not mistaken

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
```diff
@@ -0,0 +1,58 @@
...
+import static net.ripe.db.whois.common.rpsl.AttributeType.MNT_REF;
+import static net.ripe.db.whois.common.rpsl.AttributeType.MNT_ROUTES;
+
+@Component
+public class RedactionObjectMapper {
+
+    private static final Map<AttributeType, AttributeTypeRedaction> UNSUPPORTED_REGISTRANT_ATTRIBUTES = Map.of(
+            MBRS_BY_REF, AttributeTypeRedaction.MBRS_BY_REF,
+            MNT_DOMAINS, AttributeTypeRedaction.MNT_DOMAINS,
+            MNT_LOWER, AttributeTypeRedaction.MNT_LOWER,
```
maggarwal13, line 26: I still feel creating an enum is overkill and its also confusing when attributeType and AttributeTypeRedaction have same value. If we had to send this enum in a json response It would have make more sense to me. But feel free to ignore

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/Redaction.java
```diff
@@ -0,0 +1,117 @@
...
+        return prePath;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
```
maggarwal13, line 63: o can never be null here as this == o will return false anyways ? Also can you use  instanceof instaed checking via class?

pom.xml
MiguelAHM: oh yes, you are right.

whois-api/src/test/java/net/ripe/db/whois/api/rdap/WhoisRdapServiceTestIntegration.java
MiguelAHM: remarks, it was wrong before

whois-api/src/test/java/net/ripe/db/whois/api/rdap/WhoisRdapServiceTestIntegration.java
MiguelAHM: I will put the same order as before

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
```diff
@@ -0,0 +1,58 @@
...
+                .filter( rpslAttribute -> UNSUPPORTED_REGISTRANT_ATTRIBUTES.containsKey(rpslAttribute.getType()))
+                .map( rpslAttribute -> createRedaction(
+                        UNSUPPORTED_REGISTRANT_ATTRIBUTES.get(rpslAttribute.getType()),
+                        String.format(UNSUPPORTED_ENTITIES_SYNTAX, String.join(",", rpslAttribute.getCleanValues()))))
+                .collect(Collectors.toSet());
+    }
+
+
+    public static Set<Redaction> createContactEntityRedaction(final RpslObject referencedRpslObject, final List<Role> roles) {
+        final String joinedRoles = roles.stream().map(Role::getValue).collect(Collectors.joining(" && "));
```
maggarwal13, line 47: why not use  roles.stream().sort() ? It will give us consistent result and resolve issue of flaky test?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/Redaction.java
MiguelAHM: this was autogenerated by the code. So I am not sure, I will check

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -340,9 +337,16 @@ private RdapObject mapCommonNoticesAndPort(final RdapObject rdapResponse, final
     private RdapObject mapCommonConformances(final RdapObject rdapResponse) {
         rdapResponse.getRdapConformance().addAll(List.of(RdapConformance.CIDR_0.getValue(),
             RdapConformance.LEVEL_0.getValue(), RdapConformance.NRO_PROFILE_0.getValue()));
+        mapRedactionConformance(rdapResponse);
         return rdapResponse;
```
maggarwal13, line 341: I think we should directly add redaction in conformance , it is the tag that tells the user that we are following the redaction RFC. So if fields are not included or changed you will see redacted element

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -340,9 +337,16 @@ private RdapObject mapCommonNoticesAndPort(final RdapObject rdapResponse, final
...
     }
 
+    private void mapRedactionConformance(final RdapObject rdapResponse){
+        if (!rdapResponse.getRedacted().isEmpty() || (rdapResponse.getEntitySearchResults() != null && rdapResponse.getEntitySearchResults().stream().map(RdapObject::getRedacted).anyMatch(list -> !list.isEmpty()))){
```
maggarwal13, line 345: Correct me if I am wrong, even if redacted field is empty we will still see empty element in json response?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -620,43 +626,35 @@ private Domain createDomain(final RpslObject rpslObject, final String requestUrl
...
             domain.setSecureDNS(secureDNS);
         }
-
-        domain.getEntitySearchResults().addAll(createContactEntities(rpslObject, requestUrl));
+        mapContactEntitiesAndRedaction(domain, rpslObject, requestUrl);
         return domain;
     }
 
     private static VCard createVCard(final RpslObject rpslObject) {
+        return commonVcard(rpslObject).build();
```
maggarwal13, line 634: Can you revert this change as it was before ? if its calling the method directly no point in creating two methods?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -495,21 +499,23 @@ private List<Entity> createContactEntities(final RpslObject rpslObject, final St
...
         }
 
-        return entities;
+        rdapObject.getRedacted().addAll(RedactionObjectMapper.createEntityRedaction(rpslObject.getAttributes()));
+        rdapObject.getEntitySearchResults().addAll(entities);
     }
 
-    private void mapEntityVcard(final Map.Entry<CIString, Set<AttributeType>> entry, final Entity entity,
-                                final Set<ObjectType> objectPossibleTypes) {
+    private void mapEntityVcardAndContactRedaction(final RdapObject rdapObject, final Map.Entry<CIString, Set<AttributeType>> entry, final Entity entity,
```
maggarwal13, line 510: rename to mapVcardAndRedaction? Feel free to ignore

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
```diff
@@ -0,0 +1,58 @@
...
+            MBRS_BY_REF, AttributeTypeRedaction.MBRS_BY_REF,
+            MNT_DOMAINS, AttributeTypeRedaction.MNT_DOMAINS,
+            MNT_LOWER, AttributeTypeRedaction.MNT_LOWER,
+            MNT_REF, AttributeTypeRedaction.MNT_REF,
+            MNT_ROUTES, AttributeTypeRedaction.MNT_ROUTES);
+
+    private static final Map<AttributeType, AttributeTypeRedaction> UNSUPPORTED_PERSONAL_ATTRIBUTES = Map.of(NOTIFY, AttributeTypeRedaction.NOTIFY);
+
+    public static String UNSUPPORTED_ENTITIES_SYNTAX = "$.entities[?(@.handle=='%s')]";
+
```
maggarwal13, line 33: rename to REDACTED_ENTITIES_SYNTAX?

whois-api/src/test/java/net/ripe/db/whois/api/rdap/WhoisRdapServiceTestIntegration.java
```diff
@@ -2283,6 +2291,78 @@ public void lookup_org_entity() {
...
+                "org-type:      LIR\n" +
+                "descr:         Test organisation\n" +
+                "address:       One Org Street\n" +
+                "e-mail:        test@ripe.net\n" +
+                "language:      EN\n" +
+                "admin-c:       TP2-TEST\n" + //has email and notify
+                "tech-c:        TP1-TEST\n" +
+                "tech-c:        TP2-TEST\n" + //has email and notify
+                "mnt-ref:       OWNER-MNT\n" +
+                "mnt-ref:       OWNER-MNT\n" +
```
maggarwal13, line 2321: multiple mnt-ref with same value ? Can you change the value ?

MiguelAHM: I want to check that there are not two redactions created for same attribute + same value. 
If you think that it is pointless I will

maggarwal13: I think we need a test with multiple values for same attribute. Rpsl parsing takes care of same value and same attribute

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/Redaction.java
```diff
@@ -0,0 +1,117 @@
...
+import jakarta.xml.bind.annotation.XmlAccessType;
+import jakarta.xml.bind.annotation.XmlAccessorType;
+import jakarta.xml.bind.annotation.XmlElement;
+import jakarta.xml.bind.annotation.XmlType;
+
+import java.io.Serializable;
+import java.util.Objects;
+
+@XmlAccessorType(XmlAccessType.FIELD)
+@XmlType(name = "Redaction", propOrder = {
```
maggarwal13, line 13: should be lowercase "redaction"?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
MiguelAHM: I this that the RFC says to add just the redaction when the removal is applied I will double check

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
MiguelAHM: you once told me not to modify the main code because of a tests situation, I discarded doing that because I remember that suggestion.
But yes, I agree

maggarwal13: No its more about giving consistent result to user when they hit same url multiple times

whois-api/src/test/java/net/ripe/db/whois/api/rdap/WhoisRdapServiceTestIntegration.java
MiguelAHM: let me check this

whois-api/src/test/java/net/ripe/db/whois/api/rdap/WhoisRdapServiceTestIntegration.java
MiguelAHM: Done

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
MiguelAHM: ¬¬ 
done, for me it is the same. Even switch will be loaded in memory as the enum will because both are static. 
But if you feel more comfortable with the switch it is fine

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
MiguelAHM: previously we were giving a random order as well in the roles. Updated

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
MiguelAHM: nop, I sent you a json output from dev without redaction

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
MiguelAHM: Done

whois-api/src/test/java/net/ripe/db/whois/api/rdap/WhoisRdapServiceTestIntegration.java
MiguelAHM: Done

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/Redaction.java
MiguelAHM: yes

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java

whois-api/src/test/java/net/ripe/db/whois/api/rdap/WhoisRdapServiceTestIntegration.java
MiguelAHM: yes, you are right it should come in two different redactions

MiguelAHM: Done

whois-api/src/test/java/net/ripe/db/whois/api/elasticsearch/WhoisRdapElasticServiceTestIntegration.java
MiguelAHM: Done

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
MiguelAHM: I will add the conformance I don't see anything in the rfc now

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/Redaction.java
MiguelAHM: Done

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
```diff
@@ -0,0 +1,65 @@
...
+    public static String UNSUPPORTED_VCARD_SYNTAX = "$.entities[?(@.roles=='%s')].vcardArray[1][?(@[0]=='%s')]";
+
+    public static Set<Redaction> createEntityRedaction(final List<RpslAttribute> rpslAttributes){
+        return rpslAttributes.stream()
+                .filter( rpslAttribute -> unsupportedRegistrantAttributes.contains(rpslAttribute.getType()))
+                .flatMap( rpslAttribute -> {
+                    final List<Redaction> redactions = Lists.newArrayList();
+                    for (final CIString values : rpslAttribute.getCleanValues()) {
+                        redactions.add(createRedactionByAttributeType(AttributeType.getByName(rpslAttribute.getKey()),
```
maggarwal13, line 39: AttributeType.getByName(rpslAttribute.getKey()) is rpslAttribute.getType gives you same result?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/Redaction.java
MiguelAHM: Done

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
MiguelAHM: Done

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapServiceTestIntegration.java
```diff
@@ -2283,6 +2295,108 @@ public void lookup_org_entity() {
...
+        assertThat(entity.getRedacted().get(2).getReason().getDescription(), is("No registrant mntner"));
+        assertDoesNotThrow(() -> JsonPath.compile(entity.getRedacted().get(2).getPrePath()));
+        assertThat("$.entities[?(@.handle=='ANOTHER-MNT')]", is(entity.getRedacted().get(2).getPrePath()));
+        assertThat(entity.getRedacted().get(2).getMethod(), is("removal"));
+
+        assertThat(entity.getRedacted().get(3).getName().getDescription(), is("Authenticate incoming references"));
+        assertThat(entity.getRedacted().get(3).getReason().getDescription(), is("No registrant mntner"));
+        assertDoesNotThrow(() -> JsonPath.compile(entity.getRedacted().get(3).getPrePath()));
+        assertThat("$.entities[?(@.handle=='OWNER-MNT')]", is(entity.getRedacted().get(3).getPrePath()));
+        assertThat(entity.getRedacted().get(3).getMethod(), is("removal"));
```
maggarwal13, line 2365: can you create a method for assertations and pass on entity.getRedacted().get(index) ?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
maggarwal13: Sorry to unresolve this, I read RFC and I did not see anything specific . I would like @eshryane  suggestions on this .

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapServiceTestIntegration.java
MiguelAHM: Done, looks much better and with less duplicated code. Thanks!

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
MiguelAHM: yes, I saw nothing in the RFC related to this as well

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapServiceTestIntegration.java
```diff
@@ -2282,6 +2296,71 @@ public void lookup_org_entity() {
...
+                "last-modified:   2019-02-28T10:14:46Z\n" +
+                "source:        TEST");
+
+        final Entity entity = createResource("entity/ORG-ONE-TEST")
+                .request(MediaType.APPLICATION_JSON_TYPE)
+                .get(Entity.class);
+
+        assertThat(entity.getRedacted().size(), is(9));
+
+        assertRedaction(entity.getRedacted().get(0), "Updates notification e-mail information", "Personal data", "$.entities[?(@.roles=='registrant')].vcardArray[1][?(@[0]=='notify')]");
```
maggarwal13, line 2342: Notify is optional attribute , there can be few registrant with no notify attribute and some might, so we are telling user that we removed notify from all entities which is wrong, I think we need to mention the index here

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapServiceTestIntegration.java
```diff
@@ -2282,6 +2296,71 @@ public void lookup_org_entity() {
...
+        assertRedaction(entity.getRedacted().get(1),  "Updates notification e-mail information", "Personal data","$.entities[?(@.roles=='technical && administrative')].vcardArray[1][?(@[0]=='notify')]");
```
maggarwal13, line 2343: does JSON path checks technical && administrative takes order of element also in account , if yes than since I ask to sort the result it will not work? . Also Its not sorting anyways.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
eshryane: @maggarwal13 I'm OK to always add the redaction profile in conformance

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapServiceTestIntegration.java
MiguelAHM: email is optional as well and it is the example that the RFC is using.
I think that as far as we notify the user about the changes that we do it is not necessary to exactly point in what objects we did the change.

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapServiceTestIntegration.java
MiguelAHM: I don't really understand what you mean here

maggarwal13: Explained in zoom

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapServiceTestIntegration.java
maggarwal13: I do not think RFC examples are made specifically thinking about that. Although email is mandatory for few object types.  prePath should point to one or more elements that we redacted from the original json. Since we include key value in original response, we should not give false impression for removing a field that is not there

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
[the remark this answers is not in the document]
MiguelAHM: DOne, we will add the key of the objects in the handler of the prePath

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -101,15 +101,13 @@ class RdapObjectMapper {
...
-    private static final Map<AttributeType, Role> CONTACT_ATTRIBUTE_TO_ROLE_NAME = Maps.newHashMap();
-    static {
-        CONTACT_ATTRIBUTE_TO_ROLE_NAME.put(ADMIN_C, Role.ADMINISTRATIVE);
-        CONTACT_ATTRIBUTE_TO_ROLE_NAME.put(TECH_C, Role.TECHNICAL);
-        CONTACT_ATTRIBUTE_TO_ROLE_NAME.put(MNT_BY, Role.REGISTRANT);
-        CONTACT_ATTRIBUTE_TO_ROLE_NAME.put(ZONE_C, Role.ZONE);
-        CONTACT_ATTRIBUTE_TO_ROLE_NAME.put(ORG, Role.REGISTRANT);// TODO: [MA] both mnt_by and org have same role
-        CONTACT_ATTRIBUTE_TO_ROLE_NAME.put(MNT_IRT, Role.ABUSE);
-    }
+    private static final Map<AttributeType, Role> CONTACT_ATTRIBUTE_TO_ROLE_NAME = Map.of(
```
MiguelAHM, line 104: To be consistence with the code of the redactionMapper, I will put this as a switch case. We can fix that TODO with this step.

MiguelAHM: right, we are mapping attribute with static field and in the other class we are mapping some values for the redaction.

MiguelAHM:
test ok

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapServiceTestIntegration.java
```diff
@@ -2494,4 +2668,85 @@ private void assertTnCNotice(final Notice notice, final String value) {
...
+    }
+    private void assertPersonalRedactionInsideEntities(final Redaction redaction, final String json) {
+        final String vcardPathMatch = "$.entities[?(@.handle=='TP2-TEST')].vcardArray[1][*][0]";
+        final List<Object> vcards = JsonPath.read(json, vcardPathMatch);
+        assertThat(vcards, hasItem("version"));
+        assertThat(vcards, not(contains("notify")));
+
+        assertDoesNotThrow(() -> JsonPath.compile(redaction.getPrePath())); //prePath in correct format
+        final List<Object> entities = JsonPath.read(json, redaction.getPrePath());
+        assertThat(entities.size(), is(0)); //role, pkey and attribute do not exist in the json
```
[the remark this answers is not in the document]
MiguelAHM: The main important thing of the removal method is that the attribute does not exist in the object. This test is testing that.
I think that we should check that the redaction is not there.

maggarwal13 (approved):

PR #1322, opened 2023-10-09
create Config class

maggarwal13:


whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticSearchConfigurations.java
```diff
@@ -0,0 +1,115 @@
+package net.ripe.db.whois.api.elasticsearch;
+
+import net.ripe.db.whois.common.rpsl.AttributeSyntax;
+import net.ripe.db.whois.common.rpsl.AttributeType;
+import org.elasticsearch.common.xcontent.XContentBuilder;
+import org.elasticsearch.common.xcontent.XContentFactory;
+
+import java.io.IOException;
+
+public class ElasticSearchConfigurations {
```
eshryane, line 10: What's the difference between the 2 methods? Both return an `XContentBuilder`, can they be combined?
Can you use the `@Configuration` annotation and inject the dependency?

eshryane: Reply from Mahesh: 
"Reason is settings and mappings are two different parameter one is giving you index settings and other is giving you field mappings. Although both return same type but cannot be merged. And these are static field so I don't see a need to add a dependency injection for it" 
👍 from me

PR #1327, opened 2023-10-19
Removed obsolete geofeed prefix size tests

eshryane:


eshryane:
Integration tests passed

PR #1328, opened 2023-10-20
Support Basic HTTP Authentication

MiguelAHM:


whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisRestService.java
```diff
@@ -127,6 +133,8 @@ public Response delete(
             ssoTranslator.populateCacheAuthToUsername(updateContext, originalObject);
             originalObject = ssoTranslator.translateFromCacheAuthToUsername(updateContext, originalObject);
 
+            addBasic(request, passwords);
```
eshryane, line 136: Can the Basic auth be treated like a password for all requests?

MiguelAHM: yes and no.
- Yes, because we can use the passwds list that we are already using to pass the passwd to lower level. I am already using the same list
- No, because we need to check two things here, the name of the mntner TEST-MNT:X-PASWD. If we update a person objects for example it could have a list of mnts as mnt-by. We tackle this when we know that mnt-bys.

So we can not do the same in both situations. In one of them just a passwd will be checked and in the other one the passwd and the name

MiguelAHM: Conversation over Mattermost:

Try to unify all methods to use PasswordCredentials. Get currently is just using FilterAuthFunction and not PasswordCredentials .
This will required a little refactor of the current code.

whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisRestService.java
```diff
@@ -301,6 +313,8 @@ public Response lookup(
             queryBuilder.addFlag(QueryFlag.NO_FILTERING);
         }
 
+        addBasic(request, passwords);
+
```
maggarwal13, line 317: Do we allow both basic authentication and password as query as a parameter in the same request or it should be either one of them ? What happens when basic authentication is failed but password in query parameter is correct or vice-versa?

whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisRestService.java
```diff
@@ -427,4 +433,19 @@ void setDryRun(final UpdateContext updateContext, final String dryRun) {
             updateContext.dryRun();
         }
     }
+
+    private void addBasic(final HttpServletRequest request, final List<String> passwords){
+        final String authHeader = request.getHeader(HttpHeaders.AUTHORIZATION);
+        if (authHeader== null || !authHeader.toUpperCase().startsWith(BASIC_AUTH)){
+            return;
+        }
+        if (!HttpScheme.HTTPS.is(request.getHeader(HttpHeaders.X_FORWARDED_PROTO))){
```
maggarwal13, line 442: original query password is  a list of passwords and here do we just allow one password or did it support multiple passwords?

whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisRestService.java
```diff
@@ -231,6 +241,8 @@ public Response create(
             final RpslObject submittedObject = getSubmittedObject(request, resource, isQueryParamSet(unformatted));
             validateSubmittedCreateObject(request, submittedObject, objectType);
 
+            addBasic(request, passwords);
+
```
maggarwal13, line 245: is it only for this specific WhoisRestService or any http request through port 43? If I check WhoisServletDeployer there are lot more controllers that uses passwords as query parameter.

whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisRestService.java
maggarwal13: We should give priority to basic authentication and ignore passwords as query parameter as we plan to send warning to user to switch to basic auth. If basic auth passwords are wrong it should failed authentication

whois-api/src/main/java/net/ripe/db/whois/api/rest/DomainObjectService.java
```diff
@@ -96,10 +86,8 @@ public Response create(
 
             auditlogRequest(request);
 
-            final Credentials credentials = createCredentials(updateContext, passwords);
```
MiguelAHM, line 99: We are already creating the credentials in InternalUpdatePerformed when we do updatePerformer.createUpdate

whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisRestService.java
MiguelAHM: I think that this is a question for Ed. I think that if the user is passing one correct password we shouldn't return a 401 even if the basicAuth is not correct...We are limiting the use of the passwords query parameter which should still have the same behaviour as before (we are just deprecating it, we shouldn't modify anything there)

whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisRestService.java
MiguelAHM: we support to combine passwords and basic auth and you can still add more than one password in password parameter. I created tests for this

whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisRestService.java
MiguelAHM: I am creating a PasswordCredentials for the basicAuth yes. For the get response I was taking a look and it is not easy to merge everything with the update. Maybe we can merge in the future in another task, but the complexity of that task will be 5 for sure.

whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisRestService.java
maggarwal13: Does basic Auth supports multiple password? Can you point to test case ? Reason I am asking is password as query parameter is list of password which means user can specify multiple password (may be they need to authenticate with multiple passwords)

whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisRestService.java
maggarwal13: If we are  saying password as a query parameter is deprecated and user should use basic auth then we should always give priority to basicvAuth and ignore password as parameter if basicAuth is available else it is confusing and error prone and makes it less secure

whois-api/src/main/java/net/ripe/db/whois/api/rest/ReferencesService.java
```diff
@@ -187,21 +187,18 @@ public Response create(
...
 
-            final WhoisResources updatedResources = performUpdates(request, actionRequests, passwords, crowdTokenKey, null, SsoAuthForm.ACCOUNT, null);
+            final Origin origin = updatePerformer.createOrigin(request);
+            final UpdateContext updateContext = updatePerformer.initContext(origin, crowdTokenKey, request);
+            updateContext.setBatchUpdate();
+            updateContext.setBasicAuth(BasicAuthExtractor.getBasicAuth(request));
+            auditlogRequest(request);
+
+            final List<Update> updates = prepareUpdates(updateContext, actionRequests, passwords, null, SsoAuthForm.ACCOUNT, null);
+            final WhoisResources updatedResources = performUpdates(origin, updateContext, updates, request);
```
MiguelAHM, line 197: We were adding in one transaction somethings not related to the call to the database. 
I split this big method in two, one of them to prepareUpdates and the other one "performUpdates" just to apply the calls to the database in the transaction

whois-api/src/main/java/net/ripe/db/whois/api/rest/ReferencesService.java
```diff
@@ -487,16 +492,12 @@ private void removeDuplicatesAndRestoreReplacedReferences(final WhoisResources w
...
-    private Response performUpdate(
+    private void performUpdate(
                 final HttpServletRequest request,
                 final RpslObject rpslObject,
                 final String deleteReason,
-                final List<String> passwords,
-                final String crowdTokenKey) {
+                final List<String> passwords, final Origin origin, final UpdateContext updateContext) {
         try {
-            final Origin origin = updatePerformer.createOrigin(request);
```
MiguelAHM, line 497: In Delete we were calculating origin and updateContext twice.
One inside the method and the other in performUpdates. Now it is just calculated once and passing the origin and the updateContext to each method

whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisRestService.java
MiguelAHM: lets discuss with Ed, I see your point of view but Ed told me to add the basicAuth to passwords...not to overlap passwords with basic. This is easy to do with the current code, so let's see what he says because I think that each of you have a different idea.

whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisRestService.java
MiguelAHM: I think that make not sense to create a test for it (this test will always throw an error 401 doesn't matter if we change the code or not)...you can not add multiple basic auth as far as I know. 
I will using manual tests but I think that this is not possible

https://stackoverflow.com/questions/29282578/multiple-http-authorization-headers
No, it's not possible. See the syntax definition in http://greenbytes.de/tech/webdav/rfc7235.html#header.authorization

whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisRestService.java
eshryane: Don't ignore `password` yet we can't deprecate it until we've informed users and see decreased usage, the impact is too great otherwise

eshryane: Just consider basic as another `password` for now

whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisRestService.java
eshryane: This is an drawback to basic authentication, and an advantage of using `password` as sometimes you need to authenticate as multiple mntners, although it's an edge case - we should check if anyone actually does that

whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisRestService.java
eshryane: I agree you should also add to Syncupdates

PR #1329, opened 2023-10-20
Add HTTPS Support to Whois Backend

eshryane:


whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisServletDeployer.java
```diff
@@ -49,6 +50,7 @@ public class WhoisServletDeployer implements ServletDeployer {
     private final FullTextSearchService fullTextSearch;
     private final BatchUpdatesService batchUpdatesService;
     private final HealthCheckService healthCheckService;
+    private final ClientCertificateService clientCertificateService;
```
maggarwal13, line 54: Do we need this ClientCertificateService for RdapServletDeployer and Nrtmv4 servlet as well?

eshryane: No - not for now - we only need Client Certificate auth on the Whois REST API. This `ClientCertificateService` class is a placeholder until we integrate it into update authentication

maggarwal13 (approved):

eshryane:
Integration tests passed

PR #1330, opened 2023-10-20
Add warn to deprecated password parameter

MiguelAHM:
depends on https://github.com/RIPE-NCC/whois/pull/1328/files


MiguelAHM:
We will not add the deprecated message for now, it will show up in the web app and we need to think a way of doing that

PR #1332, opened 2023-10-27
Db 4900 filter email rdap

MiguelAHM:


MiguelAHM:
We will delete the email in all the objects, so it will not go per object type.
I will close this PR. But is it s draft for next attributes which depends on the object type

PR #1335, opened 2023-10-30
Remove organisation e-mail from RDAP readme

MiguelAHM:
E-mail is a redacted field, as notify
we will not send e-mail or notify in the output


eshryane (approved):

eshryane:
I updated the title to make it clearer

PR #1339, opened 2023-10-31
Only use X-Forwarded-For as client address if loadbalancer is proxying HTTP requests

eshryane:
Depends on #1329


fmurodov:
This should be updated by checking the request protocol when X-FORWARDED_PROTO is not available.
https://github.com/RIPE-NCC/whois/blob/ef381f2d932ac848a1c93bfdf778ab7d9cdd5b12/whois-api/src/main/java/net/ripe/db/whois/api/nrtm4/NrtmHttpSchemeFilter.java#L40-L42

eshryane:
> This should be updated by checking the request protocol when X_FORWARDED_PROTO is not available.
> 
> https://github.com/RIPE-NCC/whois/blob/ef381f2d932ac848a1c93bfdf778ab7d9cdd5b12/whois-api/src/main/java/net/ripe/db/whois/api/nrtm4/NrtmHttpSchemeFilter.java#L40-L42

Thanks @fmurodov I was looking for use of X-Forwarded-Proto, I'll update this

fmurodov (approved):

PR #1340, opened 2023-10-31
Add multiple lang country redaction

maggarwal13:


whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
```diff
@@ -33,28 +40,43 @@ public static void mapRedactions(final RdapObject rdapObject) {
...
-            switch(attributeType){
-                case NOTIFY -> redactions.add(new Redaction("Updates notification e-mail information", String.format("%s.vcardArray[1][?(@[0]=='%s')]", prefix, attributeType.getName()),
-                        "Personal data"));
-                case E_MAIL -> redactions.add(new Redaction("Personal e-mail information", String.format("%s.vcardArray[1][?(@[0]=='%s')]", prefix, attributeType.getName()),
-                        "Personal data"));
+
+        final Map<AttributeType, List<CIString>> attributeTypeByValues =  rpslAttributes.stream()
+                .collect(Collectors.groupingBy(RpslAttribute::getType, Collectors.mapping(RpslAttribute::getCleanValue, Collectors.toList())));
+
+        attributeTypeByValues.entrySet().forEach(entry -> {
```
MiguelAHM, line 55: you don't need entrySet() -> attributeTypeByValues.forEach((key, value) -> {

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
```diff
@@ -33,28 +40,43 @@ public static void mapRedactions(final RdapObject rdapObject) {
...
+            final String attributeName = entry.getKey().getName();
+            final String values = entry.getValue().stream().collect(Collectors.joining(", "));
```
MiguelAHM, line 57: after previous change you no longer need to get the values from the entry.
so you can directly:
final String values = String.join(", ", value)

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
```diff
@@ -1,29 +1,36 @@
...
-    public static final List<AttributeType> REDACTED_PERSONAL_ATTR = Lists.newArrayList(AttributeType.NOTIFY,
-            AttributeType.E_MAIL);
     public static void mapRedactions(final RdapObject rdapObject) {
 
-        addEntityRedaction(rdapObject);
+        rdapObject.getRedacted().addAll(getRedactions(rdapObject.getRedactedRpslAttrs(), "$"));
 
         if(rdapObject.getNetwork() != null) {
-            addEntitiesRedaction(rdapObject, rdapObject.getNetwork().getEntitySearchResults(), "$.network");
+            final String prefix = "$.network";
```
MiguelAHM, line 23: could we create again the method addEntityRedaction which will call addEntitiesRedaction
and move
rdapObject.getRedacted().addAll(getRedactions(rdapObject.getNetwork().getRedactedRpslAttrs(), prefix));
inside that method?

Then we could just put addEntityRedaction(rdapObject, "$.network");
Then we could just put addEntityRedaction(rdapObject, "$.networks[?(@.handle=='%s')]");

I had to debug to see what line 25 is doing, basically is create the redaction of the main entity.

In this case it is clear where we create a redaction for whole entity and when we don't. for me the code is much easier on that way

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapServiceTestIntegration.java
```diff
@@ -2773,38 +2836,63 @@ private void assertTnCNotice(final Notice notice, final String value) {
...
+                String.format("There are multiple %s attributes %s found, but only the first %s %s returned.", type.getName(), multipleValues, type.getName(), multipleValues.split(",")[0]));
+
+        assertThat(redaction, equalTo(expectedRedaction));
 
-        List<Object> vcards = JsonPath.read(entityJson, entity.getRedacted().get(redaction).getPrePath());
+        final String entityJson = getEntityJson(rdapObject);
+        final Object redactedElement =  JsonPath.read(entityJson, redaction.getPostPath());
+
+        if(redactedElement instanceof JSONArray) {
+            //incase of networks  it is a list
```
MiguelAHM, line 2853: //in case

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapServiceTestIntegration.java
```diff
@@ -2773,38 +2836,63 @@ private void assertTnCNotice(final Notice notice, final String value) {
         assertThat(notice.getLinks().get(0).getValue(), is(value));
     }
 
-    private void assertPersonalRedaction(final Entity entity, final int redaction, final AttributeType attribute) throws JsonProcessingException {
-        final String entityJson = getObjectMapper().writeValueAsString(entity);
+    private void assertMultipleValuesRedaction(final RdapObject rdapObject, final String prefix, final AttributeType type,final String multipleValues) {
+        final String rdapAttrName = (type == LANGUAGE) ? "lang" : type.getName();
+        final Redaction redaction = rdapObject.getRedacted().stream().filter( redaction1 -> redaction1.getPostPath()!= null && redaction1.getPostPath().contains(rdapAttrName)).findFirst().get();
```
MiguelAHM, line 2841: redactedField instead of redaction1? I can get a better name

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapServiceTestIntegration.java
```diff
@@ -2773,38 +2836,63 @@ private void assertTnCNotice(final Notice notice, final String value) {
...
-        List<Object> vcards = JsonPath.read(entityJson, entity.getRedacted().get(redaction).getPrePath());
+        List<Object> vcards = JsonPath.read(entityJson, redaction.getPrePath());
         assertThat(vcards.size(), is(0));
 
         final Entity insideEntity = entities.stream().filter( contacEntity -> contacEntity.getHandle().equals(personKey)).findFirst().get();
         assertCommonPersonalRedaction(entity, redaction, insideEntity, attribute);
     }
 
-    private void assertCommonPersonalRedaction(final RdapObject entity, final int redaction,
+    private void assertCommonPersonalRedaction(final RdapObject entity, final Redaction redaction,
```
MiguelAHM, line 2883: you can remove throws JsonProcessingException

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
maggarwal13: Not really, Its a single line so I do not  see a need for putting in a method also not every entity calls entities.

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapServiceTestIntegration.java
maggarwal13: No it will not work as its a map and maps are not ordered so has to be like this

maggarwal13:
> Could we add a test about one organisation which multiple language and countries? and with networks? more or less the same that we are doing the ORG-TEST1-TEST but with multiple country and languages? To see how is the behaviour in the main entity as well?

we do not map country for Entity Objects and there is test for each scenario "lookup_org_single_language_codes" , lookup_org_multiple_language_codes and lookup_multiple_attr_redactions_in_networks

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
MiguelAHM: I think that it is much readable to be honest. But up to you

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RedactionObjectMapper.java
MiguelAHM: yes, but if you need to use both the keys and the values in the code that if inside the foreach. It is more handy to have key and value and not just the entry.

PR #1342, opened 2023-11-02
Allow basic auth

maggarwal13:


whois-api/src/main/java/net/ripe/db/whois/api/rest/HttpsBasicAuthCustomizer.java
```diff
@@ -0,0 +1,90 @@
...
+
+        chain.doFilter(new HttpBasicAuthRequestWrapper((HttpServletRequest) request), response);
+    }
+
+    private boolean canProceed(final ServletRequest request) {
+       if(! (request instanceof HttpServletRequest)) {
+           return false;
+       }
+
+       if(!RestServiceHelper.isBasicAuth((HttpServletRequest) request)) {
```
MiguelAHM, line 49: here you can directly 
return !RestServiceHelper.isBasicAuth((HttpServletRequest) request)

and remove the return true; below, previous code will tackle this already

whois-api/src/main/java/net/ripe/db/whois/api/rest/HttpsBasicAuthCustomizer.java
```diff
@@ -0,0 +1,90 @@
...
+            final String basicAuthPassword = getBasicAuthPassword(getHeader(HttpHeaders.AUTHORIZATION));
+            if(basicAuthPassword == null) {
+                return super.getQueryString();
+            }
+
+            final StringBuilder modifiedQueryString = (super.getQueryString() == null) ?
+                                                             new StringBuilder() :
+                                                             new StringBuilder(super.getQueryString()).append("&");
+
+            return modifiedQueryString.append("password=").append(basicAuthPassword).toString();
```
MiguelAHM, line 78: will this code add password=xxxx in all the endpoints even if the endpoint do not have that parameter?

I mean if the user uses basicAuth in one endpoint that does not support "password" query parameter, we will add that parameter there, right?

This will be applied to all the API services

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestBasicAuthTestIntegration.java
```diff
@@ -0,0 +1,273 @@
...
+        databaseHelper.updateObject(TEST_PERSON);
+        databaseHelper.updateObject(TEST_ROLE);
+        testDateTimeProvider.setTime(LocalDateTime.parse("2001-02-04T17:00:00"));
+    }
+
+    @Test
+    public void lookup_mntner_incorrect_basic_auth_password_without_unfiltered_param_is_fully_filtered() {
+        final WhoisResources whoisResources = SecureRestTest.target(getSecurePort(), "whois/test/mntner/OWNER-MNT")
+                .request()
+                .header("Authorization", getBasicAuthenticationHeader("test", "incorrect"))
```
MiguelAHM, line 108: HttpHeaders.AUTHORIZATION ? instead of Authorization?

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestBasicAuthTestIntegration.java
```diff
@@ -0,0 +1,273 @@
...
+                .post(Entity.entity(map(PAULETH_PALTHEN), MediaType.APPLICATION_XML), Response.class);
+
+        assertThat(response.getHeaderString(HttpHeaders.WWW_AUTHENTICATE), is(BASIC_CHARSET_ISO_8859_1_LATIN_1));
+    }
+
+    @Test
+    public void create_failed_with_no_basic_auth_incorrect_password_no_www_authenticate_cookie() {
+        final Response response = SecureRestTest.target(getSecurePort(), "whois/test/person?password=incorrect")
```
MiguelAHM, line 173: This should thrown an  NotAuthorizedException, right?

it is not better doing 

 NotAuthorizedException badAuth = assertThrows(NotAuthorizedException.class, () -> {
}

assertThat(badAuth.WWW_AUTH, is null)?

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestBasicAuthTestIntegration.java
```diff
@@ -0,0 +1,273 @@
...
+                .request()
+                .post(Entity.entity(map(PAULETH_PALTHEN), MediaType.APPLICATION_XML), Response.class);
+
+        assertThat(response.getHeaderString(HttpHeaders.WWW_AUTHENTICATE), is(nullValue()));
+    }
+
+    @Test
+    public void create_failed_with_basic_auth_no_https() {
+
+        final Response response =  RestTest.target(getPort(), "whois/test/person")
```
MiguelAHM, line 183: same as before. 
Can we use assertThrows here as well?

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestBasicAuthTestIntegration.java
```diff
@@ -0,0 +1,273 @@
...
+                    .request()
+                    .header("Authorization", getBasicAuthenticationHeader("test", "test"))
+                    .post(Entity.entity(map(PAULETH_PALTHEN), MediaType.APPLICATION_XML), Response.class);
+
+        assertThat(response.getStatus(), is(HttpStatus.UPGRADE_REQUIRED_426));
+    }
+
+    @Test
+    public void create_failed_unauthorized_WWW_authenticate_header() {
+        final Response response = SecureRestTest.target(getSecurePort(), "whois/test/person?password=incorrect")
```
MiguelAHM, line 193: Can we use assertThrows here as well?

MiguelAHM:
Could we add a some tests for multiple passwords?

- multiple password parameter (one of them correct), adding an incorrect basic auth -> succeed
- multiple password parameter (all wrong), adding a correct basic auth -> succeed

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestBasicAuthTestIntegration.java
maggarwal13: It is checking the status code if it is not https  as HttpStatus.UPGRADE_REQUIRED_426. DO you mean to check the message inside the response? I will add add that

whois-api/src/main/java/net/ripe/db/whois/api/rest/HttpsBasicAuthCustomizer.java
maggarwal13: Yes t will be applied to all services and Yes parameters are optional in Http request

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestBasicAuthTestIntegration.java
MiguelAHM: example

        final ForbiddenException throwable = assertThrows(ForbiddenException.class, () ->
            RestTest.target(getPort(), "test/person/TP1-TEST?password=123")
                    .request()
                    .header(HttpHeaders.HOST, getHost(restApiBaseUrl))
                    .header(HttpHeader.X_FORWARDED_PROTO.toString(), HttpScheme.HTTP)
                    .get(WhoisResources.class)
        );

        final Entity entity = throwable.getResponse().readEntity(Entity.class);
        assertThat(entity.getErrorCode(), is(status));

doing this it is clear that there is an exception

whois-api/src/main/java/net/ripe/db/whois/api/rest/HttpsBasicAuthCustomizer.java
MiguelAHM: mmm...okok. Thanks for the answer!

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestBasicAuthTestIntegration.java
maggarwal13: I think checking status code is perfectly fine too. Status code are unique. For me its more readable and less lines.

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestBasicAuthTestIntegration.java
maggarwal13: I will add an assert for response code.

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestBasicAuthTestIntegration.java
MiguelAHM: as you wish, but I prefer to get a solution to apply always. Because in the code I have already seen more than 2 different ways of checking error.
Time ago I talked to Ed about it, and he recommended to use the assertThrows this is why I am doing some refactors time to time in some IT applying this.

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestBasicAuthTestIntegration.java
MiguelAHM: then change the name.
create_failed -> create_succeed ?

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestBasicAuthTestIntegration.java
maggarwal13: May be purpose there is to check the exceptions but purpose here is to Response contains upgrade required

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestBasicAuthTestIntegration.java
MiguelAHM: I really prefer to have a common way of testing errors.
Sometimes we are checking the error directly in the call, another times asserting the code, some times using assertThrows...
I would say to use one way and not multiple of them.
If we say that we will use assert for the response always I will go with that as well. 
My only concern is to do this in different ways through the code

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestBasicAuthTestIntegration.java
maggarwal13: Ah thanks for pointing out its a missed scenario. I will change it

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestBasicAuthTestIntegration.java
maggarwal13: But the purpose is to check WWW_authenticate header  is not there not to check exceptions. But may be you are correct, I do not mind assertThrows as well

maggarwal13:
> Could we add a some tests for multiple passwords?
> 
> * multiple password parameter (one of them correct), adding an incorrect basic auth -> succeed
> * multiple password parameter (all wrong), adding a correct basic auth -> succeed

I will add it but again this test  class is specific for basic auth not to test multiple password

MiguelAHM:
> > Could we add a some tests for multiple passwords?
> > 
> > * multiple password parameter (one of them correct), adding an incorrect basic auth -> succeed
> > * multiple password parameter (all wrong), adding a correct basic auth -> succeed
> 
> I will add it but again this test class is specific for basic auth not to test multiple password

It is related I think because in the end you are adding a new password parameter if the basic auth is in the request

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestBasicAuthTestIntegration.java
maggarwal13: I do not like putting two lines when it can be done is single line , also its much clearer with .header then property

MiguelAHM: Ok, it is more readable on your way I think yes
but could you change "Authorization" by HttpHeaders.AUTHORIZATION?

MiguelAHM (approved):

PR #1345, opened 2023-11-10
Return bad request without log when the URL is wrongly formatted RDAP

MiguelAHM:
I decided to send a 400 because the only way to make this break is when the user is putting wrong redirect in the request. So 400 seems better option than 500


eshryane:
Can you add a test case? Maybe the example in the bug

MiguelAHM:
> Can you add a test case? Maybe the example in the bug

I have tried but I receive an error because the URL is wrongly formatted. I put "?redirect:%25\\{333*444\\" and it complains about the *

eshryane:
> > Can you add a test case? Maybe the example in the bug
> 
> I have tried but I received an error because the URL is wrongly formatted (I mean before entering in the code, it is throwed by jetty). I put "?redirect:%25\{333*444\" and it complains about the *

Good point! Apologies, indeed the URL can't be parsed on the client side either. I added a test by connecting to the port directly, please review it

eshryane (approved):

PR #1347, opened 2023-11-13
Db 4939 fill redaction remaining attributes

MiguelAHM:


whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -656,38 +658,27 @@ private static AsBlockRange getAsBlockRange(final String asBlock) {
...
-
-        final String commaSeperatedValues = values.stream().map(RpslAttribute::getCleanValue).collect(Collectors.joining(", "));
-        final String title = String.format("Multiple %s attributes found", type.getName());
-        final String desc = String.format("There are multiple %s attributes %s in %s, but only the first %s %s was returned.", type.getName(), commaSeperatedValues, key, type.getName(), values.get(0).getCleanValue());
-
-        final Notice notice = new Notice();
-        notice.setTitle(title);
-        notice.getDescription().add(desc);
-        rdapObject.getNotices().add(notice);
+    private static void handlePartialValueAttributes(final RpslObject rpslObject, final RdapObject rdapObject){
```
eshryane, line 661: Does this logic need to be combined into one method? The term "partial value" is not defined. Can language and country handling be kept separate?

eshryane, line 661: The warnings on multiple language or country attributes have been moved into the redacted code, but those warnings are not related to redaction. Can they be kept with the attribute parsing and the redaction kept separate?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/VCardBuilder.java
```diff
@@ -45,11 +45,6 @@ public VCardBuilder addAdr(final Set<CIString> addresses) {
         return this;
     }
 
-    public VCardBuilder addEmail(final Set<CIString> emails) {
```
eshryane, line 48: Maybe move the cleanup into a separate PR so it's better separated, it's confusing to do unrelated work in the same PR.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
MiguelAHM: I think so, if not both method will have exactly the same logic. The difference will be the attribute...which for me makes it better to keep in just one method (same logic - just different attributes)
On the other hand, if you keep it into different methods you will need to call it from 

createIp -> To the one related to the country and lang
createEntity -> To the one that just create the country

So you will need to do the call from different parts and to do exactly the same logic.

About the name I am totally agree that can be better, I accept suggestions I will try to think about a better name.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/VCardBuilder.java
MiguelAHM: I will

MiguelAHM: https://github.com/RIPE-NCC/whois/pull/1350

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
MiguelAHM: "handleMultipleValuesAttributes" this could be a better name because we already know that in the RDAP class object we just can have a key-value attributes. So this name I think that makes clear that we are handling those attributes with multiple values

eshryane (approved):

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
eshryane: Agreed that warning notices are replaced by redaction element

whois-api/src/main/java/net/ripe/db/whois/api/rdap/VCardBuilder.java
eshryane: Thanks

PR #1351, opened 2023-11-16
Personal object accounting by sso

maggarwal13:


whois-query/src/main/java/net/ripe/db/whois/query/acl/AccessControlListManager.java
```diff
@@ -55,40 +73,61 @@ public boolean requiresAcl(final RpslObject rpslObject, final Source source) {
...
+    public boolean canQueryPersonalObjects(final InetAddress remoteAddress, final String ssoToken) {
+        return getPersonalObjects(remoteAddress,ssoToken) >= 0;
     }
 
     public boolean isTrusted(final InetAddress remoteAddress) {
         return ipRanges.isTrusted(IpInterval.asIpInterval(remoteAddress));
     }
 
-    public int getPersonalObjects(final InetAddress remoteAddress) {
+    public int getPersonalObjects(final InetAddress remoteAddress, final String ssoToken) {
```
eshryane, line 101: Overload `getPersonalObjects()` with only a `remoteAddress` parameter so you don't need to invoke it with `null` from other classes?

maggarwal13: Actually I wanted to have only one way to account for personal objects. So that we do not end up up calling a method for remoteAddr by mistake from somewhere even though sso is available. Plus lots of these services will eventually have sooToken available when db-webui changes are in place. Else we will end up having logic in all service calls that decide which method to call remoteAddr or sso

whois-query/src/test/resources/applicationContext-query-test-mock.xml
```diff
@@ -15,7 +15,7 @@
         <constructor-arg value="net.ripe.db.whois.query.handler.QueryHandler"/>
     </bean>
 
-    <bean id="accessControlListManager" class="org.mockito.Mockito" factory-method="mock" primary="true">
+    <bean id="AccessControlListManager" class="org.mockito.Mockito" factory-method="mock" primary="true">
```
eshryane, line 18: Use lowercase for the first word of the bean-id by convention, i.e. `accessControlListManager`

whois-query/src/main/java/net/ripe/db/whois/query/acl/AccessControlListManager.java
```diff
@@ -55,71 +76,172 @@ public boolean requiresAcl(final RpslObject rpslObject, final Source source) {
...
+            return null;
+        }
+
+        try {
+            final UserSession userSession = ssoTokenTranslator.translateSsoToken(ssoToken);
+            if(userSession != null && !StringUtils.isEmpty(userSession.getUsername())) {
+                return userSession.getUsername();
+            }
+        } catch (AuthServiceClientException e) {
+            LOGGER.warn("Cannot translate ssoToken, will account by remoteAddr", e.getMessage());
```
MiguelAHM, line 136: Are you missing the {}, right?

whois-query/src/main/java/net/ripe/db/whois/query/acl/AccessControlListManager.java
```diff
@@ -55,71 +76,172 @@ public boolean requiresAcl(final RpslObject rpslObject, final Source source) {
...
+        }
+
+        return null;
     }
 
     /**
      * Account for the ResponseObject given
      *
-     * @param remoteAddress The remote address.
+     * @param accountingIdentifier The remote address and ssoken
```
MiguelAHM, line 145: typo? sso token?

whois-query/src/main/java/net/ripe/db/whois/query/acl/AccessControlListManager.java
```diff
@@ -55,71 +76,172 @@ public boolean requiresAcl(final RpslObject rpslObject, final Source source) {
...
+        accountingManager.accountPersonalObjects(amount);
     }
 
-    public void blockTemporary(final InetAddress hostAddress, final int limit) {
-        IpInterval<?> maskedAddress;
-        if (hostAddress instanceof Inet6Address) {
-            maskedAddress = Ipv6Resource.parse(mask(hostAddress, IPV6_NETMASK).getHostAddress() + "/" + IPV6_NETMASK);
-        } else {
-            maskedAddress = Ipv4Resource.asIpInterval(hostAddress);
+    public class SSOAccountingManager implements PersonalAccountingManager {
```
MiguelAHM, line 157: why public? it is being called just from AccessControlListManager class, right?

whois-query/src/main/java/net/ripe/db/whois/query/acl/AccessControlListManager.java
```diff
@@ -55,71 +76,172 @@ public boolean requiresAcl(final RpslObject rpslObject, final Source source) {
...
+        }
+
+        @Override
+        public int getPersonalDataLimit() {
+            return ssoResourceConfiguration.getLimit();
         }
-        accessControlListDao.saveAclEvent(maskedAddress, dateTimeProvider.getCurrentDate(), limit, BlockEvent.Type.BLOCK_TEMPORARY);
     }
 
+    public class RemoteAddrAccountingManager implements PersonalAccountingManager {
```
MiguelAHM, line 194: why public? it is being called just from AccessControlListManager class, right?

whois-query/src/main/java/net/ripe/db/whois/query/acl/HazelcastPersonalObjectAccounting.java
```diff
@@ -18,14 +18,15 @@
 @Component
 public class HazelcastPersonalObjectAccounting implements PersonalObjectAccounting {
     private static final Logger LOGGER = LoggerFactory.getLogger(HazelcastPersonalObjectAccounting.class);
-
-    private final IMap<InetAddress, Integer> counterMap;
+    private final IMap<InetAddress, Integer> remoteAddrCounterMap;
+    private final IMap<String, Integer> ssoCounterMap;
     private final HazelcastInstance hazelcastInstance;
```
MiguelAHM, line 23: We are not using this. It is just composing rest of the variables and the log, so I suggest removing it from the variables and use the parameter of the constructor to compose the rest of the variables

whois-query/src/main/java/net/ripe/db/whois/query/acl/HazelcastPersonalObjectAccounting.java
```diff
@@ -34,34 +35,43 @@ public HazelcastPersonalObjectAccounting(final HazelcastInstance hazelcastInstan
...
+        Integer count = null;
+        try {
+            count = ssoCounterMap.get(ssoId);
+        } catch (OperationTimeoutException | IllegalStateException e) {
+            // no answer from hazelcast, expected, don't rethrow
+            LOGGER.debug("{}: {}", e.getClass().getName(), e.getMessage());
         }
 
-        return count;
+        return count == null ? 0 : count;
```
MiguelAHM, line 57: Instead of doing this, why not instanciate Integer count with a 0 instead a null?
Integer count = 0;
...
return count;

whois-query/src/main/java/net/ripe/db/whois/query/acl/HazelcastPersonalObjectAccounting.java
maggarwal13: It was like this before , so I did not change. Lets do it in separate refactoring Jira to reduce scope. Its alreday a big PR

whois-query/src/main/java/net/ripe/db/whois/query/acl/HazelcastPersonalObjectAccounting.java
maggarwal13: Because by default Integer class has initilialization null and you cannot initialize 0 for all remoteAdress. so if entry is not available in map it will be returned null not zero

whois-query/src/main/java/net/ripe/db/whois/query/acl/HazelcastPersonalObjectAccounting.java
```diff
@@ -74,18 +84,53 @@ public int accountPersonalObject(final InetAddress remoteAddress, final int amou
         return 0;
     }
 
-    private void unlockKey(InetAddress remoteAddress) {
+    @Override
+    public int accountPersonalObject(final String ssoId, final int amount) {
+        boolean isLocked = false;
```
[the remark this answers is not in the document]
maggarwal13: Since the map is different and its signature as well so I preferred to do it separately as there is not much logic in it mostly putting an entry in map

whois-query/src/main/java/net/ripe/db/whois/query/acl/HazelcastPersonalObjectAccounting.java
MiguelAHM: Agree that it is out of scope

whois-query/src/main/java/net/ripe/db/whois/query/dao/jdbc/JdbcSSOAccessControlListDao.java
```diff
@@ -0,0 +1,120 @@
...
+        }
+    }
+
+
+    private static class BlockEventsExtractor implements ResultSetExtractor<List<BlockEvents>> {
+        @Override
+        public List<BlockEvents> extractData(final ResultSet rs) throws SQLException, DataAccessException {
+            final Map<String, List<BlockEvent>> blockEventsMap = new HashMap<>();
+            while (rs.next()) {
+                final String prefix = rs.getString("sso_id");
```
MiguelAHM, line 65: ssoId instead of prefix?

whois-query/src/main/java/net/ripe/db/whois/query/dao/jdbc/JdbcSSOAccessControlListDao.java
```diff
@@ -0,0 +1,120 @@
...
+            }
+
+            return result;
+        }
+    }
+
+    @Override
+    public List<BlockEvents> getTemporaryBlocks(final LocalDate blockTime) {
+        return jdbcTemplate.query(
+                "SELECT sso_id, event_time, daily_limit, event_type FROM acl_sso_event WHERE event_time >= ? ORDER BY prefix, event_time ASC",
```
MiguelAHM, line 90: ORDER BY sso_id I guess. Correct me if I am wrong.

MiguelAHM (approved):

PR #1352, opened 2023-11-17
Include AS-SET in NONAUTH dump

eshryane:


whois-scheduler/src/main/java/net/ripe/db/whois/scheduler/task/export/FilenameStrategy.java
```diff
@@ -32,7 +33,7 @@ public String getFilename(final ObjectType objectType) {
     class NonAuthSplitFile implements FilenameStrategy {
         @Override
         public String getFilename(final ObjectType objectType) {
-            return objectType == AUT_NUM || objectType == ROUTE || objectType == ROUTE6?
+            return (objectType == AS_SET || objectType == AUT_NUM || objectType == ROUTE || objectType == ROUTE6) ?
```
MiguelAHM, line 36: Question:

Is it possible to use the ExportFilter of the previous class and do something list
return exportFilter.getObjects.contains(objectType) ?
"ripe-nonaith.db." ....etc

It would be handy just to do the modification in one place instead of in two places.

MiguelAHM (approved):

whois-scheduler/src/main/java/net/ripe/db/whois/scheduler/task/export/FilenameStrategy.java
eshryane: I'll check

MiguelAHM: Thanks for the change ^^
Now just one change is needed if in the future we need to add another object to nonauth source.

eshryane:
Successfully tested that RIPE-NONAUTH as-set objects are added to the non-auth as-set split file and combined nonauth dump file.

PR #1354, opened 2023-11-17
Fix "export" endtoend Tests

eshryane:
Expect export files to be named after "test" source and not "ripe" source.
Use "public" for external (filtered) dump vs. "internal" for private dump.


eshryane:
Integration tests passed [6e414c3](https://github.com/RIPE-NCC/whois/pull/1354/commits/6e414c379a47bb2612520b1e96843895e33ea290)

PR #1361, opened 2023-11-29
Fulltext search api sso accounting

maggarwal13:


eshryane (approved):

MiguelAHM (approved):

whois-query/src/main/java/net/ripe/db/whois/query/query/Query.java
```diff
@@ -105,6 +105,27 @@ public static Query parse(final String args, final Origin origin, final boolean
         }
     }
 
+    public static Query parse(final String args, final String ssoToken, final Origin origin, final boolean trusted) {
```
eshryane, line 108: Can you call `parse(final String args, final Origin origin, final boolean trusted)` then set `query.ssoToken` then return? This avoids code duplication.

eshryane: It's already done this way in the other parse method: `parse(final String args, final String ssoToken, final List<String> passwords, final boolean trusted)`

maggarwal13: But this does not take origin ? and the first one I need to pass on ssoToken hence different method.

PR #1362, opened 2023-12-05
RDAP Object is Assumed to be an IP on Country Attribute

MiguelAHM:


whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -670,13 +670,14 @@ private static void handleMultipleValuesAttributes(final RpslObject rpslObject,
                     rdapObject.setLang(attributeValuesAttribute.next().toString());
                 }
                 case COUNTRY -> {
-                    final Ip ip = (Ip)rdapObject;
-                    if (!StringUtil.isNullOrEmpty(ip.getCountry())){
-                        rdapObject.getRedactedRpslAttrs().addAll(rpslObject.findAttributes(COUNTRY));
-                        break;
+                    if (rdapObject instanceof final Ip ip) {
+                        if (!StringUtil.isNullOrEmpty(ip.getCountry())) {
```
maggarwal13, line 674: If its only set in IP, can you call this method from createIp in that case you do not need to type cast or use instance of? Similarly for Lang attribute call it from createEnitity and CreateIp only

eshryane:
I updated the title to make it clearer, both for the community and ourselves in the future.

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapServiceTestIntegration.java
```diff
@@ -374,6 +374,7 @@ public void lookup_org_single_language_codes() {
                 "geoloc:         52.375599 4.899902\n" +
                 "language:       DK\n" +
                 "admin-c:        TP1-TEST\n" +
+                "country:        NL\n"+
```
eshryane, line 377: Can you also add a test to `lookup_org_single_language_codes` and `lookup_org_multiple_language_codes` that country is set properly in the organisation `Entity` response

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
eshryane: Could you create a `mapRedactions` method specifically for Ip ?

eshryane: Although then `language` is common to both IP and Organisation - maybe a separate method for that also, but call it with an Entity / RDAP object?

MiguelAHM: Pair programming:
We use just one method with all the logic and we call that method from IPs related logic and Entity related logic.
If in the future this is applied to all the object is very easy to change to the main method

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapServiceTestIntegration.java
MiguelAHM: There is already one test for lookup_org_multiple_language_codes 
I will create lookup_org_single_language_codes_no_redaction

eshryane: "country" is only specified in IP and aut-num objects, not in entities, so we can't test for it there.

eshryane (approved):

maggarwal13 (approved):

MiguelAHM:
Test ok
https://gitlab.ripe.net/swe-database-team/whois/-/pipelines/457698

PR #1367, opened 2023-12-12
Refresh Elasticsearch Index after Update/Create object

MiguelAHM:
We want to refresh the Elasticsearch index after updating it so the document count matches the RIPE database.

1. Use Slave datasource to get serials
2. Indent refactor
3. Rename add method which actually update or add
4. Add refresh to catch up last changes when requested for the Id
5. Remove the -1 because we fixed the have already fixed the POEM object
6. Tested performance with more then 1000 objects syncupdate, It took less that 0.5 sec (343.6 ms)
Number of objects found: 1014
Number of objects processed successfully: 1014

More info:
https://stackoverflow.com/questions/66873411/elasticsearch-high-level-rest-client-indexing-has-latency
https://www.elastic.co/guide/en/elasticsearch/client/java-rest/current/java-rest-high-refresh.html


whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticIndexService.java
```diff
@@ -131,14 +131,20 @@ protected void deleteAll() throws IOException {
...
+    }
+
     protected long getWhoisDocCount() throws IOException {
         if (!isElasticRunning()) {
             throw new IllegalStateException("ES is not running");
         }
 
         final CountRequest countRequest = new CountRequest(whoisAliasIndex);
-        final SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder();
-        searchSourceBuilder.query(QueryBuilders.matchAllQuery());
```
maggarwal13, line 141: I think thats the problem, countRequest needs to setQuery? Can you use setQuery for countRequest and then try out ?

whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticIndexService.java
```diff
@@ -131,14 +131,20 @@ protected void deleteAll() throws IOException {
         client.deleteByQuery(request, RequestOptions.DEFAULT);
     }
 
+    protected void refreshIndex(){
+        try {
+            client.indices().refresh(new RefreshRequest(whoisAliasIndex), RequestOptions.DEFAULT);
```
maggarwal13, line 136: Did you check the performance impact for this ? https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-refresh.html. I think issue is we forgot to set the query in countRequest if  that does not solve the problem I suggest instead of refreshing entries every 60 seconds may be have just a count compare once in 2-3 hours ?

whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticIndexService.java
```diff
@@ -92,7 +92,7 @@ public boolean isEnabled() {
         return true;
     }
 
-    protected void addEntry(final RpslObject rpslObject) throws IOException {
+    protected void updateIfExistCreateIfNot(final RpslObject rpslObject) throws IOException {
```
maggarwal13, line 95: rename to CreateOrUpdateEntry? Although  addEntry  is also correct as operation for both update and create is Update only

whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticIndexService.java
MiguelAHM: CountRequest does not have a setQuery, I checked the documentation and they have source for adding a SearchSourceBUilder which is deprecated
https://www.elastic.co/guide/en/elasticsearch/client/java-rest/current/java-rest-high-count.html
I have already tried this, it is not working. If this was the problem, the refresh will not work. The problem is that the index is not refreshed and is taking previous docs count because the loop that we have takes less than 1 second

whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticIndexService.java
MiguelAHM: If you check the output of the index is has two responses
CREATE or UPDATE
so I think that addEntry is not a good name, we are mixing things even if we are calling same service.
but I will do the change

whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticIndexService.java
MiguelAHM: Discussed via Mattermost, we will create a job that perform the check each hour

whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticIndexService.java
maggarwal13: Sory countRequest takes query builder.  countRequest.query(QueryBuilders.matchAllQuery());

MiguelAHM: ok, let me check. I did not try this

whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticIndexService.java
maggarwal13: No worries, I am ok with This name as well. I meant we are updating the entries via serial entry table and it has only two operation update and delete so update is both for create and update so we do not differentiate between them.

whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticIndexService.java
maggarwal13: Even if it does not work its best to set the queryBuilder in countRequest or check what is the default behaviour.

MiguelAHM: Same issue, I will leave it there.

whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticFullTextIndex.java
```diff
@@ -67,6 +68,38 @@ public void scheduledUpdate() {
...
+            alert();
+        } catch (DataAccessException | IOException | IllegalStateException e) {
+            LOGGER.error("Unable to perform fulltext index checks due to {}: {}", e.getClass(), e.getMessage());
+        }
+
+        LOGGER.info("Completed Elasticsearch index check");
+    }
+
+    protected void alert() throws IOException{
+        if (shouldRebuild()) {
```
maggarwal13, line 89: Why you need this condition? If ES is enabled just directly check the objects.

whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticFullTextIndex.java
```diff
@@ -67,6 +68,38 @@ public void scheduledUpdate() {
...
+            LOGGER.error("ES indexes needs to be rebuild");
+            return;
+        }
+
+        final Map<Integer, Integer> maxSerialIdWithObjectCount = serialDao.getMaxSerialIdWithObjectCount();
+        final int countInDb = ((int) maxSerialIdWithObjectCount.values().toArray()[0]);
```
maggarwal13, line 95: This will not work. As current serial Id in database might be behind in ES index. You need to get number of objects for a serialId saved in ES and then compare otherwise this will always result in error

whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticFullTextIndex.java
MiguelAHM: This will be true when there is not document in Elasticsearch, or no metadata or not serial.
In those cases I guess that making this check makes no sense

whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticFullTextIndex.java
```diff
@@ -67,6 +68,38 @@ public void scheduledUpdate() {
...
+        final long countInES = elasticIndexService.getWhoisDocCount();
+        if (countInES != countInDb) {
+            final int dbMaxSerialId = (Integer) maxSerialIdWithObjectCount.keySet().toArray()[0];
+            LOGGER.error(String.format("Number of objects in DB (%s) does not match to number of objects indexed in ES (%s) for serialId (%s)", countInDb, countInES, dbMaxSerialId));
```
maggarwal13, line 99: Don't you need to call refresh now? Else we will have same issue?

MiguelAHM: I need to take a coffee, yes, of course that I need to call refresh

whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticFullTextIndex.java
MiguelAHM: Updated

whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticFullTextIndex.java
```diff
@@ -33,7 +33,7 @@ public class ElasticFullTextIndex {
 
     @Autowired
     public ElasticFullTextIndex(final ElasticIndexService elasticIndexService,
-                                @Qualifier("jdbcSerialDao") final SerialDao serialDao,
+                                @Qualifier("jdbcSlaveSerialDao") final SerialDao serialDao,
```
eshryane, line 36: The `serialDao.getMaxSerialIdWithObjectCount()` method is annotated with a `REQUIRES_NEW` transaction, so may return a different serial than when using the `jdbcTemplate`. Maybe move that code into `JdbcRpslObjectOperations` so you can use the same `jdbcTemplate` instead ?

whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticFullTextIndex.java
```diff
@@ -92,30 +92,28 @@ protected void update() throws IOException {
...
             }
         }
 
         LOGGER.debug("Updated index in {}", stopwatch.stop());
+
+        elasticIndexService.refreshIndex();
+
         elasticIndexService.updateMetadata(new ElasticIndexMetadata(dbMaxSerialId, source));
 
         // One Object POEM-CDMA can not be parsed to RPSl so cannot be indexed
```
eshryane, line 115: Remove this comment as you're no longer subtracting one?

whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticFullTextIndex.java
MiguelAHM: I don't understand why is may return different serial.
We are just using jdbcTemplate once (JdbcSerialDao, line 61/62) so the `REQUIRES_NEW` should be fine

eshryane: A new transaction may give us a different view of the database, with different serials, if there have been any changes since the encompassing transaction was started

whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticIndexService.java
eshryane: @MiguelAHM tested that the performance impact of a refresh is minimal (less than 0.5s)

whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticFullTextIndex.java
MiguelAHM: we will ask Mahesh when he is back

eshryane (approved):

PR #1370, opened 2023-12-20
Fix Tests Broken by #1367 Refresh Elasticsearch Index

MiguelAHM:
Integration tests broken by #1367 merge.

1) Documentation link
2) Use slave datasource in ElasticFullTextIndex causes problems in the IT. "jdbcSlaveSerialDao" is using "whoisSlaveDataSource" which we have already injected in the ElasticFullTextIndex class. Solution:
- Stop using "SerialDao" and move "getMaxSerialIdWithObjectCount" to "JdbcRpslObjectOperations".

After talking with Ed we decided to go with this solution

### DIFF
--- a/whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticFullTextIndex.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticFullTextIndex.java
@@ -3,7 +3,6 @@ package net.ripe.db.whois.api.elasticsearch;
 import com.google.common.base.Stopwatch;
 import jakarta.annotation.PostConstruct;
 import net.javacrumbs.shedlock.spring.annotation.SchedulerLock;
-import net.ripe.db.whois.common.dao.SerialDao;
 import net.ripe.db.whois.common.dao.jdbc.JdbcRpslObjectOperations;
 import net.ripe.db.whois.common.domain.serials.SerialEntry;
 import net.ripe.db.whois.common.rpsl.RpslObject;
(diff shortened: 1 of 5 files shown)


eshryane:
I updated the PR title and pointed to the previous merged PR #1367 which broke the tests.

In general please make the PR title and description self-explanatory and as simple as possible, both for our users and for ourselves when we come across the PR in future!

whois-api/src/main/java/net/ripe/db/whois/api/elasticsearch/ElasticFullTextIndex.java
```diff
@@ -128,6 +128,15 @@ private SerialEntry getSerialEntry(final int serial) {
         }
     }
 
+    private Map<Integer, Integer> getMaxSerialIdWithObjectCount() {
+        try {
+            return JdbcRpslObjectOperations.getMaxSerialIdWithObjectCount(jdbcTemplate);
+        } catch (Exception e) {
+            LOGGER.debug("Caught exception reading max serial Id with object count", e);
+            return null;
```
eshryane, line 136: Returning `null` will cause a nullpointerexception on the next line after the method call. I suggest logging at ERROR and rethrowing the exception.

MiguelAHM: yes, you are right. It will do null.keyset() -> null pointed

MiguelAHM:
test ok
https://gitlab.ripe.net/swe-database-team/whois/-/pipelines/467219

eshryane (approved):

PR #1372, opened 2023-12-21
feat: use work instead of abuse for email type

MiguelAHM:


MiguelAHM:
We could break our users if we change the type of the abusemailbox.
The rfc says MAY about "work" and "home" so I understand that the type may have those values or could have another values.
Anyway, we decided to document the README with this difference

PR #1374, opened 2023-12-21
Put geofeed as a link to conform with RFC draft v1

MiguelAHM:
In order to meet the RFC https://datatracker.ietf.org/doc/draft-ietf-regext-rdap-geofeed/
We need to move the GeoFeed from the entity to a link. We need to modify the rdap conformance.

### DIFF
--- a/whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
@@ -108,6 +108,8 @@ import static net.ripe.db.whois.common.rpsl.ObjectType.INET6NUM;
 class RdapObjectMapper {
 
     private static final String TERMS_AND_CONDITIONS = "http://www.ripe.net/data-tools/support/documentation/terms";
+
+    private static final String GEOFEED_CONTENT_TYPE = "application/geofeed+csv";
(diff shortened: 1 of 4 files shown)


whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -693,11 +693,11 @@ private static String getAndHandleMultipleAttributes(final RpslObject rpslObject
...
     }
 
-    private static void setGeoFeed(final RpslObject rpslObject, final Ip ip) {
-        ip.getRdapConformance().add(GEO_FEED_V1.getValue());
+    private static void setGeoFeed(final RpslObject rpslObject, final Ip ip, final String requestUrl) {
+        ip.getRdapConformance().add(GEO_FEED_1.getValue());
 
         if(rpslObject.containsAttribute(GEOFEED)) {
-            ip.setGeofeedv1_geofeed(rpslObject.getValueForAttribute(GEOFEED).toString());
+            ip.getLinks().add(new Link(requestUrl, "geo", rpslObject.getValueForAttribute(GEOFEED).toString(), null, "application/geofeed+csv"));
```
eshryane, line 700: Create a static field for the content-type as it's used twice

eshryane (approved):

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java

PR #1386, opened 2024-01-23
Fix issue with Whois User Auth Service cache null input

MiguelAHM:
- We are caching user information. If that cache receives a null in the input, the cache will throw an error
- Added a condition to avoid the error and tackle the error inside the method
- Complete IT for Auth Service including error situations

### DIFF
--- a/whois-api/src/test/java/net/ripe/db/whois/api/sso/AuthServiceClientTestIntegration.java
+++ b/whois-api/src/test/java/net/ripe/db/whois/api/sso/AuthServiceClientTestIntegration.java
@@ -2,6 +2,7 @@ package net.ripe.db.whois.api.sso;
 
 import net.ripe.db.whois.api.AbstractIntegrationTest;
 import net.ripe.db.whois.common.sso.AuthServiceClient;
+import net.ripe.db.whois.common.sso.AuthServiceClientException;
 import net.ripe.db.whois.common.sso.domain.HistoricalUserResponse;
 import net.ripe.db.whois.common.sso.domain.ValidateTokenResponse;
 import org.junit.jupiter.api.BeforeEach;
@@ -16,11 +17,14 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.not;
 import static org.hamcrest.Matchers.nullValue;
+import static org.junit.Assert.assertThrows;
 
 @Tag("IntegrationTest")
 public class AuthServiceClientTestIntegration extends AbstractIntegrationTest {
 
     private static final String UUID = "8ffe29be-89ef-41c8-ba7f-0e1553a623e5";
+
+    private static final String USER_EMAIL = "test@ripe.net";
     @Autowired
     private CacheManager cacheManager;
     @Autowired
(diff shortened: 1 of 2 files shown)


whois-api/src/test/java/net/ripe/db/whois/api/sso/AuthServiceClientTestIntegration.java
```diff
@@ -61,4 +85,54 @@ public void get_historical_event_date_time() {
...
+
+    @Test
+    public void get_sso_uuid_response_null_input_then_error() {
+        assertThat(cacheManager.getCache("ssoUuid").get(USER_EMAIL), is(nullValue()));
+
+        final AuthServiceClientException authServiceClientException = assertThrows(AuthServiceClientException.class, () -> {
+            authServiceClient.getUuid(null);
+        });
+
+        assertThat(authServiceClientException.getMessage(), is("No UUID."));
```
maggarwal13, line 110: Should be no Username? as you are getting UUID from username so username can be null ? Are the tests passing?

whois-api/src/test/java/net/ripe/db/whois/api/sso/AuthServiceClientTestIntegration.java
```diff
@@ -61,4 +85,54 @@ public void get_historical_event_date_time() {
...
+
+    @Test
+    public void get_user_details_response_null_input_then_error() {
+        assertThat(cacheManager.getCache("ssoUserDetails").get(UUID), is(nullValue()));
+
+        final AuthServiceClientException authServiceClientException = assertThrows(AuthServiceClientException.class, () -> {
+            authServiceClient.getUserDetails(null);
+        });
+
+        assertThat(authServiceClientException.getMessage(), is("No username."));
```
maggarwal13, line 122: Should be no UUID? as its getting userdetails given UUID?

whois-api/src/test/java/net/ripe/db/whois/api/sso/AuthServiceClientTestIntegration.java
```diff
@@ -61,4 +85,54 @@ public void get_historical_event_date_time() {
...
+    // Errors
+    @Test
+    public void get_validate_token_response_null_input_then_error() {
+        assertThat(cacheManager.getCache("ssoValidateToken").get(UUID), is(nullValue()));
+
+        final AuthServiceClientException authServiceClientException = assertThrows(AuthServiceClientException.class, () -> {
+            authServiceClient.validateToken(null);
+        });
+
+        assertThat(authServiceClientException.getMessage(), is("No UUID."));
```
maggarwal13, line 98: Should be no token instead of no UUID?

whois-api/src/test/java/net/ripe/db/whois/api/sso/AuthServiceClientTestIntegration.java
MiguelAHM: you are right, thank you for point this out

whois-api/src/test/java/net/ripe/db/whois/api/sso/AuthServiceClientTestIntegration.java
MiguelAHM: I did not change the returned message of the other class (AuthService)
Here UUID and Token are the same. I can change it, but they are the same

whois-api/src/test/java/net/ripe/db/whois/api/sso/AuthServiceClientTestIntegration.java
MiguelAHM: Yes, I changed this one instead the previous one...I will update now with the correct changes

whois-api/src/test/java/net/ripe/db/whois/api/sso/AuthServiceClientTestIntegration.java
maggarwal13: UUID and token are not same. UUID is assigned against an email id and Token is used for authentication purposes. Correct me if I am wrong. Can you change it ?

MiguelAHM: You are right, the token is in the header as a bearer token and it is used for auth.
The UUID is the SSO identification for the user

I mean in tests code is the same, we are using same value for both (same value for variable UUID). I will refactor this, to use a different value, the test will be clear and return token instead uuid

MiguelAHM: Updated, now the Token looks like a token in the tests and not like an UUID

MiguelAHM:
test ok
https://gitlab.ripe.net/swe-database-team/whois/-/pipelines/489910

maggarwal13 (approved):

PR #1387, opened 2024-01-24
Remove error logs when 404 in RDAP

MiguelAHM:
- Remove logs added in #1267


whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapService.java
```diff
@@ -428,7 +428,6 @@ private Response redirect(final String requestPath, final Query query) {
         try {
             uri = delegatedStatsService.getUriForRedirect(requestPath, query);
         } catch (WebApplicationException e) {
-            LOGGER.error(e.getMessage(), e);
             throw new RdapException("404 Not found", "Redirect URI not found", HttpStatus.NOT_FOUND_404);
```
maggarwal13, line 431: Can you put logger.debug instead I am not sure in what all cases it throws this exception

maggarwal13 (approved):

PR #1389, opened 2024-01-31
Do not accept UUID values in SSO

MiguelAHM:
We send an error if an user tries to update the RIPE Database using UUID in the SSO instead the e-mail


eshryane:
Compare with previous branch: https://github.com/RIPE-NCC/whois/compare/mntner_sso_uuid...DB-1049-uuid-values

whois-update/src/main/java/net/ripe/db/whois/update/sso/SsoTranslator.java
```diff
@@ -29,12 +27,12 @@ public void populateCacheAuthToUsername(final UpdateContext updateContext, final
...
-            public RpslAttribute translate(final String authType, final String authToken, final RpslAttribute originalAttribute) {
+            public RpslAttribute translate(final String authType, final String uuid, final RpslAttribute originalAttribute) {
                 if (authType.equals("SSO")) {
-                    if (!updateContext.hasSsoTranslationResult(authToken)) {
+                    if (!updateContext.getSsoTranslation().containsUuid(uuid)) {
                         try {
-                            final String username = authServiceClient.getUsername(authToken);
-                            updateContext.addSsoTranslationResult(authToken, username);
+                            final String username = authServiceClient.getUsername(uuid);
+                            updateContext.getSsoTranslation().put(username, uuid);
```
eshryane, line 35: Do we need the extra `contains()` checks within the `put()` method? Is it necessary to prevent overwriting existing values? It seems excessive if there's already a contains() call first.

whois-client/src/main/java/net/ripe/db/whois/common/sso/SsoTranslation.java
```diff
@@ -0,0 +1,52 @@
+package net.ripe.db.whois.common.sso;
+
+
+import com.google.common.collect.Maps;
+
+import javax.annotation.Nullable;
+import java.util.Map;
+
+public class SsoTranslation {
```
maggarwal13, line 9: We already have a class for called SsoTranslator what is difference between this class? THis name is quite confusing with the other similar name

whois-client/src/main/java/net/ripe/db/whois/common/sso/SsoTranslation.java
```diff
@@ -0,0 +1,52 @@
...
+
+    private final Map<String, String> usernameToUuidCache = Maps.newHashMap();
+
+    public void put(final String username, final String uuid) {
+        if (!containsUsername(username) && !containsUuid(uuid)) {
```
maggarwal13, line 14: Correct me if I am wrong, This seems to me some kind of cache for uuid against username and vice-versa. If that is the case Authservice client is already using cache so why not directly call the Authservice client instead of maintaining another cache ?

MiguelAHM:
This is not ready for being reviewed yet, just create the PR and forgot to put the PR as a draft.
Maybe we don't need any change just the tests

whois-client/src/main/java/net/ripe/db/whois/common/sso/SsoTranslation.java
MiguelAHM: you are right, we are caching authservices
yes, this map of username and uuid is like a cache. However, I think that the main purpose is to detect whenever there is a duplicated SSO.
Example: 
auth: SSO a@ripe.net //uuid = 123
auth: SSO 123 //a@ripe.net

However, we don't really care about this, because if we receive an uuid we will thrown an exception. So I think that the main idea of using this is to detect this behaviour and to send (in the code of the old branch, was an error...which was breaking the syncupdates, returning 500...I updated that to send a warn instead...however in the standup you told me that we just want to not allow directly the UUID so return an ***Error in a synupdate response and 400 in the rest call)

So you are right but I think that the purpose of this is to cache and to avoid duplications between UUID and Email

whois-client/src/main/java/net/ripe/db/whois/common/sso/SsoTranslation.java
MiguelAHM: yes, this code was a bit tricky I had to do a draft in a paper to understand

whois-update/src/main/java/net/ripe/db/whois/update/sso/SsoTranslator.java
MiguelAHM: inside the put we are checking both the username and the uuid
here there is a contains to directly avoid doing rest of the code (the call to auth included) if that uuid already exist.
SO we have a first contains to check just the UUID because is the info that we have to avoid doing unnecessary code, and once we have both information username and uuid we check both

MiguelAHM: so, you are right we can leave just the contains inside the put() this will work.
However, there is the possibility that this code will be execute when the UUID already exist in the map

MiguelAHM:
not need of this changes.

We are already checking if the UUID has the correct syntax, and returning an error if not.
In the Integration test we are using a mocked auth service, so this behaviour is not easily reproducible with the current code.

PR #1390, opened 2024-02-01
Create secure connector for Client Certificate

MiguelAHM:


whois-api/src/main/java/net/ripe/db/whois/api/httpserver/JettyBootstrap.java
```diff
@@ -288,11 +301,10 @@ protected TrustManagerFactory getTrustManagerFactoryInstance() {
         sslContextFactory.setKeyStorePassword(whoisKeystore.getPassword());
         sslContextFactory.setCipherComparator(HTTP2Cipher.COMPARATOR);
 
-        // enable optional client certificates
-        sslContextFactory.setWantClientAuth(true);
+        sslContextFactory.setNeedClientAuth(isClientCertificate);
```
eshryane, line 304: Don't do this for both secure connectors, only for the one with the client cert auth, i.e.
```
if (isClientCertificate) {
        // accept self-signed client certificates for authentication
        sslContextFactory.setNeedClientAuth(true);
        sslContextFactory.setValidateCerts(false);
        sslContextFactory.setTrustAll(true);
}
```

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/RewriteEngine.java
```diff
@@ -27,18 +27,27 @@ public class RewriteEngine {
...
     private final String nonAuthSource;
 
+    private final String clientAuthHost;
+
     @Autowired
     public RewriteEngine(final @Value("${api.rest.baseurl}") String baseUrl,
                          final @Value("${whois.source}") String source,
-                         final @Value("${whois.nonauth.source}") String nonAuthSource) {
+                         final @Value("${whois.nonauth.source}") String nonAuthSource,
+                         final @Value("${api.client.auth.baseurl}") String clientAuthBaseUrl) {
```
eshryane, line 36: Make this String optional, i.e. it doesn't have to be set if `client.auth.enabled` is false. Also check for an empty String below.

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/RewriteEngine.java
```diff
@@ -54,6 +63,12 @@ public RewriteHandler getRewriteHandler() {
         rewriteHandler.addRule(restVirtualHostRule);
         restRedirectRules(restVirtualHostRule);
 
+        // Client Auth
+        VirtualHostRuleContainer clientAuthVirtualHostRule = new VirtualHostRuleContainer();
```
eshryane, line 67: Make this section conditional on `client.auth.enabled` is true

whois-commons/src/test/resources/whois.properties
```diff
@@ -19,6 +19,9 @@ dir.grs.import.download=var${jvmId:}/grs
 # API configuration
 api.rest.baseurl=http://rest.db.ripe.net
 
+# API Client Certificate Authentication
+api.client.auth.baseurl=https://rest-cert-dev.db.ripe.net
```
eshryane, line 23: This is an optional property, I suggest only setting it when `client.cert.auth.enabled` is true (i.e. in `AbstractClientCertificateIntegrationTest.java`)

whois-update/src/main/java/net/ripe/db/whois/update/authentication/credential/ClientCertificateCredentialValidator.java
```diff
@@ -36,7 +36,7 @@ public class ClientCertificateCredentialValidator implements CredentialValidator
     public ClientCertificateCredentialValidator(final RpslObjectDao rpslObjectDao,
                                                 final DateTimeProvider dateTimeProvider,
                                                 final LoggerContext loggerContext,
-                                                final @Value("${client.cert.auth.enabled:false}") boolean enabled) {
+                                                final @Value("${client.auth.enabled:false}") boolean enabled) {
```
eshryane, line 39: Remember to also add the new property `client.auth.enabled` to all the Whois property files and add a TODO to remove `client.cert.auth.enabled`

MiguelAHM: yes, I will leave both for now in all the env and put to remove client.cert.auth.enabled in the release notes.
So we can do it when we do the release. I think that TODO is not necessary if we have already that specified in the release notes.
What do you think?

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/JettyBootstrap.java
MiguelAHM: I wanted to leave the previous code as it is.
the only difference between the previous connector and the one with client certificate is the setNeedClientAuth

eshryane: As discussed these 3 lines go together, only do this for the client auth connector

MiguelAHM: updated

whois-update/src/main/java/net/ripe/db/whois/update/authentication/credential/ClientCertificateCredentialValidator.java
eshryane: OK! That's fine

whois-api/src/test/java/net/ripe/db/whois/api/httpserver/AbstractHttpsIntegrationTest.java
```diff
@@ -23,7 +23,7 @@ public static void disableHttps() {
     }
 
     public int getSecurePort() {
-        return jettyBootstrap.getSecurePort();
+        return jettyBootstrap.getClientAuthPort();
```
eshryane, line 26: Don't do this here - as the *client auth* connector will be used. Use the secure port for Https tests and override this method in `AbstractClientCertificateIntegrationTest` instead

eshryane: Or instead use `getClientCertificatePort()` and don't override or change  `getSecurePort()`

MiguelAHM: yes, I was testing something and I left that there by mistake.
It is already in AbstractClientCertificationIntegrationTests
Just forgot to turn that return back to the previous value

MiguelAHM:
tests ok
https://gitlab.ripe.net/swe-database-team/whois/-/pipelines/500316

eshryane (approved):

PR #1391, opened 2024-02-01
Use client ip flag as remote address from trusted source

maggarwal13:


whois-api/src/main/java/net/ripe/db/whois/api/httpserver/RemoteAddressCustomizer.java
```diff
@@ -3,34 +3,51 @@
...
     @Nullable
     private String getRemoteAddress(final Request request) {
+        final Interval ipResource = IpInterval.asIpInterval(InetAddresses.forString(request.getRemoteAddr()));
+
+        final String clientIp = request.getParameterMap().containsKey(QUERY_PARAM_CLIENT_IP) ? request.getParameter(QUERY_PARAM_CLIENT_IP) : null;
+        if(isTrusted(ipResource, trusted) && StringUtils.isNotEmpty(clientIp)) {
+            return clientIp;
+        }
+
         final Enumeration<String> headers = request.getHeaders(HttpHeaders.X_FORWARDED_FOR);
```
MiguelAHM, line 51: should we remove this? we are not longer be using X-Forwarded-For
or should we do it later on? Maybe create a task in the backlog to clean up this header?

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/RemoteAddressCustomizer.java
```diff
@@ -43,4 +60,12 @@ private String getRemoteAddress(final Request request) {
 
         return Iterables.getLast(COMMA_SPLITTER.split(header));
     }
+
+    private Set<Interval> getIntervals(final String[] trusted) {
+        return Arrays.stream(trusted).map(IpInterval::parse).collect(Collectors.toSet());
+    }
+
+    private boolean isTrusted(final Interval ipResource, final Set<Interval> ipRanges) {
+        return ipRanges.stream().anyMatch(ipRange -> ipRange.getClass().equals(ipResource.getClass()) && ipRange.contains(ipResource));
```
MiguelAHM, line 69: Do we need to review the class?
If one of them is an Ipv4 and the other is an ipv6 the contains will take care of it anyways.
and instead of the contains is should be equals right? in the end you are comparing an Interval with another Interval. You are not checking if the interval exists in one list

whois-api/src/test/java/net/ripe/db/whois/api/httpserver/RemoteAddrCustomizerNonTrustedTestIntegration.java
```diff
@@ -0,0 +1,38 @@
...
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsString;
+
+@Tag("IntegrationTest")
+public class RemoteAddrCustomizerNonTrustedTestIntegration extends AbstractIntegrationTest {
```
MiguelAHM, line 15: shall we merge this test case with "RemoteAddressCustomizerTestIntegration"?
It seems that in this test case we only can test one thing.
"the remote is not trusted then take the X_FORWARDED_FOR"
This make me think that maybe this should go in a more generic test class.

Btw we are going to remove X_FORWARDED_FOR, so do we need a test case for it? If we enable client auth X_FORWARDED_FOR will not longer be working...

whois-api/src/test/java/net/ripe/db/whois/api/httpserver/RemoteAddrCustomizerNonTrustedTestIntegration.java
```diff
@@ -0,0 +1,38 @@
...
+        System.setProperty("ipranges.trusted", "172.16.0.0/12");
+    }
+
+    @AfterAll
+    public static void clearProperty() {
+        System.clearProperty("ipranges.trusted");
+    }
+
+    @Test
+    public void help_client_ip_flag_trusted_ip() {
```
MiguelAHM, line 29: "help_client_ip_flag_non_trusted_ip" ??

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/RemoteAddressCustomizer.java
```diff
@@ -3,34 +3,51 @@
...
 
 import javax.annotation.Nullable;
 import java.net.InetSocketAddress;
+import java.util.Arrays;
 import java.util.Enumeration;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 /**
  * When HTTP requests are proxied via a loadbalancer, the X-Forwarded-For value will replace the request address.
```
eshryane, line 24: N.B. Be sure to update the class javadoc comment to describe `clientIp`

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/JettyBootstrap.java
```diff
@@ -223,7 +223,7 @@ private Connector createConnector(final Server server) {
         final HttpConfiguration httpConfiguration = new HttpConfiguration();
         if (isHttpProxy()) {
             // client address is set in X-Forwarded-For header by HTTP proxy
-            httpConfiguration.addCustomizer(new RemoteAddressCustomizer());
+            httpConfiguration.addCustomizer(new RemoteAddressCustomizer(trustedIpRanges));
```
eshryane, line 226: This `if` means that the `RemoteAddressCustomizer` is only used if the LB proxies the request (i.e. HTTPS is *off* in the backend).

If you want that class to always handle 'clientIp` then you need to move the LB check there also.

whois-api/src/test/java/net/ripe/db/whois/api/httpserver/RemoteAddrCustomizerNonTrustedTestIntegration.java
eshryane, line 15: Also add an integration test for HTTPS enabled mode (e.g. in `WhoisRestServiceHttpsTestIntegration`) where a `clientIp` is specified and not `X-Forwarded-For`

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/JettyBootstrap.java
maggarwal13: Thanks for pointing this out I missed this

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/RemoteAddressCustomizer.java
maggarwal13: Yes as its two seperate class for ipv4 and ipv6 else it can throw class cast exception

whois-api/src/test/java/net/ripe/db/whois/api/httpserver/RemoteAddrCustomizerNonTrustedTestIntegration.java
maggarwal13: We cannot  merge this test case with "RemoteAddressCustomizerTestIntegration" as trusted ipRanges are set during spring initialization through system.setProperty hence we need a separate test class to test this scenario

maggarwal13: We do not pass on the client ip incase of rest service for syncupdates we do so its easy to test that. I think we can test this scenario using request log . So may be better to have asomething similar to JettyRequestLogTestIntegration for https ?

eshryane: Agreed

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/RemoteAddressCustomizer.java
```diff
@@ -1,46 +1,55 @@
...
-        if (headers == null || !headers.hasMoreElements()) {
-            return request.getRemoteAddr();
-        }
+    private String getRemoteAddrForTrustedSource(final Request request, final String resourceAddr) {
+        final Interval ipResource = IpInterval.asIpInterval(InetAddresses.forString(resourceAddr));
 
-        final String header = headers.nextElement();
-        if (Strings.isNullOrEmpty(header)) {
-            return request.getRemoteAddr();
+        final String clientIp = request.getParameterMap().containsKey(QUERY_PARAM_CLIENT_IP) ? request.getParameter(QUERY_PARAM_CLIENT_IP) : null;
```
MiguelAHM, line 38: I have a question related to this changes.
I see that you are using hierarchy, being this class the parent class and RemoteAddresXCustomizer the child.
This change seems that belong to RemoteAddressHttpsCustomizer and not to the parent class.

Should this customize be in the children and RemoteAddressHttpCustomizer do what it is in master branch 
and just apply those changes to RemoteAdressHttpsCustomizer? or RemoteAdressHttpCustomizer need to be affected for this changes as well?

MiguelAHM: reading the story: We need to make sure that the coming IP is a trusted IP
so I guess that this is for both http or https

MiguelAHM (approved):

MiguelAHM:
test are green
https://gitlab.ripe.net/swe-database-team/whois/-/pipelines/508066

MiguelAHM:
test ok
https://gitlab.ripe.net/swe-database-team/whois/-/pipelines/509631

whois-api/src/test/java/net/ripe/db/whois/api/httpserver/ClientCertificateServiceTestIntegration.java
```diff
@@ -26,6 +29,7 @@ public void client_certificate() {
     }
 
     @Test
+    @RetryFor(value = IOException.class, attempts = 10, intervalMs = 10000)
```
eshryane, line 32: Why is a retry needed?

MiguelAHM: flaky tests, sometimes the server is not up and it return "unreachable server error"
This is now in master and it doesn't happen too often, but it happens

whois-api/src/test/java/net/ripe/db/whois/api/log/JettyRequestNonTrustedHttpsTestIntegration.java
```diff
@@ -0,0 +1,112 @@
...
+                .request()
+                .get(WhoisResources.class);
+
+        assertThat(getRequestLog(), startsWith("127.0.0.1"));
+    }
+
+    // helper methods
+
+    // Ref. https://logging.apache.org/log4j/2.x/manual/customconfig.html
+    private void addLog4jAppender() {
```
eshryane, line 83: Is this code duplicated? Move to a utility class or define methods in a Log interface?

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/RemoteAddressCustomizer.java
```diff
@@ -4,43 +4,104 @@
...
+            if (StringUtils.isNotEmpty(clientIp)) {
+                return clientIp;
+            }
+        } catch (BadMessageException ex){
+            //ignore
         }
+        return null;
+    }
+
+    private String formaliseAddress(final String address){
```
eshryane, line 66: Rename to `normaliseAddress` or `stripSquareBrackets`

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/RemoteAddressCustomizer.java
```diff
@@ -4,43 +4,104 @@
...
 
     private static final Splitter COMMA_SPLITTER = Splitter.on(',').omitEmptyStrings().trimResults();
+    public static final String QUERY_PARAM_CLIENT_IP = "clientIp";
+    private final Set<Interval> trusted;
+
+    // if client address is set in X-Forwarded-For header by HTTP proxy
+    private final boolean usingForwardedForHeader;
+
+    public RemoteAddressCustomizer(final String trustedIpRanges, final boolean xForwardedFor) {
+        this.trusted = getIntervals(trustedIpRanges.split(","));
```
eshryane, line 36: Use COMMA_SPLITTER that's already defined rather than String.split(",")

whois-api/src/test/java/net/ripe/db/whois/api/httpserver/ClientCertificateServiceTestIntegration.java
eshryane: OK. Please add a TODO comment, we could do something later like add a `Before` method to poll for the server coming up.

whois-api/src/test/java/net/ripe/db/whois/api/log/JettyRequestNonTrustedHttpsTestIntegration.java
MiguelAHM: we are using hierarchy in the tests so I would prefer moving it to AbstractIntegrationTest class and take it from there. Instead creating an static class, which I know that you don't like, or put something related to the tests in the main code

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/RemoteAddressCustomizer.java
MiguelAHM: I will but why not formalise???

whois-api/src/test/java/net/ripe/db/whois/api/log/JettyRequestNonTrustedHttpsTestIntegration.java
MiguelAHM: Done

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/RemoteAddressCustomizer.java
MiguelAHM: Done

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/RemoteAddressCustomizer.java
MiguelAHM: Done

whois-api/src/test/java/net/ripe/db/whois/api/httpserver/ClientCertificateServiceTestIntegration.java
MiguelAHM: Done

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/RemoteAddressCustomizer.java
eshryane: "formalise" isn't really the correct term, normalise is better (i.e. convert to a normal form) or be explicit about just removing square brackets. Thanks

eshryane (approved):

PR #1392, opened 2024-02-05
redirect system console to logger

eshryane:


eshryane:
Redirect only works for `System.out.println()` etc. which we don't use. Thread dump output still goes to the system log.

PR #1405, opened 2024-02-27
Allow unsubscribe option in delivered emails

MiguelAHM:
https://datatracker.ietf.org/doc/html/rfc8058
We follow this RFC to implement unsubscribed on-click feature in delivered emails from WHOIS

### DIFF
--- a/whois-update/src/main/java/net/ripe/db/whois/update/mail/MailGatewaySmtp.java
+++ b/whois-update/src/main/java/net/ripe/db/whois/update/mail/MailGatewaySmtp.java
@@ -1,5 +1,7 @@
 package net.ripe.db.whois.update.mail;
 
+import jakarta.mail.MessagingException;
+import jakarta.mail.internet.MimeMessage;
 import net.ripe.db.whois.common.Message;
 import net.ripe.db.whois.common.Messages;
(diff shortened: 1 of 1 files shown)


whois-update/src/main/java/net/ripe/db/whois/update/mail/MailGatewaySmtp.java
```diff
@@ -108,4 +109,11 @@ private void sendEmailAttempt(final String to, final String replyTo, final Strin
             }
         }
     }
+
+    private void setHeaders(final MimeMessage mimeMessage, final String messageId) throws MessagingException {
```
eshryane, line 113: Do we need a separate method?

MiguelAHM: I wanted to keep it separated. We don't need it but it is more clear at least for me. I prefer this rather than having a big method (sendEmailAttempt)
anyways this task is not ready yet, I need to create the messageId and I will do this in the other task about the bounce detection

eshryane: OK

eshryane (approved):

PR #1413, opened 2024-03-22
Use unlimited threads during builds. Also updated plugin versions.

eshryane:
Use unlimited threads during unit test build (only). Using only one thread makes running tests very slow.

I did not change the integration test configuration as using multiple threads per code causes unpredictable test failures.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -913,7 +913,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
-                <version>3.8.1</version>
+                <version>3.13.0</version>
                 <configuration>
                     <release>${java.version}</release>
                     <encoding>${project.build.sourceEncoding}</encoding>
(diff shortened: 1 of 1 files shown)


MiguelAHM (approved):

pom.xml
```diff
@@ -987,9 +988,10 @@
                         <exclude>${test.files.integration}</exclude>
                     </excludes>
                     <runOrder>random</runOrder>
+                    <parallel>true</parallel>
                     <forkCount>1C</forkCount>
-                    <threadCount>1</threadCount>
                     <reuseForks>true</reuseForks>
+                    <useUnlimitedThreads>true</useUnlimitedThreads>
```
maggarwal13, line 994: If I remeber correctly we used thread count as 1 as Tetss became unpredictable. Not sure though. Can we run tests multiple times  to see correct beaviour?

pom.xml
```diff
@@ -1012,9 +1015,10 @@
                         <exclude>${test.files.excludes}</exclude>
                     </excludes>
                     <excludedGroups>ManualTest</excludedGroups>
+                    <runOrder>random</runOrder>
+                    <parallel>true</parallel>
```
maggarwal13, line 1019: Same goes for this change

pom.xml
eshryane: Using only a single thread makes the unit test build very slow, particularly noticeable in whois-internal for some reason. I ran the build multiple times locally and in GitLab and all seems fine.

pom.xml
eshryane: The test was effectively `parallel` already, I just made it explicit

PR #1417, opened 2024-03-28
Handle Multiple Delivery Status Notification Failure Responses for the Same Email Address

MiguelAHM:


whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageService.java
```diff
@@ -53,7 +54,11 @@ public void verifyAndSetAsUndeliverable(final EmailMessageInfo message){
         }
 
         LOGGER.debug("Undeliverable message-id {} email {}", message.messageId(), StringUtils.join(message.emailAddresses(), ", "));
-        message.emailAddresses().forEach(email -> emailStatusDao.createEmailStatus(email, EmailStatus.UNDELIVERABLE));
+        try {
+            message.emailAddresses().forEach(email -> emailStatusDao.createEmailStatus(email, EmailStatus.UNDELIVERABLE));
+        } catch (DuplicateKeyException ex) {
```
maggarwal13, line 59: You need to handle the exception per email instead of all email , else it will skip creating other emails in list as soon as this exception is thrown

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageService.java
```diff
@@ -53,7 +54,11 @@ public void verifyAndSetAsUndeliverable(final EmailMessageInfo message){
...
+            LOGGER.warn("Email already exist in EmailStatus table {}", StringUtils.join(message.emailAddresses(), ", "), ex);
```
eshryane, line 60: Log at DEBUG or not at all, this is an expected situation and we don't need to take any action

MiguelAHM: I turned into a debug

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageService.java
```diff
@@ -53,7 +54,13 @@ public void verifyAndSetAsUndeliverable(final EmailMessageInfo message){
         }
 
         LOGGER.debug("Undeliverable message-id {} email {}", message.messageId(), StringUtils.join(message.emailAddresses(), ", "));
-        message.emailAddresses().forEach(email -> emailStatusDao.createEmailStatus(email, EmailStatus.UNDELIVERABLE));
+        message.emailAddresses().forEach(emails -> {
+            try {
+                message.emailAddresses().forEach(email -> emailStatusDao.createEmailStatus(email, EmailStatus.UNDELIVERABLE));
```
maggarwal13, line 59: Its a nested FOR loop. Can you create individual email status when you catch exception inside a FOR loop?

MiguelAHM: Copy paste issue, I was looping twice. It is not needed to loop twice.

eshryane (approved):

PR #1419, opened 2024-04-02
Roa Rpki Validator From Internal

MiguelAHM:


whois-query/src/main/java/net/ripe/db/whois/query/rpki/RoutinatorDataProvider.java
```diff
@@ -0,0 +1,46 @@
+package net.ripe.db.whois.query.rpki;
```
eshryane, line 1: Don't put these classes in `whois-query` that package is for querying *whois*, instead put them in `whois-commons`

whois-query/src/main/java/net/ripe/db/whois/query/rpki/RoutinatorDataProvider.java
```diff
@@ -0,0 +1,46 @@
...
+@Profile({WhoisProfile.DEPLOYED})
+public class RoutinatorDataProvider implements RpkiDataProvider{
+
+    private static final int CLIENT_CONNECT_TIMEOUT = 20_000;
+    private static final int CLIENT_READ_TIMEOUT = 120_000;
+
+    private final Client client;
+    private final String rpkiBaseUrl;
+
+    public RoutinatorDataProvider(@Value("${rpki.base.url}") final String rpkiBaseUrl) {
```
eshryane, line 26: This client becomes *mandatory* if you have to set `rpki.base.url`. Instead make it optional and fail if the client is used.

whois-query/src/main/java/net/ripe/db/whois/query/rpki/WhoisRoaChecker.java
```diff
@@ -0,0 +1,28 @@
...
+
+
+import net.ripe.db.whois.common.rpsl.RpslObject;
+
+import java.util.Map;
+import java.util.Optional;
+
+import static net.ripe.db.whois.query.rpki.ValidationStatus.INVALID;
+
+public class WhoisRoaChecker extends RpkiRoaChecker {
```
eshryane, line 11: Is this class a Component?

whois-query/src/main/java/net/ripe/db/whois/query/rpki/RpkiService.java
```diff
@@ -0,0 +1,72 @@
...
+import net.ripe.db.whois.common.ip.Ipv6Resource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class RpkiService {
```
eshryane, line 17: Is this class a Component?

eshryane:
One more thing, maybe reduce `roas.json` in size, it's 18MB

whois-query/src/main/java/net/ripe/db/whois/query/rpki/RpkiService.java
MiguelAHM: I would say Service

whois-query/src/main/java/net/ripe/db/whois/query/rpki/RoutinatorDataProvider.java
MiguelAHM: moved to whois-commons

whois-query/src/main/java/net/ripe/db/whois/query/rpki/RoutinatorDataProvider.java
MiguelAHM: I made it optional and return a null with an error log.
Upper class is throwing an error if it gets a null from that call.
I marked the method as nullable as well

whois-query/src/main/java/net/ripe/db/whois/query/rpki/WhoisRoaChecker.java
MiguelAHM: added a component

whois-query/src/main/java/net/ripe/db/whois/query/rpki/RoutinatorDataProvider.java
eshryane: Thanks

whois-query/src/main/java/net/ripe/db/whois/query/rpki/RpkiService.java
eshryane: Looks good thanks

eshryane (approved):

PR #1420, opened 2024-04-02
Allow MimeMessageHelper as parameter in MailGatewaySmtp

MiguelAHM:
Refactor MailGatewaySmtp for allowing previous configuration.
Whois should not care about other mail sender configuration, just add common configuration, add the message_id, send the mail and store in outgoing


MiguelAHM:
test ok
https://gitlab.ripe.net/swe-database-team/whois/-/pipelines/549323

eshryane (approved):

PR #1431, opened 2024-04-09
Allow Inverse Query by SPONSORING-ORG for either trusted and non trusted

MiguelAHM:


whois-api/src/test/java/net/ripe/db/whois/api/rest/TrustedQueryTestIntegration.java
```diff
@@ -86,18 +86,12 @@ public void setup() {
     // inverse lookup on sponsoring org attribute
 
     @Test
-    public void inverse_lookup_sponsoring_org_from_untrusted_range_returns_empty() {
+    public void inverse_lookup_sponsoring_org_from_untrusted_range_succeeds() {
```
eshryane, line 89: Move this test to a different Query Integration test class (alongside "org:" maybe?), as "trusted" or not is no longer relevant

whois-endtoend/src/test/java/net/ripe/db/whois/query/InverseQueryTestIntegration.java
```diff
@@ -216,6 +216,35 @@ public void inverse_mnt_ref_no_results() {
...
+                "descr:          description\n" +
+                "sponsoring-org: ORG-SPONSOR\n" +
+                "admin-c:        PP1-TEST\n" +
+                "tech-c:         PP1-TEST\n" +
+                "mnt-by:         OWNER-MNT\n" +
+                "source:         TEST\n");
+
+        final String response = query("-i sponsoring-org ORG-SPONSOR");
+
+        assertThat(response, containsString("ORG-SPONSOR"));
```
eshryane, line 245: Can you check for the aut-num `AS102` also in the response

whois-api/src/test/java/net/ripe/db/whois/api/rest/TrustedQueryTestIntegration.java
MiguelAHM: I have already WhoisSearchServiceTestIntegration.search_inverse_sponsoring_org
So I will just remove it...we will have two test doing the same otherwise

whois-endtoend/src/test/java/net/ripe/db/whois/query/InverseQueryTestIntegration.java
MiguelAHM: Done

eshryane (approved):

PR #1434, opened 2024-04-10
Enable parallel tests but only one thread per core and eight forks maximum

eshryane:
Help limit impact on GitLab runner (many parallel jobs kills the runner).


MiguelAHM:
It seems that there is some kind of issue with groovy tests

eshryane:
> It seems that there is some kind of issue with groovy tests

whois-api and whois-groovy modules are failing in GitLab, seems on multiple branches not just this one.

PR #1437, opened 2024-04-10
Move Business Logic from Dao to Service And Case Insensitive Test

MiguelAHM:
Move business logic from DAO to service for outgoing check

### DIFF
--- a/whois-commons/src/main/java/net/ripe/db/whois/common/dao/OutgoingMessageDao.java
+++ b/whois-commons/src/main/java/net/ripe/db/whois/common/dao/OutgoingMessageDao.java
@@ -30,11 +30,4 @@ public class OutgoingMessageDao {
                 "SELECT email from outgoing_message where message_id = ?",
                 String.class, messageId);
     }
-
-    public boolean isEmailExists(final String messageId, final String email){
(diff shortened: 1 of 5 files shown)


whois-commons/src/main/java/net/ripe/db/whois/common/dao/OutgoingMessageDao.java
```diff
@@ -30,11 +30,4 @@ public List<String> getEmails(final String messageId) {
                 "SELECT email from outgoing_message where message_id = ?",
                 String.class, messageId);
     }
-
-    public boolean isEmailExists(final String messageId, final String email){
-        return Boolean.TRUE.equals(jdbcTemplate.query("select email from outgoing_message where message_id = ? and email = ?",
```
eshryane, line 35: Isn't this WHERE clause case-insensitive already?

MiguelAHM: yes, it is. I will change the name of the PR...In the end is just a little refactor to remove the logic from Dao

eshryane: Is this what we want? We could leave the SELECT to match the message-id and email address

eshryane (approved):

PR #1438, opened 2024-04-10
Always return an instance name (was failing locally when running tests)

eshryane:


maggarwal13 (approved):

eshryane:
Tested whois-api integration tests all pass now locally

PR #1442, opened 2024-04-11
set test elasticsearch memory limit from 1GB initially to 4GB max

eshryane:


fmurodov:
[The minimum and maximum values must be the same](https://www.elastic.co/guide/en/elasticsearch/reference/current/advanced-configuration.html#set-jvm-heap-size)

fmurodov (commented):
[The minimum and maximum values must be the same.](https://www.elastic.co/guide/en/elasticsearch/reference/current/advanced-configuration.html#set-jvm-heap-size)

fmurodov:
> > [The minimum and maximum values must be the same.](https://www.elastic.co/guide/en/elasticsearch/reference/current/advanced-configuration.html#set-jvm-heap-size)
> 
> This is for running ES in production I think and not for tests?

There might be an issue during the test as well, as it throw warning on startup
`"o.e.b.BootstrapChecks", "cluster.name": "docker-cluster", "node.name": "dd0cd17874e1", "message": "initial heap size [1073741824] not equal to maximum heap size [4294967296]; this can cause resize pauses" }`

eshryane:
> > > [The minimum and maximum values must be the same.](https://www.elastic.co/guide/en/elasticsearch/reference/current/advanced-configuration.html#set-jvm-heap-size)
> > 
> > 
> > This is for running ES in production I think and not for tests?
> 
> There might be an issue during the test as well, as it throw warning on startup `"o.e.b.BootstrapChecks", "cluster.name": "docker-cluster", "node.name": "dd0cd17874e1", "message": "initial heap size [1073741824] not equal to maximum heap size [4294967296]; this can cause resize pauses" }`

Thanks for pointing that out. I suggest we try these settings in GitLab to see if tests pass, and whether any resize pauses are noticable.

PR #1443, opened 2024-04-11
Fix inetnum spec test

eshryane:


maggarwal13 (approved):
Thanks for fixing this on my behalf :)

PR #1416, opened 2024-03-28
Route ROA checker

MiguelAHM:

PR #1444, opened 2024-04-12
Schedule ROA load each 15 minutes

MiguelAHM:
Depends on https://github.com/RIPE-NCC/whois/pull/1416

### DIFF
--- a/whois-commons/src/main/java/net/ripe/db/whois/common/rpki/RpkiService.java
+++ b/whois-commons/src/main/java/net/ripe/db/whois/common/rpki/RpkiService.java
@@ -8,6 +8,7 @@ import net.ripe.db.whois.common.ip.Ipv4Resource;
 import net.ripe.db.whois.common.ip.Ipv6Resource;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.stereotype.Component;
 import org.springframework.stereotype.Service;
 
@@ -24,7 +25,15 @@ public class RpkiService {
     private final NestedIntervalMap<Ipv4Resource, Set<Roa>> ipv4Tree = new NestedIntervalMap<>();
     private final NestedIntervalMap<Ipv6Resource, Set<Roa>> ipv6Tree = new NestedIntervalMap<>();
 
+    private final RpkiDataProvider rpkiDataProvider;
+
     public RpkiService(final RpkiDataProvider rpkiDataProvider) {
+        this.rpkiDataProvider = rpkiDataProvider;
+        loadRoas();
+    }
+
+    @Scheduled(cron = "*/15 * * * *")
+    private void loadRoas() {
         final List<Roa> loadedRoas = rpkiDataProvider.loadRoas();
         if (loadedRoas != null && !loadedRoas.isEmpty()){
             final List<Roa> roas = loadedRoas.stream()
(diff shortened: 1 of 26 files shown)


whois-commons/src/main/java/net/ripe/db/whois/common/rpki/RpkiService.java
```diff
@@ -20,24 +23,35 @@ public class RpkiService {
...
+
+    @Scheduled(cron = "* /15 * * * *")
+    private void loadRoas() {
         final List<Roa> loadedRoas = rpkiDataProvider.loadRoas();
         if (loadedRoas != null && !loadedRoas.isEmpty()){
             final List<Roa> roas = loadedRoas.stream()
                     .filter(roa -> roa.getTrustAnchor() != TrustAnchor.UNSUPPORTED)
                     .collect(Collectors.toList());
 
+            ipv4Tree.clear();
```
eshryane, line 44: Asynchronous queries to the tree will find nothing after `clear()`. Maybe generate *new* ipv4Tree and ipv6Tree instances from scratch, and then assign them to the fields so there is no gap. Similar to `IpTreeCacheManager`

MiguelAHM: Done

whois-commons/src/main/java/net/ripe/db/whois/common/rpki/RpkiService.java
```diff
@@ -20,16 +21,27 @@ public class RpkiService {
...
+
+    @Scheduled(cron = "0 */15 * * * *")
+    private void loadRoas() {
         final List<Roa> loadedRoas = rpkiDataProvider.loadRoas();
         if (loadedRoas != null && !loadedRoas.isEmpty()){
             final List<Roa> roas = loadedRoas.stream()
                     .filter(roa -> roa.getTrustAnchor() != TrustAnchor.UNSUPPORTED)
                     .collect(Collectors.toList());
 
+            ipv4Tree = new NestedIntervalMap<>();
```
eshryane, line 42: This is still clearing the map until it is updated, which will cause problems for asynchronous queries. 
Could you split the update by tree, e.g.
```
this.ipv4Tree = buildTree(...);
this.ipv6Tree = buildTree(...);
```
and within buildTree create a new NestedIntervalMap and return it? Then the map instance is replaced in one operation.

eshryane (approved):

whois-commons/src/main/java/net/ripe/db/whois/common/rpki/RpkiService.java
```diff
@@ -20,31 +21,43 @@ public class RpkiService {
...
         final List<Roa> loadedRoas = rpkiDataProvider.loadRoas();
         if (loadedRoas != null && !loadedRoas.isEmpty()){
             final List<Roa> roas = loadedRoas.stream()
                     .filter(roa -> roa.getTrustAnchor() != TrustAnchor.UNSUPPORTED)
                     .collect(Collectors.toList());
 
             LOGGER.info("downloaded {} roas from rpki", roas.size());
-            buildTrees(roas, ipv4Tree, ipv6Tree);
+
+            final Pair<NestedIntervalMap<Ipv4Resource, Set<Roa>>, NestedIntervalMap<Ipv6Resource, Set<Roa>>> trees = buildTree(roas);
```
eshryane, line 44: No need to return anything from `buildTrees` if you want to update *both* fields, do it directly in that method?

whois-commons/src/main/java/net/ripe/db/whois/common/rpki/RpkiService.java
MiguelAHM: talk via Mattermost. I will try to avoid looping twice same list.
So we will go with overlapping those variables but in the method that just loops once

whois-commons/src/main/java/net/ripe/db/whois/common/rpki/RpkiService.java
MiguelAHM: yes, you are right. I don't need to return a pair.
I did update this

PR #1453, opened 2024-04-25
Fixed Syncupdates HELP Bug

eshryane:
Pull Request "Use client ip flag as remote address from trusted source" (#1391) introduced a bug where HELP text is returned even if a request contains data. This is because a Rewrite Regex Rule re-wrote an empty path to include the HELP command by default. This is now changed to check for HELP or DATA within the service class itself, and return HELP text only if there is no data.

### DIFF
--- a/whois-api/src/main/java/net/ripe/db/whois/api/httpserver/RewriteEngine.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/httpserver/RewriteEngine.java
@@ -78,7 +78,6 @@ public class RewriteEngine {
         final VirtualHostRuleContainer rdapVirtualHostRule = new VirtualHostRuleContainer();
         rdapVirtualHostRule.addVirtualHost(rdapVirtualHost);
         rewriteHandler.addRule(rdapVirtualHostRule);
-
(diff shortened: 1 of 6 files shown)


whois-api/src/main/java/net/ripe/db/whois/api/httpserver/RewriteEngine.java
```diff
@@ -87,19 +86,11 @@ public RewriteHandler getRewriteHandler() {
         final VirtualHostRuleContainer syncupdatesVirtualHostRule = new VirtualHostRuleContainer();
         syncupdatesVirtualHostRule.addVirtualHost(syncupdatesVirtualHost);
         rewriteHandler.addRule(syncupdatesVirtualHostRule);
-
-        final RewriteRegexRule syncupdatesEmptyQueryStringRule = new RewriteRegexRule(
```
MiguelAHM, line 91: I am checking the history of GIT and this has been there from 05/2022
Do you know why it started breaking now?

MiguelAHM: ok, it is in the description

MiguelAHM (approved):

PR #1454, opened 2024-04-29
Refactor: Telnet Client must not be used as Static class

MiguelAHM:


MiguelAHM:
I will just do a little refactor of the constructors

PR #1460, opened 2024-05-08
Allow custom ROAs to be loaded in TEST

eshryane:
Moved ROA code from #1458 . Also updated RpkiServiceTest.

### DIFF
--- a/whois-commons/src/test/java/net/ripe/db/whois/common/rpki/DummyRpkiDataProvider.java
+++ b/whois-commons/src/test/java/net/ripe/db/whois/common/rpki/DummyRpkiDataProvider.java
@@ -6,21 +6,32 @@ import org.springframework.context.annotation.Profile;
 import org.springframework.stereotype.Component;
 
 import java.io.IOException;
-import java.util.Collections;
 import java.util.List;
 
 @Component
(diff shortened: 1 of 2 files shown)


whois-commons/src/test/java/net/ripe/db/whois/common/rpki/DummyRpkiDataProvider.java
```diff
@@ -13,8 +13,14 @@
...
+    private List<Roa> loadedRoas;
+
     @Override
     public List<Roa> loadRoas() {
+        if (loadedRoas!= null && !loadedRoas.isEmpty()){
+            return loadedRoas;
+        }
+
         try {
             return new ObjectMapper().readValue(getClass().getResourceAsStream("/rpki/roas.json"), Roas.class).getRoas();
```
eshryane, line 25: Rename `loadCustomRoas(final List<Roa> roas)` to `loadRoas(final List<Roa> roas)` to overload the existing method. Also call `loadRoas(final List<Roa> roas)` with the `ObjectMapper()...getRoas()` and put all logic into *one* method.

MiguelAHM:
Done in a different PR

PR #1474, opened 2024-06-11
catch exception if rpki unavailable

maggarwal13:


whois-commons/src/main/java/net/ripe/db/whois/common/rpki/RoutinatorDataProvider.java
```diff
@@ -56,8 +56,8 @@ public List<Roa> loadRoas() {
                     .request(MediaType.APPLICATION_JSON_TYPE)
                     .get(Roas.class)
                     .getRoas();
-        } catch (ClientErrorException ex){
-            LOGGER.error("RPKI service returned an error, so the ROAs are not updated");
+        } catch (Exception ex){
+            LOGGER.error("RPKI service returned an error, so the ROAs are not updated", ex);
```
eshryane, line 60: Hmm what happens if Whois already has a full list of ROAs and then there is an error? The existing list will be wiped out. Maybe we should instead catch the exception *outside* the DataProvider, i.e. in the Rpki Service, and instead warn there instead of clearing the list

eshryane (approved):

PR #1476, opened 2024-06-11
Use hazelcast as cache manager for tests

maggarwal13:
We have been using Hazelcast as the deployed cache manager, but a simple cache manager for tests.

We should also use Hazelcast as the cache manager for tests, so we catch any issues earlier.


maggarwal13:
Need to run successfully after multiple runs

PR #1477, opened 2024-06-12
downgrade spring 6.1.4

maggarwal13:


whois-commons/src/main/resources/applicationContext-commons.xml
```diff
@@ -21,7 +21,7 @@
     <context:component-scan base-package="net.ripe.db.whois.common"/>
 
     <context:annotation-config/>
-    <cache:annotation-driven mode="aspectj"/>
```
eshryane, line 24: Do we not still need this?

whois-scheduler/src/main/java/net/ripe/db/whois/scheduler/SchedulerConfig.java
```diff
@@ -17,9 +18,9 @@
 import javax.sql.DataSource;
 
 @Configuration
-// @EnableAspectJAutoProxy(proxyTargetClass = true)
+@EnableAspectJAutoProxy(proxyTargetClass = true)
 @EnableScheduling
-@EnableSchedulerLock(defaultLockAtMostFor="23h", mode = AdviceMode.ASPECTJ)
```
eshryane, line 22: Also is it better to be explicit?

whois-commons/src/main/resources/applicationContext-commons.xml
maggarwal13: I wanted it to be as it was before spring upgrade.

whois-scheduler/src/main/java/net/ripe/db/whois/scheduler/SchedulerConfig.java
maggarwal13: I wanted it to be as it was before spring upgrade.

whois-commons/src/main/resources/applicationContext-commons.xml
eshryane: I think it's safer to be explicit, we weren't doing that before, which means we may have mixed AspectJ weaving and proxying. We should specify aspectj mode everywhere.

eshryane: Have a read of the issue https://github.com/spring-projects/spring-framework/issues/32882

"I still recommend revising your application setup towards exclusive weaving usage (no proxy setup at all)."

I think we're in the same situation.

eshryane: Agreed lets downgrade completely and add mode later when we re-upgrade

eshryane (approved):

whois-commons/src/main/resources/applicationContext-commons.xml
maggarwal13: Yes,  lets do it as part of spring upgrade.Thanks

PR #1484, opened 2024-06-17
NRTMv4 Key Rotation

maggarwal13:


whois-nrtm4/src/main/java/net/ripe/db/nrtm4/generator/NrtmKeyPairService.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        if(nextKey.size() != 1) {
+            throw new RuntimeException("More than two next key pair record found , this cannot happen");
+        }
+
+        return nextKey.get(0);
+    }
+
+    public NrtmKeyRecord generateOrRotateNextKey() {
```
eshryane, line 76: Can you add some logging in each if statement to confirm what happened?

eshryane, line 76: Add Nullable annotation if null can be returned

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/UpdateNrtmFileRepository.java
```diff
@@ -66,6 +68,15 @@ public void saveSnapshotVersion(final NrtmVersionInfo version, final String file
         LOGGER.info("Created {} snapshot version {}", version.source().getName(), version.version());
     }
 
+    public void rotateKey(final NrtmKeyRecord newActiveKey, final NrtmKeyRecord oldActiveKey) throws JsonProcessingException {
+
+        jdbcTemplate.update("UPDATE key_pair SET isActive = ? where id = ?", false, oldActiveKey.id());
```
eshryane, line 73: You can do this in a single statement (i.e. so it succeeds or fails together)
```
UPDATE keypair k1 JOIN keypair k2
SET k1.isActive = ?, k2.isActive = ?
WHERE k1.id = ? 
AND k2.id = ?
```

whois-commons/src/main/resources/nrtm_schema.sql
```diff
@@ -94,6 +94,7 @@ CREATE TABLE `key_pair`
     `public_key`  VARBINARY(3000) NOT NULL,
     `created`     bigint unsigned NOT NULL,
     `expires`     bigint unsigned NOT NULL,
+    `isActive`    bit(1) NOT NULL,
```
eshryane, line 97: Can you rename to `is_active` to be consistent with how the other columns are named?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/UpdateNrtmFileRepository.java
```diff
@@ -66,6 +68,15 @@ public void saveSnapshotVersion(final NrtmVersionInfo version, final String file
         LOGGER.info("Created {} snapshot version {}", version.source().getName(), version.version());
     }
 
+    public void rotateKey(final NrtmKeyRecord newActiveKey, final NrtmKeyRecord oldActiveKey) throws JsonProcessingException {
```
eshryane, line 71: Can you add a JMX managed operation to invoke this method (including generating a new key)

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/NrtmKeyConfigDao.java
```diff
@@ -23,29 +25,51 @@ public class NrtmKeyConfigDao {
         this.writeTemplate = new JdbcTemplate(writeDataSource);
     }
 
-    public byte[] getPrivateKey() {
-        return readTemplate.queryForObject("SELECT private_key FROM key_pair", (rs, rowNum) -> rs.getBytes(1));
+    public byte[] getActivePrivateKey() {
+        return readTemplate.queryForObject("SELECT private_key FROM key_pair where isActive=true", (rs, rowNum) -> rs.getBytes(1));
```
MiguelAHM, line 29: Forward the change of the column rename:
is_active

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/NrtmKeyConfigDao.java
```diff
@@ -1,13 +1,15 @@
...
 import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.stereotype.Repository;
 
 import javax.sql.DataSource;
-import java.time.ZoneOffset;
+import java.util.List;
+import java.util.stream.Collectors;
 
 @Repository
 public class NrtmKeyConfigDao {
```
MiguelAHM, line 15: DateTimeProvider is injected and never used in this class

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/NrtmKeyConfigDao.java
```diff
@@ -23,29 +25,51 @@ public class NrtmKeyConfigDao {
         this.writeTemplate = new JdbcTemplate(writeDataSource);
     }
 
-    public byte[] getPrivateKey() {
-        return readTemplate.queryForObject("SELECT private_key FROM key_pair", (rs, rowNum) -> rs.getBytes(1));
+    public byte[] getActivePrivateKey() {
```
MiguelAHM, line 28: Do we need to check if getActivePrivateKey and getActivePublicKey return just one row??
I mean, we shouldn't have 2 active keys but in case that we have a bug and there are two active keys in database...we maybe have a problem here?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/UpdateNrtmFileRepository.java
MiguelAHM, line 71: throws JsonProcessingException no needed, right?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/generator/NrtmKeyPairService.java
```diff
@@ -0,0 +1,104 @@
...
+@Service
+public class NrtmKeyPairService {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(NrtmKeyPairService.class);
+
+    private final NrtmKeyConfigDao nrtmKeyConfigDao;
+    private final DateTimeProvider dateTimeProvider;
+    private final UpdateNrtmFileRepository updateNrtmFileRepository;
+
+
```
MiguelAHM, line 30: Can you remove some empty line??

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/generator/NrtmKeyPairService.java
```diff
@@ -0,0 +1,104 @@
...
+        final List<NrtmKeyRecord> nextKey = nrtmKeyConfigDao.getAllKeyPair().stream().filter(keyRecord -> keyRecord.expires() > currentActiveKey.expires()).collect(Collectors.toList());
+        if(nextKey.isEmpty()) {
+            return null;
+        }
+
+        if(nextKey.size() != 1) {
+            throw new RuntimeException("More than two next key pair record found , this cannot happen");
+        }
+
+        return nextKey.get(0);
```
MiguelAHM, line 73: here you can use the new feature of Java 21
nextKey.getFirst()

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/generator/NrtmKeyPairService.java
```diff
@@ -0,0 +1,104 @@
...
+        final NrtmKeyRecord keyRecord = NrtmKeyRecord.of(privateKey, publicKey, isActive, createdTimestamp, expires );
+        nrtmKeyConfigDao.saveKeyPair(keyRecord);
+
+        return keyRecord;
+    }
+
+    private NrtmKeyRecord getNextkeyPairRecord() {
+        final NrtmKeyRecord currentActiveKey = nrtmKeyConfigDao.getActiveKeyPair();
+
+        final List<NrtmKeyRecord> nextKey = nrtmKeyConfigDao.getAllKeyPair().stream().filter(keyRecord -> keyRecord.expires() > currentActiveKey.expires()).collect(Collectors.toList());
```
MiguelAHM, line 64: collect(Collectors.toList());
By just toList() ???

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/generator/NrtmKeyPairService.java
```diff
@@ -0,0 +1,104 @@
...
+        final long createdTimestamp = dateTimeProvider.getCurrentDateTime().toEpochSecond(ZoneOffset.UTC);
+        final long expires = dateTimeProvider.getCurrentDateTime().plusYears(1).toEpochSecond(ZoneOffset.UTC);
+
+        final NrtmKeyRecord keyRecord = NrtmKeyRecord.of(privateKey, publicKey, isActive, createdTimestamp, expires );
+        nrtmKeyConfigDao.saveKeyPair(keyRecord);
+
+        return keyRecord;
+    }
+
+    private NrtmKeyRecord getNextkeyPairRecord() {
```
MiguelAHM, line 61: getNextKeyPair instead of "getNextkey"

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/generator/UpdateNotificationFileGenerator.java
```diff
@@ -78,7 +85,10 @@ public void generateFile() {
 
           final List<DeltaFileVersionInfo> deltaFiles = deltaFileDao.getAllDeltasForSourceSince(nrtmSource, oneDayAgo);
           final NrtmVersionInfo fileVersion = getVersion(deltaFiles, snapshotFile.get());
-          final String json = getPayload(snapshotFile.get(), deltaFiles, fileVersion, createdTimestamp);
+
+          final NrtmKeyRecord nextKey =  nrtmKeyPairService.generateOrRotateNextKey();
+
+           final String json = getPayload(snapshotFile.get(), deltaFiles, fileVersion, nextKey, createdTimestamp);
```
MiguelAHM, line 91: Indent issue?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/util/Ed25519Util.java
```diff
@@ -43,4 +44,18 @@ public static boolean verifySignature(final String signature, final byte[] publi
         verifier.update(contents, 0, contents.length);
         return verifier.verifySignature(Base64.getDecoder().decode(signature));
     }
+
+    public static String encodePublicKey(final NrtmKeyRecord keyRecord) {
```
MiguelAHM, line 48: Add nullable annotation?

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/UpdateNrtmFileRepository.java
MiguelAHM: afaik UpdateNotificationFIle will be executed an we will take care of the rotation of the key
Check:
UpdateNotificationFileGenerationTestIntegration.should_add_next_signing_key-in_notification_file()
This will call nrtmKeyPairService.generateOrRotateNextKey(); which will take care if the key need to be rotated or not.

MiguelAHM: Talked to Ed:
We need to do the rollover at ANY time...so in that case we need the JMX.
We are doing the rollover in an scheduler task right now

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/generator/UpdateNotificationFileGenerator.java
```diff
@@ -78,7 +85,10 @@ public void generateFile() {
 
           final List<DeltaFileVersionInfo> deltaFiles = deltaFileDao.getAllDeltasForSourceSince(nrtmSource, oneDayAgo);
           final NrtmVersionInfo fileVersion = getVersion(deltaFiles, snapshotFile.get());
-          final String json = getPayload(snapshotFile.get(), deltaFiles, fileVersion, createdTimestamp);
+
+          final NrtmKeyRecord nextKey =  nrtmKeyPairService.generateOrRotateNextKey();
```
eshryane, line 89: Can you move key rotation into a separate daily scheduled task? It's not related to the Update Notification File generator.

eshryane: Also add WARN logging so the team is aware that a key rotation is happening

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/UpdateNrtmFileRepository.java
maggarwal13: Adding it in a JMX will not help in case of key compromise as spec says client needs atleast 7 days for rotation. In case of compromise we need manual intervention

eshryane: I want an easy way to trigger a rotation, without needing to update the database manually, can this be done with JMX?

maggarwal13: This job creates a next signing key or rotate based on RFC spec and based on specific timeline. In case of compromise and  need to rotate key a Seperate JMX call and a new function to execute should be created and not use the same task. and the we also need someway to tell client somehow  about instantly using new key  as RFC says 7 days to rotate

eshryane: I want a way to manually trigger a rollover to a new signing key, whenever we like, and I don't want to touch the database manually. For example, once this functionality goes live, we should trigger a rollover once everyone is ready.

We should also create a separate function to trigger an instant rollover without waiting for 7 days, but again we shouldn't have to touch the database manually for that either.

maggarwal13: I will create a two JMX call, one to rotate the next key and one to create the next key  and rotate.

whois-nrtm4/src/main/java/net/ripe/db/nrtm4/dao/NrtmKeyConfigDao.java
maggarwal13: Yes in that case it will throw an exception

whois-commons/src/main/resources/nrtm_schema.sql
```diff
@@ -94,6 +94,7 @@ CREATE TABLE `key_pair`
     `public_key`  VARBINARY(3000) NOT NULL,
     `created`     bigint unsigned NOT NULL,
     `expires`     bigint unsigned NOT NULL,
+    `is_active`    bit(1) NOT NULL,
```
eshryane, line 97: Add `DEFAULT b'0'` present in patch but missing from schema

eshryane (approved):

PR #1486, opened 2024-06-24
Validate References

eshryane:
Delete when the object has more than X references is out of scope of this task. Read the Jira for more info


MiguelAHM (approved):

MiguelAHM:
I recommend this tests:
- When the property is 0
- When one attribute has 50 references and another 51

I HARDLY recommend this tests:
- When the mntner password is wrong and we have more than 100 references. We should get the multiple references error and not the auth error...because the references check is doing before the auth check

eshryane:
> I HIGHLY recommend this test:
> 
> * When the mntner password is wrong and we have more than 100 references. We should get the multiple references error and not the auth error...because the references check is doing before the auth check

I added this test

eshryane:
Integration tests passed at af13ed95

PR #1490, opened 2024-06-25
Handle Multipart/mixed Bounces and Auto-submitted Messages

eshryane:
Look for automated responses in plaintext mail messages and do not handle them as Whois updates.

Multipart/mixed messages may also contain a delivery failure report.

### DIFF
--- a/whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/AutoSubmittedMessageParser.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/AutoSubmittedMessageParser.java
@@ -14,8 +14,8 @@ import org.springframework.stereotype.Component;
 import javax.annotation.Nullable;
 import java.util.Collections;
 
-// Detect automated responses and mark them for deletion. Do not try to parse them as Whois updates.
-// TODO: attempt to find the outgoing message-id and failed recipient if possible
+// Detect automated responses and mark them for deletion. Do not try to parse them as Whois updates (for now).
+// TODO: attempt to find the outgoing message-id and failed recipient if possible by parsing the plaintext body
 @Component
 public class AutoSubmittedMessageParser {
 
@@ -25,7 +25,7 @@ public class AutoSubmittedMessageParser {
 
     @Autowired
     public AutoSubmittedMessageParser(@Value("${mail.smtp.from:}") final String smtpFrom) {
-        this.enabled = ! Strings.isNullOrEmpty(smtpFrom);
+        this.enabled = !Strings.isNullOrEmpty(smtpFrom);
     }
 
     @Nullable
--- a/whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/AutoSubmittedMessageParser.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/AutoSubmittedMessageParser.java
@@ -1,0 +1,64 @@
+package net.ripe.db.whois.api.mail.dequeue;
+
+import jakarta.mail.MessagingException;
+import jakarta.mail.internet.MimeMessage;
+import net.ripe.db.whois.api.mail.EmailMessageInfo;
+import net.ripe.db.whois.api.mail.exception.MailParsingException;
+import org.elasticsearch.common.Strings;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+import javax.annotation.Nullable;
+import java.util.Collections;
+
+// Detect automated responses and mark them for deletion. Do not try to parse them as Whois updates.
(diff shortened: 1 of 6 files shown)


whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/AutoSubmittedMessageParser.java
```diff
@@ -0,0 +1,64 @@
...
+@Component
+public class AutoSubmittedMessageParser {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AutoSubmittedMessageParser.class);
+
+    private final boolean enabled;
+
+    @Autowired
+    public AutoSubmittedMessageParser(@Value("${mail.smtp.from:}") final String smtpFrom) {
+        this.enabled = ! Strings.isNullOrEmpty(smtpFrom);
```
MiguelAHM, line 28: white space between ! and Strings?

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/BouncedMessageParser.java
```diff
@@ -45,35 +48,100 @@ public BouncedMessageParser(@Value("${mail.smtp.from:}") final String smtpFrom)
...
+                    }
+                }
+            } catch (MessagingException | IOException | IllegalStateException ex){
+                throw new MailParsingException("Error parsing multipart report");
+            }
+            // multipart report can *only* be a failure
+            throw new MailParsingException("MultiPart message without failure report");
+        }
+
+        if (isMultipartMixed(message)) {
```
MiguelAHM, line 74: In this case. Shouldn't we return a MailParsingException??
I mean, should multipartMixed be processed as an error?? as we do with multipart report??

The mail updates are always in text plain, right? So everything that is not a text plain should not go to mail updates

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/BouncedMessageParser.java
```diff
@@ -45,35 +48,100 @@ public BouncedMessageParser(@Value("${mail.smtp.from:}") final String smtpFrom)
...
-                final DeliveryStatus deliveryStatus = deliveryStatus(message);
-                if (isFailed(deliveryStatus)) {
-                    final MimeMessage returnedMessage = multipartReport.getReturnedMessage();
-                    final String messageId = getMessageId(returnedMessage.getMessageID());
-                    final List<String> recipient = extractRecipients(deliveryStatus);
-                    return new EmailMessageInfo(recipient, messageId);
+        if (isMultipartReport(message)) {
+            try {
+                final MultipartReport multipartReport = multipartReport(message.getContent());
+                if (isReportDeliveryStatus(multipartReport)) {
```
MiguelAHM, line 58: Line 58 - 66 and 77 -85 are the same
can we extract this into a method called something like "extractMailMessageInfo"?

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/BouncedMessageParser.java
```diff
@@ -45,35 +48,100 @@ public BouncedMessageParser(@Value("${mail.smtp.from:}") final String smtpFrom)
...
+                LOGGER.error(String.format("%s: %s", ex.getClass().getName(), ex.getMessage()), ex);
+                throw new MailParsingException("Error parsing multipart report");
             }
-        } catch (MessagingException | IOException | IllegalStateException ex){
-            throw new MailParsingException("Error parsing multipart report");
         }
-        throw new MailParsingException("MultiPart message without failure report");
+
+        // fall through: message is not bounced message
+        return null;
```
MiguelAHM, line 93: Maybe early check that it is a plain text and return null if it a plain text (at the beginning of the method?)
and return a MailParsingException here...this is more strict but we would just allow mail updates when text plain

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/BouncedMessageParser.java
MiguelAHM: The mail updates are always in text plain, right? -> NO

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/AutoSubmittedMessageParser.java
eshryane: Fixed

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/BouncedMessageParser.java
eshryane: I don't want to disturb the existing code, and they are a bit different still - we throw an exception in the multipart/report code that we don't throw in the multipart/mixed code.
I added a TODO for now.

eshryane:
Integration tests passed

PR #1493, opened 2024-07-05
Parse nanoseconds in historical user response event date time

eshryane:


whois-client/src/main/java/net/ripe/db/whois/common/sso/domain/IsoDateTimeAdapter.java
```diff
@@ -8,13 +8,15 @@
 
 public class IsoDateTimeAdapter extends XmlAdapter<String, LocalDateTime> {
 
-    private static final DateTimeFormatter DATE_TIME_FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
+    private static final DateTimeFormatter DATE_TIME_FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.n'Z'");
```
maggarwal13, line 11: Sorry I am not familiar with this format. Does rennovate team asked for this change?

eshryane: Yes it's only used to parse the historical user details response. I fixed it earlier but the failing value happened to be 3 digits (so interpreted as milliseconds), but actually it's a nanosecond value.

eshryane: I renamed the class to EventDateTimeAdapter as it's specifically written for the `eventDateTime` field.

maggarwal13 (approved):

PR #1494, opened 2024-07-10
Set connection tester class name

eshryane:
Fix the warning from C3P0 on startup:

```
[C3P0PooledConnectionPool] Although no ConnectionTester is set, preferredTestQuery 
(or automaticTestTable) is also set, which can only be supported by a ConnectionTester. 
Reverting to use of ConnectionTester com.mchange.v2.c3p0.impl.DefaultConnectionTester.
```


eshryane:
Confirmed fixed by deploying to test environment, warning no longer appears

PR #1495, opened 2024-07-16
Create DoS Filter for Update Request

MiguelAHM:


MiguelAHM:
test ok
https://gitlab.ripe.net/swe-database-team/whois/-/jobs/1786714

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/DoSUpdateFilter.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.io.IOException;
+
+public class DoSUpdateFilter extends WhoisDoSFilter {
+
+    @Override
+    public void doFilter(final HttpServletRequest request, final HttpServletResponse response, final FilterChain chain) throws IOException, ServletException {
+        if (request != null) {
+            final String method = request.getMethod();
+            if (!"GET".equalsIgnoreCase(method)) {
```
eshryane, line 16: Invert the check, i.e. return early if the method IS a GET, it's easier to read and consistent with the rest of the code.

eshryane: Use `jakarta.ws.rs.HttpMethod.GET`

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/JettyBootstrap.java
```diff
@@ -137,8 +142,10 @@ public JettyBootstrap(final RemoteAddressFilter remoteAddressFilter,
         this.trustedIpRanges = trustedIpRanges;
         this.rewriteEngineEnabled = rewriteEngineEnabled;
         LOGGER.info("Rewrite engine is {}abled", rewriteEngineEnabled ? "en" : "dis");
-        this.dosFilterMBeanName = ObjectName.getInstance("net.ripe.db.whois:name=DosFilter");
+        this.dosLookUpFilterMBeanName = ObjectName.getInstance("net.ripe.db.whois:name=DosFilter");
```
eshryane, line 145: Why are there two separate management beans? Can't you create *one* to maintain the whitelist, which is shared between Lookup and Update?

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/JettyBootstrap.java
```diff
@@ -258,33 +265,51 @@ private Connector createInsecureConnector(final Server server, final HttpConfigu
...
+        holder.setInitParameter("maxRequestMs", maxRequestsPerMs);
+        holder.setInitParameter("maxRequestsPerSec", maxRequestsPerSecond);
+        setCommonDoSFilterConfiguration(holder);
+
+        return holder;
+    }
+
+    private void setCommonDoSFilterConfiguration(final FilterHolder holder) {
         holder.setInitParameter("enabled", Boolean.toString(dosFilterEnabled));
-        holder.setInitParameter("maxRequestsPerSec", "50");
```
eshryane, line 275: These setters are way easier to read than using constructor arguments

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/JettyBootstrap.java
```diff
@@ -258,33 +265,51 @@ private Connector createInsecureConnector(final Server server, final HttpConfigu
...
-        holder.setName("DoSFilter");
-
+    private void createDosFilter(final WebAppContext context) throws JmxException, JMException {
         if (!dosFilterEnabled) {
             LOGGER.info("DoSFilter is *not* enabled");
+            return;
         }
+
+        final String maxRequestPerMsLookUp = "" + 10 * 60 * 1_000; // high default, 10 minutes
+        final FilterHolder lookUpDoSFilter = createDosFilter("DoSLookUpFilter", "50", maxRequestPerMsLookUp, new DoSLookUpFilter());
```
eshryane, line 279: What does "50" mean? It's easier to read setters (like below) than passing many arguments in the constructor

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/JettyBootstrap.java
```diff
@@ -258,33 +265,51 @@ private Connector createInsecureConnector(final Server server, final HttpConfigu
...
+
+        final FilterHolder updateDoSFilter = createDosFilter("DoSUpdateFilter", dosUpdatesMax, "" + Integer.parseInt(dosUpdatesMax) * 1_000, new DoSUpdateFilter());
```
eshryane, line 281: Magic numbers here, separate them into a variable so it's easier to read.

eshryane: Move arithmetic into `createDosFilter` ?

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/WhoisDoSFilter.java
```diff
@@ -108,20 +108,16 @@ public void clearWhitelist() {
...
     @ManagedOperation("adds an IP address that will not be rate limited")
     public boolean addWhitelistAddress(@Name("address") final String address) {
-        return addWhitelistAddress(address, true);
+        logWhiteList();
+        return addWhitelistAddressWithoutLog(address);
     }
 
-    private boolean addWhitelistAddress(final String address, final boolean log) {
-        boolean result;
+    private boolean addWhitelistAddressWithoutLog(final String address) {
```
eshryane, line 115: Do you need to name the method `WithoutLog` ? Is there another method that logs adding per address?

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceDoSTestIntegration.java
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    @BeforeEach
+    public void setup() {
+        databaseHelper.addObject("person: Test Person\nnic-hdl: TP1-TEST");
+        databaseHelper.addObject(OWNER_MNT);
+        testDateTimeProvider.setTime(LocalDateTime.parse("2001-02-04T17:00:00"));
+    }
+
+    // LookUps
```
eshryane, line 61: `lookup` is one word, don't camelcase `Up`

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceDoSTestIntegration.java
```diff
@@ -0,0 +1,289 @@
...
+    public void multiple_async_lookUps_per_second_then_429_too_many_requests() throws InterruptedException {
+        final Response response = RestTest.target(getPort(), "whois/test/mntner/OWNER-MNT?clientIp=10.20.30.40")
+                .request()
+                .get();
+        assertThat(HttpStatus.OK_200, is(response.getStatus()));
+
+        final AsyncInvoker lookupRequest = RestTest.target(getPort(), "whois/test/mntner/OWNER-MNT?clientIp=10.20.30.40").request().async();
+
+        // Simulate a DoS attack by sending many POST requests in a short time asynchronously
+        final List<CompletableFuture<Response>> futures = IntStream.range(0, 50)
```
eshryane, line 94: Can you simplify using a `parallel()` stream instead ?

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceDoSTestIntegration.java
```diff
@@ -0,0 +1,289 @@
...
+                .post(Entity.entity(map(person), MediaType.APPLICATION_XML));
+        assertThat(HttpStatus.OK_200, is(response.getStatus()));
+
+        final Invocation.Builder updateRequest = RestTest.target(getPort(), "whois/test/person/PP1-TEST?clientIp=10.20.30.40&password=test").request();
+        final List<Integer> statuses = Lists.newArrayList();
+        for (int i = 0; i < 10; i++) {
+            statuses.add(updateRequest.put(Entity.entity(map(person), MediaType.APPLICATION_XML)).getStatus());
+        }
+
+        assertThat(HttpStatus.OK_200, is(statuses.get(8)));
```
eshryane, line 165: There are a lot of magic numbers in the tests. Can you make it more generic and only check that *some* requests have failed due to the rate limit?

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceDoSTestIntegration.java
```diff
@@ -0,0 +1,289 @@
...
+                        return future.get().getStatus();
+                    } catch (Exception e) {
+                        return null;
+                    }
+                })
+                .toList();
+        final List<Integer> manyRequestStatuses = statuses.stream().filter(status -> HttpStatus.TOO_MANY_REQUESTS_429 == status).toList();
+        assertThat(manyRequestStatuses.size(), is(1));
+
+        TimeUnit.SECONDS.sleep(1); // Free the IP after one second
```
eshryane, line 115: Can you avoid hard-coding sleeps? Tests may randomly fail

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/JettyBootstrap.java
```diff
@@ -122,6 +126,7 @@ public JettyBootstrap(final RemoteAddressFilter remoteAddressFilter,
                           @Value("${http.idle.timeout.sec:60}") final int idleTimeout,
                           @Value("${http.sni.host.check:true}") final boolean sniHostCheck,
                           @Value("${dos.filter.enabled:false}") final boolean dosFilterEnabled,
+                          @Value("${dos.filter.max.updates:10}") final String dosUpdatesMax,
```
maggarwal13, line 129: is this per second? rename method if possible to specify

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/JettyBootstrap.java
MiguelAHM: Should be clear in the lookup tests
just 50 gets are allowed per second. It was like this before.

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/WhoisDoSFilter.java
MiguelAHM: yes:

    @Override
    @ManagedOperation("adds an IP address that will not be rate limited")
    public boolean addWhitelistAddress(@Name("address") final String address) {
        logWhiteList();
        return addWhitelistAddressWithoutLog(address);
    }

I refactor this because of duplicated log everywhere.

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/DoSLookUpFilter.java
```diff
@@ -0,0 +1,27 @@
...
+
+
+import jakarta.servlet.FilterChain;
+import jakarta.servlet.ServletException;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
+
+import java.io.IOException;
+
+public class DoSLookUpFilter extends WhoisDoSFilter {
```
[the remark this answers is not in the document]
MiguelAHM: I am @Override doFilter
I can override for updates and lookups at same time. So no, I can not merge these two in one

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/JettyBootstrap.java
MiguelAHM: yes, it is per second.
Ok I will rename the variable

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceDoSTestIntegration.java
MiguelAHM: These test are doing more than 10 updates
testing that the update 9 is 200 but the update 10 is 429. I can rename that 8 and put "penultimateUpdate"

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceDoSTestIntegration.java
MiguelAHM: Updates using streams and parallel()

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/DoSUpdateFilter.java
MiguelAHM: Done

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/JettyBootstrap.java
MiguelAHM: I did refactor this. We don't need to create any bean here

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/JettyBootstrap.java
MiguelAHM: I refactor this, could you double check again?

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/JettyBootstrap.java
MiguelAHM: Is it clear now?

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/JettyBootstrap.java
MiguelAHM: I move all of these to different FilterHolders.
I think that it is much clear if we keep the configuration per each in different classes

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/DoSLookUpFilter.java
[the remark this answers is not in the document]
MiguelAHM: I did refactor this
I am not extending WhoisDosFilter, I am using anonymous inner classes to override that doFilter

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/JettyBootstrap.java
MiguelAHM: Done and moved to UpdateDoSFilterHolder

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/JettyBootstrap.java
```diff
@@ -258,33 +259,13 @@ private Connector createInsecureConnector(final Server server, final HttpConfigu
...
-        holder.setInitParameter("maxRequestsPerSec", "50");
-        holder.setInitParameter("maxRequestMs", "" + 10 * 60 * 1_000); // high default, 10 minutes
-        holder.setInitParameter("delayMs", "-1"); // reject requests over threshold
-        holder.setInitParameter("remotePort", "false");
-        holder.setInitParameter("trackSessions", "false");
-        holder.setInitParameter("insertHeaders", "false");
-        holder.setInitParameter("ipWhitelist", trustedIpRanges);
-
-        if (!ManagementFactory.getPlatformMBeanServer().isRegistered(dosFilterMBeanName)) {
-            ManagementFactory.getPlatformMBeanServer().registerMBean(new ObjectMBean(dosFilter), dosFilterMBeanName);
```
maggarwal13, line 284: We do not want to register them as JMX bean ?

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/AbstractDoSFilterHolder.java
```diff
@@ -0,0 +1,41 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractDoSFilterHolder.class);
+
+    AbstractDoSFilterHolder(final boolean dosFilterEnabled, final String trustedIpRanges){
+        if (!dosFilterEnabled) {
+            LOGGER.info("DoSFilter is *not* enabled");
+            this.setFilter(generateBasicFilter());
+            return;
+        }
+
+        this.setInitParameter("enabled", "true");
```
maggarwal13, line 25: This should use dosFilterEnabled , instead of setting manually to true? 
If it not enabled dos filer wont be used

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/AbstractDoSFilterHolder.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.io.IOException;
+
+public class AbstractDoSFilterHolder extends FilterHolder {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractDoSFilterHolder.class);
+
+    AbstractDoSFilterHolder(final boolean dosFilterEnabled, final String trustedIpRanges){
+        if (!dosFilterEnabled) {
+            LOGGER.info("DoSFilter is *not* enabled");
```
maggarwal13, line 20: Found a better way. just logging enough , setInitParameter will take care of rest

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/LookupDoSFilterHolder.java
```diff
@@ -0,0 +1,42 @@
...
+
+    private static final String MAX_REQUEST_PER_SECOND = "50";
+    private static final String MAX_REQUEST_PER_MS = "" + 10 * 60 * 1_000; // high default, 10 minutes
+
+    public LookupDoSFilterHolder(@Value("${dos.filter.enabled:false}") final boolean dosFilterEnabled,
+                                 @Value("${ipranges.trusted}") final String trustedIpRanges) {
+        super(dosFilterEnabled, trustedIpRanges);
+        
+        this.setFilter(generateWhoisDoSLookupFilter());
+        this.setName("LookupDoSFilter");
```
maggarwal13, line 25: shift them to abstract class and create a abstract method getFilterName and each implementation have this method

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/LookupDoSFilterHolder.java
```diff
@@ -0,0 +1,42 @@
...
+        this.setInitParameter("maxRequestMs", MAX_REQUEST_PER_MS);
+        this.setInitParameter("maxRequestsPerSec", MAX_REQUEST_PER_SECOND);
+    }
+
+    private WhoisDoSFilter generateWhoisDoSLookupFilter(){
+        return new WhoisDoSFilter(){
+            @Override
+            public void doFilter(final HttpServletRequest request, final HttpServletResponse response, final FilterChain chain) throws IOException, ServletException {
+                if (request == null || !GET.equalsIgnoreCase(request.getMethod())){
```
maggarwal13, line 34: create an abstract method allowedMethod which return boolean

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/WhoisDoSFilter.java
```diff
@@ -81,9 +81,9 @@ public String getWhitelist() {
     public void setWhitelist(final String commaSeparatedList) {
         clearWhitelist();
         for (String address : StringUtil.csvSplit(commaSeparatedList)) {
-            addWhitelistAddress(address, false);
+            addWhitelistAddressWithoutLog(address);
```
maggarwal13, line 84: Since we are registering two DOS filter , is the whitelist separate for query and update? If I remove an IP from whitelist will it be removed in both for query and update ?

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceDoSTestIntegration.java
```diff
@@ -0,0 +1,249 @@
...
+        // Simulate a DoS attack by sending many GET requests in a short time
+        final List<Integer> responsesCodes = IntStream.range(0, MAXIMUM_LOOKUP_REQUESTS_ALLOWED_PER_SECOND)
+                .mapToObj(lookupCount -> lookupRequest.get())
+                .map(Response::getStatus)
+                .toList();
+
+        assertThat(HttpStatus.OK_200, is(responsesCodes.get(PENULTIMATE_OK_LOOKUP_POSITION)));
+        assertThat(HttpStatus.TOO_MANY_REQUESTS_429, is(responsesCodes.getLast()));
+
+        TimeUnit.SECONDS.sleep(SECONDS_NEEDED_TO_FREE_IP); // Free the IP after one second
```
maggarwal13, line 87: instead of sleep , can we set the dateTime ? This will make test faster and consistent

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceDoSTestIntegration.java
```diff
@@ -0,0 +1,249 @@
...
+    private static final Integer PENULTIMATE_OK_LOOKUP_POSITION = 48;
+
+    private static final Integer MAXIMUM_UPDATE_REQUESTS_ALLOWED_PER_SECOND = 10;
+
+    private static final Integer MAXIMUM_LOOKUP_REQUESTS_ALLOWED_PER_SECOND = 50;
+
+    private static final Integer SECONDS_NEEDED_TO_FREE_IP = 1;
+
+    @BeforeAll
+    public static void beforeClass() {
```
maggarwal13, line 57: clear the list or set the dos filter false in AfterEach or AfterAll?

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceDoSTestIntegration.java
```diff
@@ -0,0 +1,249 @@
...
+            "upd-to:      noreply@ripe.net\n" +
+            "auth:        MD5-PW $1$d9fKeTr2$Si7YudNf4rUGmR71n/cqk/ #test\n" +
+            "auth:        SSO person@net.net\n" +
+            "mnt-by:      OWNER-MNT\n" +
+            "source:      TEST");
+
+    private static final Integer PENULTIMATE_OK_UPDATE_POSITION = 8;
+
+    private static final Integer PENULTIMATE_OK_LOOKUP_POSITION = 48;
+
```
maggarwal13, line 49: Why these are not one less than max allowed? Do we really need these numbers?

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceDoSTestIntegration.java
```diff
@@ -0,0 +1,249 @@
...
+                .get();
+        assertThat(HttpStatus.OK_200, is(creation.getStatus()));
+
+        final Invocation.Builder lookupRequest = RestTest.target(getPort(), "whois/test/mntner/OWNER-MNT?clientIp=10.20.30.40").request();
+
+        // Simulate a DoS attack by sending many GET requests in a short time asynchronously
+        final List<Integer> responsesCodes = IntStream.range(0, MAXIMUM_LOOKUP_REQUESTS_ALLOWED_PER_SECOND)
+                .parallel()
+                .mapToObj(lookupCount -> lookupRequest.get())
+                .map(Response::getStatus)
```
maggarwal13, line 106: use groupBy so you can get success and failed  count in one query

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/WhoisDoSFilter.java
MiguelAHM: White list will be created for update and for query (1 each)
If you remove something for the whiteList it will be removed from both.

However, I would like to use just one whiteList and not 1 each but I couldn't find the way

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/AbstractDoSFilterHolder.java
MiguelAHM: if it is not enabled we will not reach that line, we early return in line 22.
So the only possible value here is "true" I prefer put directly true rather than the property. It is more clear

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/AbstractDoSFilterHolder.java
MiguelAHM: that explain why we didn't have the return in master when is disabled

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/AbstractDoSFilterHolder.java
MiguelAHM: I will change it following your next advice

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/JettyBootstrap.java
MiguelAHM: I will check the Jira. I guess that the idea here is to register the beans. Why would we need a JMX for this?
Maybe if we want to modify the whiteList at any moment? Let me check.
Thanks for pointing out this

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/LookupDoSFilterHolder.java
```diff
@@ -0,0 +1,42 @@
...
+    private static final String MAX_REQUEST_PER_MS = "" + 10 * 60 * 1_000; // high default, 10 minutes
+
+    public LookupDoSFilterHolder(@Value("${dos.filter.enabled:false}") final boolean dosFilterEnabled,
+                                 @Value("${ipranges.trusted}") final String trustedIpRanges) {
+        super(dosFilterEnabled, trustedIpRanges);
+        
+        this.setFilter(generateWhoisDoSLookupFilter());
+        this.setName("LookupDoSFilter");
+        this.setInitParameter("maxRequestMs", MAX_REQUEST_PER_MS);
+        this.setInitParameter("maxRequestsPerSec", MAX_REQUEST_PER_SECOND);
```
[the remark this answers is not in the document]
MiguelAHM: I think that this is something to do when we can have more extensions of the abstract class. So we obligate the upcoming extensions to override these methods.
However, in this case that we will just have two extensions I think that it is more clear if we leave it as it is. 
I will do the change anyways. We can discuss later on

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceDoSTestIntegration.java
MiguelAHM: arrays position starts at 0
*_POSITION
I can do [size-2] maybe it is easier to understand?

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceDoSTestIntegration.java
MiguelAHM: yes, good idea

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceDoSTestIntegration.java
maggarwal13: Thanks, this explains. But can you use the variables from LookupDosFilter and updateDosFilter instead of defining separately in Test. Also do you think it will make sense to introduce the property and use default 50 and 10 ?

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/WhoisDoSFilter.java
maggarwal13: We should be using one whitelist only.

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/LookupDoSFilterHolder.java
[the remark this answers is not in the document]
maggarwal13: If we are using abstract class its best if we create abstract methods and move functionality in abstract class if we can , even if there is one implementation

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceDoSTestIntegration.java
MiguelAHM: I think that the purpose of testDateTImeProvider is different.
It can not be used here for making a delay. This seems to be for managing RPSL object instead timestamp.

I tried with 
`testDateTimeProvider.setTime(LOCAL_DATE_TIME.plusSeconds(1));`

Instead the sleep but it doesn't work. Then I checked JdbcSerialDoaIntegration and I see that the usage is different

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceDoSTestIntegration.java
MiguelAHM: Done

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceDoSTestIntegration.java
```diff
@@ -0,0 +1,249 @@
...
+            "mnt-by:      OWNER-MNT\n" +
+            "source:      TEST");
+
+    private static final Integer PENULTIMATE_OK_UPDATE_POSITION = 8;
+
+    private static final Integer PENULTIMATE_OK_LOOKUP_POSITION = 48;
+
+    private static final Integer MAXIMUM_UPDATE_REQUESTS_ALLOWED_PER_SECOND = 10;
+
+    private static final Integer MAXIMUM_LOOKUP_REQUESTS_ALLOWED_PER_SECOND = 50;
```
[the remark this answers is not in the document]
MiguelAHM: Yes, done

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceDoSTestIntegration.java
maggarwal13: Thanks for checking

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceDoSTestIntegration.java
MiguelAHM: now, with the group by this is no required.
So I removed it

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceDoSTestIntegration.java
MiguelAHM: Done!

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/LookupDoSFilterHolder.java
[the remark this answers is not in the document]
MiguelAHM: Done

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/LookupDoSFilterHolder.java
MiguelAHM: yes, done

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/AbstractDoSFilterHolder.java
MiguelAHM: done, I rely on that parameter

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/AbstractDoSFilterHolder.java
MiguelAHM: Done

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/JettyBootstrap.java
MiguelAHM: Talked via slack. We will create a new filter for black list IPs and that is the one that need to have the JMX call and not this one

MiguelAHM:
tests ok after refactors
https://gitlab.ripe.net/swe-database-team/whois/-/pipelines/639766

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/LookupDoSFilterHolder.java
```diff
@@ -0,0 +1,46 @@
...
+        super(dosFilterEnabled, trustedIpRanges);
+
+        this.dosQueriesMaxSecs = dosQueriesMaxSecs;
+        this.setInitParameter("maxRequestMs", getMaxRequestPerms());
+        this.setInitParameter("maxRequestsPerSec", getMaxRequestPerSec());
+    }
+
+    @Override
+    protected boolean isAllowedMethod(HttpServletRequest request) {
+        return request == null || !GET.equalsIgnoreCase(request.getMethod());
```
maggarwal13, line 29: It should be and not or ??

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/UpdateDoSFilterHolder.java
```diff
@@ -0,0 +1,43 @@
...
+        super(dosFilterEnabled, trustedIpRanges);
+
+        this.dosUpdatesMaxSecs = dosUpdatesMaxSecs;
+        this.setInitParameter("maxRequestMs", getMaxRequestPerms());
+        this.setInitParameter("maxRequestsPerSec", getMaxRequestPerSec());
+
+    }
+
+    @Override
+    protected boolean isAllowedMethod(final HttpServletRequest request) {
```
maggarwal13, line 27: Should be AND not OR ?

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/LookupDoSFilterHolder.java
maggarwal13: Also can you change the condition to request!=null & GET.equalsIgnoreCase(request.getMethod());?

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/UpdateDoSFilterHolder.java
maggarwal13: Also  request != null &  (PUT.equalsIgnoreCase(request.getMethod()) || POST.equalsIgnoreCase(request.getMethod())) ?

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/WhoisDoSFilter.java
```diff
@@ -132,16 +122,16 @@ private boolean addWhitelistAddress(final String address, final boolean log) {
      * @see #addWhitelistAddress(String)
      */
     @Override
-    @ManagedOperation("removes an IP address that will not be rate limited")
     public boolean removeWhitelistAddress(@Name("address") final String address) {
-        boolean result;
+        logWhiteList();
```
maggarwal13, line 126: logging should be after removal?

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/UpdateDoSFilterHolder.java
MiguelAHM: Should be isNotAllowed of change the logic.
I think that some refactor that I did make this unclear

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/LookupDoSFilterHolder.java
MiguelAHM: Refactor, now should be more clear.
We should or rename it to isNotAllowedMethod or change the logic as you suggest.
I changed the logic to avoid negation in the name

maggarwal13 (approved):

PR #1497, opened 2024-07-24
Separate query and update logic with subclasses rather than String comparison

eshryane:
Description

Separate query and update logic with subclasses rather than String comparison

Testing

Tested 6b33b17 that queries and updates have different limits, e.g. for updates:
```
$ ab -v 3 -p file -T "application/x-www-form-urlencoded" -c 4 -n 100 "https://syncupdates-prepdev.db.ripe.net/?clientIp=8.8.8.8" 2>&1 | grep 'Response'
```
With the word "TEST" in file.

And for queries:
```
ab -c 4 -n 100 -v 3 https://rest-prepdev.db.ripe.net/ripe/mntner/shryane-mnt | grep 'Response'
```

### DIFF
--- a/whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/WhoisDoSFilter.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/WhoisDoSFilter.java
@@ -30,9 +30,11 @@ public abstract class WhoisDoSFilter extends DoSFilter {
     private final List<Ipv4Resource> ipv4whitelist = new CopyOnWriteArrayList<>();
     private final List<Ipv6Resource> ipv6whitelist = new CopyOnWriteArrayList<>();
     private final Logger logger;
+    private final String limit;
 
-    public WhoisDoSFilter(final Logger logger) {
+    public WhoisDoSFilter(final Logger logger, final String limit) {
         this.logger = logger;
+        this.limit = limit;
     }
 
     @Override
@@ -153,5 +155,7 @@ public abstract class WhoisDoSFilter extends DoSFilter {
 
     protected abstract boolean canProceed(final HttpServletRequest request);
 
-    public abstract String getLimit();
+    public String getLimit(){
+        return limit;
+    };
 }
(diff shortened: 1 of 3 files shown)


whois-api/src/main/java/net/ripe/db/whois/api/httpserver/WhoisQueryDoSFilter.java
```diff
@@ -0,0 +1,24 @@
+package net.ripe.db.whois.api.httpserver;
+
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.ws.rs.HttpMethod;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class WhoisQueryDoSFilter extends WhoisDoSFilter {
```
MiguelAHM, line 8: try making component
and pass the value in the constructor

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/WhoisDoSFilter.java
```diff
@@ -30,9 +30,11 @@ public abstract class WhoisDoSFilter extends DoSFilter {
     private final List<Ipv4Resource> ipv4whitelist = new CopyOnWriteArrayList<>();
     private final List<Ipv6Resource> ipv6whitelist = new CopyOnWriteArrayList<>();
     private final Logger logger;
+    private final String limit;
 
-    public WhoisDoSFilter(final Logger logger) {
+    public WhoisDoSFilter(final Logger logger, final String limit) {
```
eshryane, line 35: Make `getLimit()` abstract as `WhoisDosFilter` doesn't use it

eshryane: Then you don't need the private field or constructor argument

eshryane: But, implementing it *once* is also OK

MiguelAHM (approved):

eshryane:
Integration tests passed

PR #1498, opened 2024-07-25
Permanent Blocked IPs JMX

MiguelAHM:
- Create a JMX for permanent blocked IPs for both UPDATES and LOOKUPS.
- Read initial list from properties.
- Use of HZ for spreading the JMX call through all the servers


whois-api/src/main/java/net/ripe/db/whois/api/httpserver/hazelcast/WhoisHazelcastManagerFilterMBean.java
```diff
@@ -0,0 +1,18 @@
+package net.ripe.db.whois.api.httpserver.hazelcast;
+
+import net.ripe.db.whois.api.httpserver.JettyBootstrap;
+import org.eclipse.jetty.util.annotation.ManagedOperation;
+import org.springframework.jmx.export.annotation.ManagedResource;
+
+@ManagedResource(objectName = JettyBootstrap.BLOCK_LIST_JMX_NAME)
+public interface WhoisHazelcastManagerFilterMBean {
```
MiguelAHM, line 8: required interface ending in MBean for using Hazelcast with JMX.

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/hazelcast/WhoisHazelcastManagerFilter.java
```diff
@@ -0,0 +1,132 @@
...
+import org.eclipse.jetty.util.annotation.ManagedOperation;
+import org.slf4j.Logger;
+import org.springframework.jmx.export.annotation.ManagedResource;
+
+import java.io.IOException;
+
+import static org.slf4j.LoggerFactory.getLogger;
+
+@ManagedResource(objectName = JettyBootstrap.BLOCK_LIST_JMX_NAME)
+public class WhoisHazelcastManagerFilter implements WhoisHazelcastManagerFilterMBean, Filter {
```
MiguelAHM, line 29: Same name that the Interface without MBean to avoid the issue
Class net.ripe.db.whois.api.httpserver.x is not a JMX compliant MXBean

whois-rpsl/src/main/java/net/ripe/db/whois/common/ip/Ipv4Resource.java
```diff
@@ -22,7 +23,7 @@
  * are stored as signed 32-bit <code>int</code>s. Externally they are
  * represented as <code>long</code>s to avoid issues with the sign-bit.
  */
-public final class Ipv4Resource extends IpInterval<Ipv4Resource> implements Comparable<Ipv4Resource> {
+public final class Ipv4Resource extends IpInterval<Ipv4Resource> implements Comparable<Ipv4Resource>, Serializable {
```
MiguelAHM, line 26: required to implement Serializable for using hazelcast

whois-rpsl/src/main/java/net/ripe/db/whois/common/ip/Ipv6Resource.java
```diff
@@ -8,13 +8,14 @@
...
 
+import java.io.Serializable;
 import java.math.BigInteger;
 import java.net.Inet6Address;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.util.regex.Pattern;
 
-public class Ipv6Resource extends IpInterval<Ipv6Resource> implements Comparable<Ipv6Resource> {
+public class Ipv6Resource extends IpInterval<Ipv6Resource> implements Comparable<Ipv6Resource>, Serializable {
```
MiguelAHM, line 18: required to implement Serializable for using hazelcast

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/hazelcast/WhoisHazelcastManagerFilter.java
MiguelAHM: no applicable after last refactor

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/hazelcast/WhoisHazelcastManagerFilterMBean.java
MiguelAHM: no applicable after last refactor

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/IpUtil.java
```diff
@@ -0,0 +1,33 @@
...
+
+import net.ripe.db.whois.common.ip.IpInterval;
+import net.ripe.db.whois.common.ip.Ipv4Resource;
+import net.ripe.db.whois.common.ip.Ipv6Resource;
+
+import java.util.Set;
+
+public class IpUtil {
+
+    public static boolean isExistingIp(final String candidate, final Set<Ipv4Resource> ipv4Resources, final Set<Ipv6Resource> ipv6Resources){
```
MiguelAHM, line 11: Better name 
isIpInIntervalRange or something like that

whois-api/src/test/java/net/ripe/db/whois/api/HazelcastInstanceTestManager.java
```diff
@@ -0,0 +1,56 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Profile;
+
+import java.util.List;
+
+@Configuration
+@Profile({WhoisProfile.TEST})
+public class HazelcastInstanceTestManager {
```
MiguelAHM, line 17: Do no start this.
There is a task for testing this.
Just test the filter in this class

whois-commons/src/main/java/net/ripe/db/whois/common/hazelcast/BlockListJmx.java
```diff
@@ -0,0 +1,94 @@
...
+
+    private static final Logger LOGGER = getLogger(BlockListJmx.class);
+
+    private static final Joiner COMMA_JOINER = Joiner.on(',');
+
+    private final ISet<Ipv4Resource> ipv4blockedSet;
+    private final ISet<Ipv6Resource> ipv6blockedSet;
+    private final HazelcastInstance hazelcastInstance;
+
+    public BlockListJmx(final HazelcastInstance hazelcastInstance, @Value("${ipranges.untrusted:}") final String untrustedIpRanges) {
```
MiguelAHM, line 32: Split in two
one JMX and HZ

whois-commons/src/main/java/net/ripe/db/whois/common/hazelcast/BlockListJmx.java
```diff
@@ -0,0 +1,94 @@
...
+
+@Component
+@ManagedResource(objectName = JmxBase.OBJECT_NAME_BASE + "BlockedList", description = "Whois black list")
+public class BlockListJmx extends JmxBase {
+
+    private static final Logger LOGGER = getLogger(BlockListJmx.class);
+
+    private static final Joiner COMMA_JOINER = Joiner.on(',');
+
+    private final ISet<Ipv4Resource> ipv4blockedSet;
```
MiguelAHM, line 28: Use just <IpInterval> instead of a different one? wait for Mahesh feedback

whois-rpsl/src/main/java/net/ripe/db/whois/common/ip/Ipv4Resource.java
MiguelAHM: move this to IpInterval

whois-rpsl/src/main/java/net/ripe/db/whois/common/ip/Ipv6Resource.java
MiguelAHM: move this to IpInterval

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/JettyBootstrap.java
```diff
@@ -130,8 +134,9 @@ public JettyBootstrap(final RemoteAddressFilter remoteAddressFilter,
                           @Value("${port.client.auth:-1}") final int clientAuthPort,
                           @Value("${http.x_forwarded_for:true}") final boolean xForwardedForHttp,
                           @Value("${https.x_forwarded_for:true}") final boolean xForwardedForHttps,
-                          @Value("${dos.filter.enabled:false}") final boolean dosFilterEnabled
-                        )  {
+                          @Value("${dos.filter.enabled:false}") final boolean dosFilterEnabled,
+                          final BlockListFilter blockListFilter
```
maggarwal13, line 138: Rename to IpBlockListFilter ?

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/BlockListFilter.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    @Override
+    public void destroy() {
+        // do nothing
+    }
+
+    private boolean isBlockedIp(final String candidate) {
+        final IpInterval<?> parsed = IpInterval.asIpInterval(InetAddresses.forString(candidate));
+        return hazelcastBlockedIps.getIpBlockedSet().stream()
```
maggarwal13, line 53: catch all exception here , log as warn or error

whois-commons/pom.xml
```diff
@@ -134,6 +134,11 @@
             <groupId>org.elasticsearch.client</groupId>
             <artifactId>elasticsearch-rest-high-level-client</artifactId>
         </dependency>
+        <dependency>
```
maggarwal13, line 137: Why we need this dependency ?

whois-commons/src/main/java/net/ripe/db/whois/common/hazelcast/WhoisHazelcastBlockedIps.java
```diff
@@ -0,0 +1,64 @@
...
+        LOGGER.info("Ipaddress {} removed from blocked list", address);
+    }
+
+    @Override
+    public String getBlockedList() {
+        StringBuilder result = new StringBuilder();
+        COMMA_JOINER.appendTo(result, ipBlockedSet);
+
+        LOGGER.info("The blocked list contains next IPs {}", result);
+        return String.format("The blocked list contains next IPs %s", result);
```
maggarwal13, line 56: Just return the comma separated string if its needed, not with the message

whois-commons/src/main/java/net/ripe/db/whois/common/hazelcast/WhoisHazelcastBlockedIps.java
```diff
@@ -0,0 +1,64 @@
...
+
+    private final ISet<IpInterval> ipBlockedSet;
+
+    public WhoisHazelcastBlockedIps(final HazelcastInstance hazelcastInstance,
+                                    @Value("${ipranges.blocked.list:}") final String blockedListIps) {
+
+        ipBlockedSet = hazelcastInstance.getSet("ipBlockedSet");
+
+        for (final String address : StringUtil.csvSplit(blockedListIps)) {
+            addBlockedListAddress(address);
```
maggarwal13, line 32: Why use csvSplit? How about Stream.of(blockedListIps.split(",")).map(IpInterval::parse).toList() ?

whois-commons/src/main/java/net/ripe/db/whois/common/hazelcast/WhoisHazelcastBlockedIps.java
```diff
@@ -0,0 +1,64 @@
...
+    private static final Joiner COMMA_JOINER = Joiner.on(',');
+
+    private final ISet<IpInterval> ipBlockedSet;
+
+    public WhoisHazelcastBlockedIps(final HazelcastInstance hazelcastInstance,
+                                    @Value("${ipranges.blocked.list:}") final String blockedListIps) {
+
+        ipBlockedSet = hazelcastInstance.getSet("ipBlockedSet");
+
+        for (final String address : StringUtil.csvSplit(blockedListIps)) {
```
maggarwal13, line 31: use addAll instead of add

whois-commons/pom.xml
MiguelAHM: import org.eclipse.jetty.util.StringUtil;
We are using this in "WhoisHazelcastBlockedIps"

whois-commons/src/main/java/net/ripe/db/whois/common/hazelcast/WhoisHazelcastBlockedIps.java
MiguelAHM: It is not needed I am adding this just because I like that the call returns a pretty well formed message.

whois-commons/src/main/java/net/ripe/db/whois/common/hazelcast/WhoisHazelcastBlockedIps.java
MiguelAHM: We are doing the same in WhoisDoSFilter
I can change this yes...but just want to be consistent

whois-commons/pom.xml
maggarwal13: No we should use StringUtils from lang package

whois-commons/src/main/java/net/ripe/db/whois/common/hazelcast/WhoisHazelcastBlockedIps.java
maggarwal13: Looking at message name feels like you are sending blocked Ips as a string

MiguelAHM: and because probably in the future you want a way to see if your change has been applied.
I think that this is ok

maggarwal13: In that case shift the message to JMX call and logging is also fine.

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/BlockListFilter.java
maggarwal13: I meant for this :  hazelcastBlockedIps.getIpBlockedSet() we do not want users to be blocked just incase hazelcast send error

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/JettyBootstrap.java
MiguelAHM: Done

whois-commons/pom.xml
MiguelAHM: I removed anyways this

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/BlockListFilter.java
MiguelAHM: Done

whois-commons/src/main/java/net/ripe/db/whois/common/hazelcast/WhoisHazelcastBlockedIps.java
MiguelAHM: Done

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/IpBlockListFilter.java
```diff
@@ -0,0 +1,66 @@
...
+
+    public IpBlockListFilter(final HazelcastBlockedIps hazelcastBlockedIps){
+        this.hazelcastBlockedIps = hazelcastBlockedIps;
+    }
+
+    @Override
+    public void init(FilterConfig filterConfig) throws ServletException {}
+
+    @Override
+    public void doFilter(final ServletRequest servletRequest, final ServletResponse servletResponse, final FilterChain filterChain) throws IOException, ServletException {
```
maggarwal13, line 38: Is this filter gets applied after remoteAddress customizer and DosFilter?

whois-commons/src/test/java/net/ripe/db/whois/common/hazelcast/TestHazelcastBlockedIps.java
```diff
@@ -0,0 +1,49 @@
...
+import org.mockito.internal.util.collections.Sets;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Profile;
+import org.springframework.stereotype.Component;
+
+import java.util.Set;
+
+@Component
+@Profile({WhoisProfile.TEST})
+public class TestHazelcastBlockedIps implements HazelcastBlockedIps {
```
maggarwal13, line 15: rename to TestBlockedIps?

whois-commons/src/main/java/net/ripe/db/whois/common/hazelcast/WhoisHazelcastBlockedIps.java
```diff
@@ -0,0 +1,61 @@
...
+
+    private final ISet<IpInterval> ipBlockedSet;
+
+    public WhoisHazelcastBlockedIps(final HazelcastInstance hazelcastInstance,
+                                    @Value("${ipranges.blocked.list:}") final String blockedListIps) {
+
+        ipBlockedSet = hazelcastInstance.getSet("ipBlockedSet");
+        ipBlockedSet.addAll(getBlockedIntervals(blockedListIps));
+
+        LOGGER.info("hazelcast instances {} members: {} " , hazelcastInstance.getName() , hazelcastInstance.getCluster().getMembers());
```
maggarwal13, line 32: remove this logging?

whois-commons/src/test/java/net/ripe/db/whois/common/hazelcast/TestHazelcastBlockedIps.java
MiguelAHM: fair enough because we are not using hazelcast there. OK

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/IpBlockListFilter.java
MiguelAHM: Before the DoSFilter is a requirement (read the implementation section of the task).
If the IP is permanent blocked shouldn't go to DoSFilter.

After the RemoteAddress yes, you can easily check this looking the ITs. We are using the clientIP there which is being moved to remoteAddress. You can easily check this as well putting a breakpoint in RemoteAddressFilter and running an IT.

whois-commons/src/main/java/net/ripe/db/whois/common/hazelcast/WhoisHazelcastBlockedIps.java
MiguelAHM: It is there because it is there in the `HazelcastPersonalObjectAccounting`class as well.
I will remove it then

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/IpBlockListFilter.java
```diff
@@ -0,0 +1,65 @@
+package net.ripe.db.whois.api.httpserver.dos;
+
```
maggarwal13, line 2: It is not a part of dos filter , can you create new package or move to a different one?

whois-commons/pom.xml
maggarwal13: whoisDosFilter and blocklistfilter are in same module so there should not be a need to add dependency but thanks for removing it

maggarwal13: Ahh got it, I don't think this needs to be in common package. As it will not be used everywhere , do you think it makes more sense to move to some other module? May be whois-api?

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/IpBlockListFilter.java
```diff
@@ -0,0 +1,65 @@
...
+        // do nothing
+    }
+
+    private boolean isBlockedIp(final String candidate) {
+        final IpInterval<?> parsed = IpInterval.asIpInterval(InetAddresses.forString(candidate));
+        try {
+            return hazelcastBlockedIps.getIpBlockedSet().stream()
+                    .anyMatch(ipRange -> ipRange.getClass().equals(parsed.getClass()) && ipRange.contains(parsed));
+        } catch (Exception ex){
+            LOGGER.error("Hazelcast blocked ip interval error", ex);
```
maggarwal13, line 61: LOGGER.error("Failed to check if remote address is in block list due to {}", ex) ?

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/IpBlockListFilter.java
```diff
@@ -0,0 +1,65 @@
...
+    public void init(FilterConfig filterConfig) throws ServletException {}
+
+    @Override
+    public void doFilter(final ServletRequest servletRequest, final ServletResponse servletResponse, final FilterChain filterChain) throws IOException, ServletException {
+        final HttpServletRequest httpRequest = (HttpServletRequest) servletRequest;
+        final HttpServletResponse httpResponse = (HttpServletResponse) servletResponse;
+
+        if (isBlockedIp(httpRequest.getRemoteAddr())){
+            httpResponse.setStatus(HttpStatus.TOO_MANY_REQUESTS_429);
+            httpResponse.getWriter().write("You have been permanently blocked. Please contact support");
```
maggarwal13, line 43: Can you check error message ? I think we should return similar to ACL or blocking. We do give reason in few words there.

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/IpBlockListFilter.java
```diff
@@ -0,0 +1,65 @@
...
+            return;
+        }
```
maggarwal13, line 45: we should use  response.sendError?

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/jmx/BlockListJmx.java
```diff
@@ -0,0 +1,49 @@
...
+            @ManagedOperationParameter(name = "address", description = "address to remove"),
+    })
+    public String removeBlockedListAddress(String address) {
+        hazelcastBlockedIps.removeBlockedListAddress(address);
+        return String.format("Ipaddress %s removed from blocked list", address);
+    }
+
+    @ManagedOperation(description = "Retrieve blocked list")
+    public String getBlockedList() {
+        return String.format("The blocked list contains next IPs %s", hazelcastBlockedIps.getBlockedList());
```
maggarwal13, line 47: List of blocked Ips are ?

whois-commons/src/main/java/net/ripe/db/whois/common/hazelcast/WhoisHazelcastBlockedIps.java
```diff
@@ -0,0 +1,59 @@
...
+        LOGGER.info("Ipaddress {} removed from blocked list", address);
+    }
+
+    @Override
+    public String getBlockedList() {
+        StringBuilder result = new StringBuilder();
+        COMMA_JOINER.appendTo(result, ipBlockedSet);
+
+        LOGGER.info("The blocked list contains next IPs {}", result);
+        return result.toString();
```
maggarwal13, line 52: May be use StringUtils.join(ipBlockedSet, ',')? There is no need for StringBuilder as list wont be so long but feel free to ignore

whois-commons/src/main/java/net/ripe/db/whois/common/hazelcast/WhoisHazelcastBlockedIps.java
MiguelAHM: This is the suggestion that I was talking before about no using StringUtils.csvSplit
the dependency is no longer necessary so I removed it.
This change is applied as well

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/IpBlockListFilter.java
MiguelAHM: oh...true
I am moving to httpserver package

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/IpBlockListFilter.java
MiguelAHM: you don't need the brackets because is an error, so the ex is already included at the end
but I will update the text

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/jmx/BlockListJmx.java
MiguelAHM: I don't understand what you are asking
This is just to get the list of IPs that are blocked.

MiguelAHM: ah ok. Let me redo that text

whois-commons/src/test/java/net/ripe/db/whois/common/hazelcast/TestHazelcastBlockedIps.java
maggarwal13: Sorry if it was not clear. I meant to say the interface name should not be hazelcast as you have two implementation one with hazelcast and one without so service layer should have hazelcast as prefix not interface.

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/IpBlockListFilter.java
MiguelAHM: the tasks says
"You have been permanently blocked... contact support" 

similar to ACL and temporary block
I can put more detail

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/IpBlockListFilter.java
MiguelAHM: doesn't work property with UPDATE and DELETE
the text gets disappear and replaced by "".

I created a test in WhoisRestBasicAuthTestIntegration with same issue. I think that we have a bug here in the code when using sendError()

whois-commons/src/main/java/net/ripe/db/whois/common/hazelcast/WhoisHazelcastBlockedIps.java
MiguelAHM: Done

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/jmx/BlockListJmx.java
MiguelAHM: done

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/IpBlockListFilter.java
MiguelAHM: I added something else in the message but I don't think that we should create an ad-hoc code for this.
At least not in this task

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/IpBlockListFilter.java
MiguelAHM: Done

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/IpBlockListFilter.java
MiguelAHM: Done

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestBasicAuthTestIntegration.java
```diff
@@ -221,6 +223,19 @@ public void create_failed_with_basic_auth_no_https() {
         assertThat(response.readEntity(String.class), containsString("HTTPS required for Basic authorization"));
     }
 
+    @Test
+    @Disabled("PUT and DELETE fail when using sendError()")
```
MiguelAHM, line 227: If we want to keep using sendError as this tests does
we need to tackle when the METHOD is PUT or DELETE

In these situations the error that we are sending is "" even if we do httpResponse.sendError(HttpStatus.UPGRADE_REQUIRED_426, "HTTPS required for Basic authorization") in the code

whois-commons/src/test/java/net/ripe/db/whois/common/hazelcast/TestHazelcastBlockedIps.java
MiguelAHM: I changed the name of the interface

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/IpBlockListFilter.java
maggarwal13: SendError is from jetty server, so should be doing right thing. Usually PUT and DELETE never expects a message body for erros. Can you check in case of put and delete error what we do in rest of code ? My guess is we do not send back  any message body

MiguelAHM: shall I remove the DELETE and PUT from my ITs?

maggarwal13: Why remove ? You should check status code is sent properly or not if it is put or delete and ip is blocked

MiguelAHM: I mean remove the body check from those ITs

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/IpBlockListFilter.java
maggarwal13: Can you check this message directly with Ed? I am fine with whatever he says , update the JIRA if its different than whats specified. Thanks

MiguelAHM: I talked to Ed
- Your host instead of Your account
- Include IP addresses 
- Suspected abuse is good because it echos the language in the AUP

MiguelAHM: Done

eshryane: Is the `httpResponse.getWriter()` the best (i.e. recommended) way to return a response body? Is the outputstream properly closed if this is used?

eshryane: Also why does the response `sendError()` not return a response body for DELETE or PUT responses? The method signature seems to indicate it can handle a status code and message, but the response body is empty. Also the method returns successfully without throwing an exception. What happens to the response?

MiguelAHM: httpResponse.getWriter() works fine and it is an alternative to sendError()

MiguelAHM: I would recommend upgrade jetty to 12 and try with sendError again

MiguelAHM: discussed vie slack. We will tackle this when upgrading to jetty 12
https://jetty.org/docs/jetty/12/programming-guide/migration/11-to-12.html

eshryane: Thanks for investigating Miguel

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestBasicAuthTestIntegration.java
```diff
@@ -215,7 +215,7 @@ public void create_failed_with_basic_auth_no_https() {
         final Response response =  RestTest.target(getPort(), "whois/test/person")
                     .request()
                     .header(HttpHeaders.AUTHORIZATION, getBasicAuthenticationHeader("test", "test"))
-                    .post(Entity.entity(map(PAULETH_PALTHEN), MediaType.APPLICATION_XML), Response.class);
```
maggarwal13, line 218: Revert the change

MiguelAHM: yes, done

maggarwal13 (approved):

PR #1499, opened 2024-07-25
Fix column name in NRTM 114 SQL script

eshryane:


maggarwal13 (approved):
Thanks for the fix Ed

PR #1500, opened 2024-07-26
error message changed in test when no client certificate is supplied

eshryane:


maggarwal13 (approved):

eshryane:
Seems this test was passing a few days ago, but now it's returning an error. It may be a flapping test, or there may be an interaction with the test environment

eshryane:
Can't reproduce it

PR #1505, opened 2024-08-06
Fix Header Value Too Many Content Type ITs

MiguelAHM:


eshryane:
Why are there only changes to a test class? Is there an issue with the (deployed) source code or only the tests?

MiguelAHM:
> Why are there only changes to a test class? Is there an issue with the (deployed) source code or only the tests?

just tests

MiguelAHM:
I run this several times and never get that error again. See the runs
https://gitlab.ripe.net/swe-database-team/whois/-/jobs/1836420

eshryane (approved):

MiguelAHM:
not able to fix the Integration tests

PR #1507, opened 2024-08-12
fix datasource configs

maggarwal13:
DefaultConnectionTester and other datasource properties not being picked from Properties.Setting them explicitly


whois-commons/src/main/java/net/ripe/db/whois/common/DataSourceConfigurations.java
```diff
@@ -1,4 +1,4 @@
-package net.ripe.db.whois.common.source;
+package net.ripe.db.whois.common;
```
MiguelAHM, line 1: why moving out of the `source` package?

whois-commons/src/main/java/net/ripe/db/whois/common/DataSourceConfigurations.java
```diff
@@ -59,24 +58,19 @@ public ComboPooledDataSource nrtmSlaveDataSource(@Value("${nrtm.slave.database.u
     private ComboPooledDataSource createDataSource(final String jdbcDriver, final String jdbcUrl, final String jdbcUser, final String jdbcPass) throws PropertyVetoException {
         final ComboPooledDataSource source = new ComboPooledDataSource();
 
-        source.setProperties(getSourceProperties(jdbcDriver));
```
MiguelAHM, line 62: shouldn't be the same
setting properties on this way on another?
using setProperties or using setMinPoolSize...etc

whois-api/src/test/java/net/ripe/db/whois/api/rest/TestTransactionDao.java
```diff
@@ -8,7 +8,7 @@
 import net.ripe.db.whois.common.dao.jdbc.JdbcRpslObjectDao;
 import net.ripe.db.whois.common.domain.BlockEvent;
 import net.ripe.db.whois.common.mail.EmailStatus;
-import net.ripe.db.whois.common.transaction.TransactionConfiguration;
+import net.ripe.db.whois.common.TransactionConfiguration;
```
MiguelAHM, line 11: why moving out `transaction` package?

whois-commons/src/main/java/net/ripe/db/whois/common/DataSourceConfigurations.java
MiguelAHM: Ok, I see that you are doing this change in internal as well...
I don't understand why it is no working using `setProperties` 
I will close the issue but I am curious now

MiguelAHM (approved):

whois-api/src/test/java/net/ripe/db/whois/api/rest/TestTransactionDao.java
maggarwal13: because rest of configs are in base ackage

whois-commons/src/main/java/net/ripe/db/whois/common/DataSourceConfigurations.java
maggarwal13: because rest of configs are in base ackage

PR #1510, opened 2024-08-15
Change LOGGER level from Warning to Debug

MiguelAHM:
In case of mailing list we don't want to mark the message as UNDELIVERABLE and we don't need to take any action
so I suggest changing the LOGGING level from WARN to DEBUG to avoid having alerts from Sentry


eshryane:
I agree, we had initially set logging to warning on any unexpected situation, in case we weren't handling it properly. I think now we can safely set this warning to debug.

eshryane (approved):

PR #1511, opened 2024-08-16
Add Warning When Password Authentication is used in Mailupdates

MiguelAHM:


whois-api/src/test/java/net/ripe/db/whois/api/mail/dequeue/MessageServiceTestIntegration.java
```diff
@@ -24,6 +28,14 @@ public class MessageServiceTestIntegration extends AbstractMailMessageIntegratio
     @Autowired
     private MessageService messageService;
 
+    final static String PASSWORD_WARN =
+            """
+            Warning: Password authentication will be removed from Mailupdates in a future
+                        Whois release as the mail message may have been sent insecurely.
+                        Please switch to PGP signing for authentication or use a different
```
maggarwal13, line 35: Why not use the UpdateMessages?

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageDequeue.java
```diff
@@ -272,9 +274,19 @@ private void handleMessageInContext(final String messageId, final MimeMessage me
...
 
         final UpdateRequest updateRequest = new UpdateRequest(mailMessage, mailMessage.getKeyword(), updates);
         final UpdateResponse response = messageHandler.handle(updateRequest, updateContext);
         mailGateway.sendEmail(mailMessage.getReplyToEmail(), response.getStatus() + ": " + mailMessage.getSubject(), response.getResponse(), null);
     }
+
+    private static void addWarnIfPasswordExists(final UpdateContext updateContext, final List<Update> updates) {
+        for (Update update : updates) {
+            if (!update.getCredentials().ofType(PasswordCredential.class).isEmpty()){
+                updateContext.addGlobalMessage(UpdateMessages.passwordInMailUpdate());
```
maggarwal13, line 287: will it add multiple global messages or single one incase more than one update happened

MiguelAHM: single one. It will return; once the global message is added

MiguelAHM: I will create a test for this

maggarwal13: I checked it using a set so should be good

maggarwal13 (approved):

whois-api/src/test/java/net/ripe/db/whois/api/mail/dequeue/MessageServiceTestIntegration.java
MiguelAHM: Just tests and if I use that message I get an unformatted string
all the string is in one line

maggarwal13 (approved):

whois-api/src/test/java/net/ripe/db/whois/api/mail/dequeue/MessageServiceTestIntegration.java
```diff
@@ -24,6 +29,13 @@ public class MessageServiceTestIntegration extends AbstractMailMessageIntegratio
     @Autowired
     private MessageService messageService;
 
+    private static final String FORMATTED_PASSWORD_WARN = """
+            ***Warning: Password authentication will be removed from Mailupdates in a future
```
maggarwal13, line 33: Why not use UpdateMessages.getText()

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageDequeue.java
```diff
@@ -272,9 +274,19 @@ private void handleMessageInContext(final String messageId, final MimeMessage me
 
     private void handleUpdates(final MailMessage mailMessage, final UpdateContext updateContext) {
         final List<Update> updates = updatesParser.parse(updateContext, mailMessage.getContentWithCredentials());
+        addWarnIfPasswordExists(updateContext, updates);
```
maggarwal13, line 277: I thought the story says that it should be controlled via feature flag ? If feature flag is set to false return warning and if set to true return error with default sets to false?

maggarwal13: Put this change behind a feature flag so it can be enabled or disabled independently of Whois releases. If the flag is OFF then return a warning (make this the default). If the flag is ON then return an error.

PR #1513, opened 2024-08-20
Remove Jetty filter for unused HTTP/2 Server Push feature

eshryane:
The Push filter was added when we implemented support for HTTP/2. However, Whois never made use of the feature. 

Now Jetty has deprecated the filter:
`@deprecated no replacement for this deprecated http feature (will be removed in Jetty 12)`
https://github.com/jetty/jetty.project/blob/jetty-11.0.x/jetty-servlets/src/main/java/org/eclipse/jetty/servlets/PushCacheFilter.java

So we also remove the filter.


eshryane:
Tested that HTTP2 is still working properly

MiguelAHM (approved):

PR #1515, opened 2024-08-22
Text Content Type processed as text/plain for Messages

MiguelAHM:


whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageParser.java
```diff
@@ -225,6 +226,12 @@ private void parseContents(@Nonnull final MessageParts messageParts, @Nonnull fi
         }
     }
 
+    private static void handleInvalidTypes(Part part) throws MessagingException {
+        if ("text".equals(part.getContentType())){
+            part.setHeader("Content-Type", "text/plain; charset=UTF-8");
```
eshryane, line 231: Be careful, you don't know what the character set is? Why default to UTF-8 ?

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageParser.java
```diff
@@ -177,6 +177,7 @@ private void parseContents(@Nonnull final MailMessageBuilder messageBuilder, @No
     }
 
     private void parseContents(@Nonnull final MessageParts messageParts, @Nonnull final Part part, @Nullable final Part parentPart) throws MessagingException, IOException {
+        handleInvalidTypes(part);
         final ContentType contentType = new ContentType(part.getContentType());
```
eshryane, line 181: Could you use a `new ContentType("text/plain")` if it's text, rather than re-writing the message part header?

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageParser.java
eshryane: At least for HTTP, if the type is "text" and no charset is defined, then the default is ASCII :
https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#

eshryane: The RFC 6657 : https://datatracker.ietf.org/doc/html/rfc6657
updates the original rules, but still :
```
The default "charset" parameter value for "text/plain" is unchanged
   from [[RFC2046](https://datatracker.ietf.org/doc/html/rfc2046)] and remains as "US-ASCII".
```
Please review to confirm my understanding is correct, and fits your use case!

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageParser.java
MiguelAHM: If I don't fix this in the message then line 173 will throw an error
we are doing new ContentType(message.getContentType()) there as well
I don't want to do this check everywhere...if the message contentType is not parseable by Jakarta library we should fix the message directly. No if/else everywhere when we are doing getContentType()

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageParser.java
MiguelAHM: yes, agree. by default is setting the charset to ISO-8859-1 so I don't need to specify nothing there

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageParser.java
eshryane: OK

eshryane (approved):

PR #1517, opened 2024-09-02
transfer update lock to internals schema

maggarwal13:
transfer_update_lock should be in Internals schema as transfers happens there and this will also avoids  transaction on multiple datasources

### DIFF
--- a/whois-commons/src/main/resources/internals_data.sql
+++ b/whois-commons/src/main/resources/internals_data.sql
@@ -23,3 +23,5 @@ INSERT INTO authoritative_resource (source, resource) VALUES ('test', '0.0.0.0/0
 INSERT INTO legacy_autnums (autnum) VALUES ('103');
 INSERT INTO legacy_autnums (autnum) VALUES ('12666');
 INSERT INTO legacy_autnums (autnum) VALUES ('12668');
+
+INSERT INTO transfer_update_lock VALUES (0);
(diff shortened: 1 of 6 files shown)


whois-commons/src/main/resources/internals_data.sql
```diff
@@ -22,4 +22,6 @@ INSERT INTO authoritative_resource (source, resource) VALUES ('test', '0.0.0.0/0
 -- legacy autnums
 INSERT INTO legacy_autnums (autnum) VALUES ('103');
 INSERT INTO legacy_autnums (autnum) VALUES ('12666');
-INSERT INTO legacy_autnums (autnum) VALUES ('12668');
\ No newline at end of file
+INSERT INTO legacy_autnums (autnum) VALUES ('12668');
+
+INSERT INTO transfer_update_lock VALUES (0);
```
MiguelAHM, line 27: Maybe add a comment on top of this? To follow the same format that rest of INSERTS?

whois-commons/src/main/resources/patch/whois-1.197.sql
```diff
@@ -0,0 +1,4 @@
+DROP TABLE IF EXISTS `transfer_update_lock`;
```
MiguelAHM, line 1: why this?

MiguelAHM: I see we have that table in WHOIS_UPDATE_RIPE.
OK

maggarwal13: because we are moving this table to internals so delete this table from whois schema

MiguelAHM (approved):

PR #1550, opened 2024-09-24
Normalise email address value before adding to index table

eshryane:


eshryane:
Integration tests passed 8ddb94b

whois-commons/src/main/java/net/ripe/db/whois/common/dao/jdbc/index/IndexWithEmailAddressValueAndType.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    @Override
+    public int addToIndex(JdbcTemplate jdbcTemplate, RpslObjectInfo objectInfo, RpslObject object, String value) {
+        return super.addToIndex(jdbcTemplate, objectInfo, object, normaliseEmailAddress(value));
+    }
+
+    private String normaliseEmailAddress(final String emailAddress) {
+        try {
+            return EMAIL_PARSER.parse(emailAddress).getAddress();
```
maggarwal13, line 29: this function is  duplicated with  indexWithEmailAddressValue?

whois-commons/src/main/java/net/ripe/db/whois/common/dao/jdbc/index/IndexWithEmailAddressValueAndType.java
```diff
@@ -0,0 +1,35 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(IndexWithEmailAddressValueAndType.class);
+
+    private static final AttributeParser.EmailParser EMAIL_PARSER = new AttributeParser.EmailParser();
+
+    public IndexWithEmailAddressValueAndType(final AttributeType attributeType, final String lookupTableName, final String lookupColumnName) {
+        super(attributeType, lookupTableName, lookupColumnName);
+    }
+
+    @Override
+    public int addToIndex(JdbcTemplate jdbcTemplate, RpslObjectInfo objectInfo, RpslObject object, String value) {
```
maggarwal13, line 23: shoudln't we make sure that this function is called for email type only?

whois-commons/src/main/java/net/ripe/db/whois/common/dao/jdbc/index/IndexWithEmailAddressValueAndType.java
eshryane: There's no easy way to share this method as the classes extend different classes (the hierarchy is different)

whois-commons/src/main/java/net/ripe/db/whois/common/dao/jdbc/index/IndexWithEmailAddressValueAndType.java
eshryane: That's done in `IndexStrategies.java` you define which strategy to use for different attribute types. This class is only added to attributes that contain an email address.

whois-commons/src/main/java/net/ripe/db/whois/common/dao/jdbc/index/IndexWithEmailAddressValueAndType.java
eshryane: We have to extend `IndexWithValue` and `IndexWithValueAndType` with separate classes

whois-commons/src/main/java/net/ripe/db/whois/common/dao/jdbc/index/IndexWithEmailAddressValueAndType.java
maggarwal13: Normalise email address function can be static to make sure both the functions are doing same thing and code is not duplicated. May be shift it to util class?

eshryane: Good idea. I'll make it static but I'll leave it in one of the Index classes and call it from the other one

eshryane: Updated code changes

whois-commons/src/main/java/net/ripe/db/whois/common/dao/jdbc/index/IndexWithEmailAddressValueAndType.java
eshryane: Updated code changes

PR #1558, opened 2024-10-15
Let Jackson deserialise eventDateTime in Historical User response

eshryane:


eshryane:
Tested OK with SSO Historical User API.

PR #1559, opened 2024-10-17
MailUpdates Error when Password Property

MiguelAHM:


whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageDequeue.java
```diff
@@ -281,10 +284,11 @@ private void handleUpdates(final MailMessage mailMessage, final UpdateContext up
         mailGateway.sendEmail(mailMessage.getReplyToEmail(), response.getStatus() + ": " + mailMessage.getSubject(), response.getResponse(), null);
     }
 
-    private static void addWarnIfPasswordExists(final UpdateContext updateContext, final List<Update> updates) {
+    private void addWarnIfPasswordExists(final UpdateContext updateContext, final List<Update> updates) {
```
maggarwal13, line 287: rename to validatePasswordCredentials?

whois-api/src/test/java/net/ripe/db/whois/api/mail/dequeue/MessageServicePasswdErrorTestIntegration.java
```diff
@@ -0,0 +1,159 @@
...
+
+@Tag("IntegrationTest")
+public class MessageServicePasswdErrorTestIntegration extends AbstractMailMessageIntegrationTest {
+
+    @Autowired
+    private MailSenderStub mailSenderStub;
+
+
+    private static final String FORMATTED_PASSWORD_ERROR = """
+            ***Error:   Password authentication is not allowed in Mailupdates, because your
```
maggarwal13, line 25: Why not use UpdateMessages.getText?

MiguelAHM: We discussed this already
UpdateMessages.getText doesn't give you the correct format
so it will never match.
The format of the output and what UpdateMessages.getText or getFormattedText is not the same

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageDequeue.java
```diff
@@ -67,6 +67,9 @@ public class MessageDequeue implements ApplicationService {
     @Value("${mail.dequeue.interval:1000}")
     private int intervalMs;
 
+    @Value("${mailupdates.passwd.error:false}")
+    private boolean passwdError;
```
maggarwal13, line 71: mailupdates.use.password.allowed?  By default should be true 
and variable name is passwordAllowed

MiguelAHM: no, by default is not allowed
` If the flag is OFF then return a warning (make this the default)`

maggarwal13 (approved):

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageDequeue.java
MiguelAHM: I rename it by `errorIfPassword`

PR #1562, opened 2024-10-18
Store All Message in Database When Undeliverable Email

MiguelAHM:


whois-api/src/main/java/net/ripe/db/whois/api/mail/EmailMessageInfo.java
```diff
@@ -2,5 +2,5 @@
 
 import java.util.List;
 
-public record EmailMessageInfo(List<String> emailAddresses, String messageId) {
+public record EmailMessageInfo(List<String> emailAddresses, String messageId, byte[] message) {
```
eshryane, line 5: Can you use a `MimeMessage` until you store it in the DAO as a `byte[]` ?

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/BouncedMessageParser.java
```diff
@@ -246,4 +251,10 @@ private static ContentType contentType(final String contentType) {
         }
     }
 
+    private static byte[] getMimeMessageBytes(final MimeMessage message) throws MessagingException, IOException {
```
eshryane, line 254: Move this method to the DAO where the MimeMessage is written to the database

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageService.java
```diff
@@ -90,7 +90,7 @@ public void verifyAndSetAsUnsubscribed(final EmailMessageInfo message) {
         }
 
         LOGGER.debug("Unsubscribe message-id {} email {}", message.messageId(), unsubscribeRequestEmail);
-        emailStatusDao.createEmailStatus(unsubscribeRequestEmail, EmailStatusType.UNSUBSCRIBE);
+        emailStatusDao.createEmailStatus(unsubscribeRequestEmail, EmailStatusType.UNSUBSCRIBE, null);
```
eshryane, line 93: May as well store the message blob for unsubscribe also?

MiguelAHM: Why? I don't think we will need it for anything

whois-api/src/main/java/net/ripe/db/whois/api/mail/EmailMessageInfo.java
MiguelAHM: I though that it was better to do this here because we are parsing the message.
and transforming it into bytes is part of a parsing procedure.
But I don't care to move it to the dao...however, the dao shouldn't do anything else rather that adding a record into a table...

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/BouncedMessageParser.java
MiguelAHM: I will do this, but I think that we shouldn't add extra logic into a dao layer

whois-commons/src/main/resources/patch/internals-1.116.sql
```diff
@@ -0,0 +1,13 @@
+DROP TABLE IF EXISTS `email_status`;
```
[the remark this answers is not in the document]
MiguelAHM: pfff...yes, thanks for pointing into this

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageService.java
eshryane: Good point, as discussed we don't need unsubscribed messages, so leave it NULL

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageService.java
```diff
@@ -58,24 +59,26 @@ public EmailMessageInfo getAutomatedFailureMessageInfo(final MimeMessage message
         return autoSubmittedMessageParser.parse(message);
     }
 
-    public void verifyAndSetAsUndeliverable(final EmailMessageInfo message) {
-        final List<String> outgoingEmail = outgoingMessageDao.getEmails(message.messageId());
+    public void verifyAndSetAsUndeliverable(final EmailMessageInfo messageInfo) throws MailParsingException {
```
eshryane, line 62: Where does a `MailParsingException` end up? Make sure it doesn't break the Mailupdate message flow (i.e. the message should still be handled cleanly)

MiguelAHM: at this point the message has been detected as bounced message and this error occurs because it cannot be transform into bytes[]
Do we still want to continue with the flow? in this situation should I attempt just storing the bounced email without storing the byte[]?

MiguelAHM: I pushed doing this, in case of this issue the message will not be stored, just the email and the status (as we are doing now)

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/UnsubscribeMessageParser.java
```diff
@@ -42,7 +42,7 @@ public EmailMessageInfo parse(final MimeMessage message) throws MessagingExcepti
             final String messageId = getMessageIdFromSubject(message);
             final String from = getFrom(message);
             if ((messageId != null) && (from != null)) {
-                return new EmailMessageInfo(List.of(from), messageId);
+                return new EmailMessageInfo(List.of(from), messageId, null);
```
maggarwal13, line 45: Instead of null can you save the message that you are unable to parse?

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageService.java
```diff
@@ -58,21 +59,24 @@ public EmailMessageInfo getAutomatedFailureMessageInfo(final MimeMessage message
...
+        for (String email : messageInfo.emailAddresses()) {
             try {
-                emailStatusDao.createEmailStatus(email, EmailStatusType.UNDELIVERABLE);
+                emailStatusDao.createEmailStatus(email, EmailStatusType.UNDELIVERABLE, messageInfo.message());
             } catch (DuplicateKeyException ex) {
-                LOGGER.debug("Email already exist in EmailStatus table {}", StringUtils.join(message.emailAddresses(), ", "), ex);
+                LOGGER.debug("Email already exist in EmailStatus table {}", StringUtils.join(messageInfo.emailAddresses(), ", "), ex);
+            } catch (MessagingException | IOException e) {
+                LOGGER.error("Unable to transform the bounced message of {} into byte[]", email);
+                emailStatusDao.createEmailStatus(email, EmailStatusType.UNDELIVERABLE);
```
maggarwal13, line 77: Instead of null can you save the message that you are unable to parse?

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageService.java
```diff
@@ -58,21 +59,24 @@ public EmailMessageInfo getAutomatedFailureMessageInfo(final MimeMessage message
...
             return;
         }
 
-        LOGGER.debug("Undeliverable message-id {} email {}", message.messageId(), StringUtils.join(message.emailAddresses(), ", "));
-        message.emailAddresses().forEach(email -> {
+        LOGGER.debug("Undeliverable message-id {} email {}", messageInfo.messageId(), StringUtils.join(messageInfo.emailAddresses(), ", "));
+        for (String email : messageInfo.emailAddresses()) {
             try {
-                emailStatusDao.createEmailStatus(email, EmailStatusType.UNDELIVERABLE);
+                emailStatusDao.createEmailStatus(email, EmailStatusType.UNDELIVERABLE, messageInfo.message());
```
maggarwal13, line 72: Shoudl the existing entry needs to be updated with latest messageInfo ? Why it is undeliverable reason can be different?

MiguelAHM: We can not have two same undeliverable emails
If an email is already undeliverable we don't seen another email to it?

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageService.java
MiguelAHM: the data in database is a longblob type...
can I store a raw string message there? I thought that I need to get the byte[] to store it.

whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageService.java
maggarwal13: MailParsingException is now handled properly and not propogated.

maggarwal13 (approved):

PR #1566, opened 2024-10-21
Fixed Auth Service test

eshryane:
Fixed regression introduced by #1558


eshryane:
Integration tests passed

PR #1574, opened 2024-10-25
NRTM Client Integration Test Set Up

MiguelAHM:


whois-nrtm4-client/src/test/java/net/ripe/db/nrtm4/client/NrtmServerDummy.java
```diff
@@ -0,0 +1,138 @@
...
+import org.springframework.test.util.ReflectionTestUtils;
+
+import java.io.IOException;
+import java.nio.charset.Charset;
+import java.util.List;
+
+@Profile({WhoisProfile.TEST})
+@Component
+public class NrtmServerDummy implements Stub {
+
```
maggarwal13, line 31: You should not be needing this  if whois server is started and nrtm.emabled is true

MiguelAHM: yes, we are not using an external service...it is the same project. You are right.
Let me refactor this

maggarwal13: Ok after the discussion on chat, let's keep the stub and mocking nrtmv4 server side. We can change this later on but its fine to use stubs for now

maggarwal13 (approved):

PR #1575, opened 2024-10-29
NRTM Client Persist Snapshot

MiguelAHM:


whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/client/NrtmRestClient.java
```diff
@@ -82,6 +95,31 @@ public UpdateNotificationFileResponse getNotificationFile(final String source){
...
+    @Nullable
+    public SnapshotFileResponse getSnapshotFile(final String url){
+        try {
+            final Response response =  client.target(url)
+                    .request(MediaType.APPLICATION_OCTET_STREAM)
+                    .header(HttpHeader.X_FORWARDED_PROTO.asString(), HttpScheme.HTTPS.asString())
+                    .get(Response.class);
+
+            final byte[] payload = response.readEntity(byte[].class);
+            final String[] records = getSnapshotRecords(payload);
```
maggarwal13, line 107: Did you compare the performance? I think loading the whole snapshot file  in one go can increase memory usage quite drastically. We should read it per record, since its a json text sequence read it by a record seprator and pass it to a seperate thread  to process i. You can use the parallelstream to do so

MiguelAHM: the snapshot is compressed in .gz
a json text sequence is inside the .gz

MiguelAHM: [SnapshotImporter] loading snapshot took 3045 ms -> 3 seconds

maggarwal13: Yes RIPE NONAUTH is very small file so 3sec for it to load is still lot of time. Can you compare for RIPE?  RIPE It doesn't work , means does it throw OOM?

MiguelAHM: yes, even using parallel is throwing OOM
Maybe is because I am testing in a DEV machine which doesn't have too much memory?

whois-commons/src/main/resources/nrtm_client_schema.sql
```diff
@@ -59,7 +59,7 @@ CREATE TABLE `last_mirror`
     `object_id` int(10) unsigned NOT NULL AUTO_INCREMENT,
     `sequence_id` int(10) unsigned NOT NULL DEFAULT '1',
     `timestamp` int(10) unsigned NOT NULL DEFAULT '0',
-    `object_type` tinyint(3) unsigned NOT NULL DEFAULT '0',
+    `object_type` varchar(254) NOT NULL DEFAULT '0',
```
eshryane, line 62: Why are you not using `tinyint(3)` like the last and history tables?

whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/importer/SnapshotImporter.java
```diff
@@ -0,0 +1,170 @@
...
+            }
+        }, 0, 10000);
+    }
+
+    private static String[] getSnapshotRecords(byte[] compressed) throws IOException {
+        return StringUtils.split(decompress(compressed), RECORD_SEPARATOR);
+    }
+
+    private static String decompress(final byte[] compressed) throws IOException {
+        final int BUFFER_SIZE = 4096;
```
eshryane, line 136: If you intend `BUFFER_SIZE` to be a static field then declare it so at the top of the class

whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/importer/SnapshotImporter.java
```diff
@@ -0,0 +1,170 @@
...
+            final MessageDigest digest = MessageDigest.getInstance("SHA-256");
+            final byte[] encodedSha256hex = digest.digest(bytes);
+            return byteArrayToHexString(encodedSha256hex);
+        } catch (final NoSuchAlgorithmException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private static String byteArrayToHexString(final byte[] bytes) {
+        final StringBuilder hexString = new StringBuilder(2 * bytes.length);
```
eshryane, line 160: Check the consequences of allocating a new `StringBuilder` that's double the size of the snapshot (which is 100's of megabytes). Maybe there is a more efficient way of calculating the SHA hash ?

whois-nrtm4-client/src/test/resources/mock/nrtm-snapshot.1.RIPE.json
```diff
@@ -0,0 +1,5 @@
+{"nrtm_version":4,"type":"snapshot","source":"RIPE","session_id":"4521174b-548f-4e51-98fc-dfd720011a0c","version":1}
```
eshryane, line 1: Are there NULL bytes in this file?

whois-commons/src/main/resources/nrtm_client_schema.sql
MiguelAHM: I think that storing the ID is something internal for us.
I am storing the object name instead of the object type.

I can use ObjectTypeIds and get the id of the object. But this means, that I am using something of whois that the client may be doesn't have (I would be using whois-common module)

eshryane: You should be using the whois schema to store RPSL objects, not duplicating it. How does the NRTMv3 client write RPSL into the mirror database?

whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/importer/SnapshotImporter.java
MiguelAHM: I did refactor this. I found a better solution related to performance.
I pushed the change

whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/importer/SnapshotImporter.java
```diff
@@ -0,0 +1,167 @@
...
+        try {
+            final MessageDigest digest = MessageDigest.getInstance("SHA-256");
+            final byte[] encodedSha256hex = digest.digest(bytes);
+            return byteArrayToHexString(encodedSha256hex);
+        } catch (final NoSuchAlgorithmException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private static String byteArrayToHexString(final byte[] bytes) {
```
eshryane, line 160: You can also use something off the shelf like `org.apache.commons.codec.binary.Hex.encodeHexString(byte[])`

whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/client/NrtmRestClient.java
MiguelAHM: now I don't have this issue.
I am reading in batches and parallelising those batches. I get the OOM issue if I try to parallelise everything.
I will play with the size of the batches

whois-nrtm4-client/src/test/resources/mock/nrtm-snapshot.1.RIPE.json
MiguelAHM: no, there are no nulls.
This is a copy paste of the uncompressed snapshot (taking just a few records)

whois-commons/src/main/resources/nrtm_client_schema.sql
MiguelAHM: maybe we can discuss this with Mahesh?

whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/importer/SnapshotImporter.java
MiguelAHM: I will try with the library org.apache.commons.codec.binary.Hex.encodeHexString(byte[])
That you pointed out. I am doing on this way because we are doing the same in IT for Nrtm v4

whois-commons/src/main/resources/nrtm_client_schema.sql
MiguelAHM: this will be done as part of [DB-5870](https://jira.ripe.net/browse/DB-5870)

whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/client/NrtmRestClient.java
```diff
@@ -82,6 +86,21 @@ public UpdateNotificationFileResponse getNotificationFile(final String source){
...
+        try {
+            final Response response =  client.target(url)
+                    .request(MediaType.APPLICATION_OCTET_STREAM)
+                    .header(HttpHeader.X_FORWARDED_PROTO.asString(), HttpScheme.HTTPS.asString())
+                    .get(Response.class);
+
+            return response.readEntity(byte[].class);
+        } catch (Exception ex){
+            LOGGER.error("Unable to get the records from the snapshot", ex);
+            return null;
```
eshryane, line 100: throw exception?

eshryane: Doesn't matter, null is handled by caller

whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/dao/Nrtm4ClientMirrorRepository.java
```diff
@@ -34,13 +39,20 @@ public Nrtm4ClientMirrorRepository(@Qualifier("nrtmClientMasterDataSource") fina
...
+                                        final String sessionID){
+        saveVersionInfo(source, version, sessionID, NrtmClientDocumentType.SNAPSHOT);
     }
 
-    public List<NrtmClientVersionInfo> getNrtmLastVersionInfo(){
+    public List<NrtmClientVersionInfo> getNrtmLastVersionInfoForUpdateNotificationFile(){
         final String sql = """
             SELECT id, source, MAX(version), session_id, type, created
             FROM version_info
+            WHERE type = 'update-notification-file'
```
eshryane, line 55: Use '?' for values

whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/dao/Nrtm4ClientMirrorRepository.java
```diff
@@ -49,11 +61,24 @@ SELECT id, source, MAX(version), session_id, type, created
...
+    public void persistRpslObject(final RpslObject rpslObject){
+        try {
+            final long now = JdbcRpslObjectOperations.now(dateTimeProvider);
+            jdbcMasterTemplate.update("INSERT INTO last_mirror (object, object_type, pkey, timestamp) VALUES (?, ?, ?, ?)",
+                    getRpslObjectBytes(rpslObject),
+                    ObjectTypeIds.getId(rpslObject.getType()),
+                    rpslObject.getKey().toString(),
+                    now);
+        } catch (IOException e) {
+            LOGGER.error("unable to get the bytes of the object {}", rpslObject.getKey(), e);
```
eshryane, line 78: Separate `getRpslObjectBytes` into variable and catch `IOException` separately

whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/importer/SnapshotImporter.java
```diff
@@ -0,0 +1,132 @@
...
+        }, 0, 10000);
+    }
+
+    private static String calculateSha256(final byte[] bytes) {
+        try {
+            final MessageDigest digest = MessageDigest.getInstance("SHA-256");
+            final byte[] encodedSha256hex = digest.digest(bytes);
+            return encodeHexString(encodedSha256hex);
+        } catch (final NoSuchAlgorithmException e) {
+            throw new RuntimeException(e);
```
eshryane, line 103: Throw specific exception

eshryane (approved):

whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/dao/Nrtm4ClientMirrorRepository.java
MiguelAHM: I return an error as well, because if not maybe some objects are added and some of them are not...
That would generate an incomplete snapshot

MiguelAHM:
test OK
https://gitlab.ripe.net/swe-database-team/whois/-/pipelines/743761

PR #1581, opened 2024-11-11
Fix Replicas issue of Deprecation

MiguelAHM:
Best Practice:
- Development/Testing: You may want to leave deprecation warnings at a lower log level (like INFO or DEBUG) to catch any issues early.
- Production: Setting it to ERROR could be useful to reduce log noise but should be monitored periodically to ensure no important warnings are missed.


whois-api/src/main/java/net/ripe/db/whois/api/fulltextsearch/ElasticFullTextRebuild.java
```diff
@@ -174,6 +176,13 @@ private void createIndex(final String indexName) throws IOException {
         request.mapping(getMappings());
 
         elasticIndexService.getClient().indices().create(request, RequestOptions.DEFAULT);
+
+        final ClusterUpdateSettingsRequest clusterUpdateSettingsRequest = new ClusterUpdateSettingsRequest();
+        Settings persistentSettings = Settings.builder()
+                .put("logger.deprecation.level", "ERROR")
```
eshryane: Will we miss out on warnings about upcoming deprecated features ?

eshryane: You committed as I commented, log level is now configurable

eshryane (approved):

fmurodov (approved):

PR #1589, opened 2024-11-18
add new column for pem format

maggarwal13:
Add new column pem_format so we can manually add when key is generated automatically
Send alerts if next key is available but no pem format


eshryane:
As discussed we will remove the original column later so we will have only one column to store the key

eshryane (approved):

PR #1591, opened 2024-11-19
Method Not Allowed When 405 Instead of 500

MiguelAHM:


eshryane (approved):

whois-api/src/main/java/net/ripe/db/whois/api/nrtm4/NrtmExceptionMapper.java
```diff
@@ -26,6 +27,10 @@ public Response toResponse(final Exception exception) {
             return createErrorResponse(HttpServletResponse.SC_NOT_FOUND, exception.getMessage());
         }
```
maggarwal13, line 29: Is it also being logged as error in sentry ?

MiguelAHM: I don't see any alert in sentry and I've just run the curl there

PR #1592, opened 2024-11-19
Fallback to Original-recipient if invalid address in Final-recipient field in bounced mail message

eshryane:
Fallback to Original-recipient if invalid address in Final-recipient field in bounced mail message.

Also ignore invalid angle-brackets in email address (not part of addr-spec according to RFC 822).


eshryane:
We should *not* fallback to the Original-Recipient as it may be an alias. We don't want to block an entire alias (e.g. mailing list), the Original-Recipient, because mail could not be delivered to one member (the Final-Recipient).

PR #1593, opened 2024-11-20
Configure auth client to use Jersey JDK connector.

eshryane:
Our JAX-RS client (Jersey) does not support the PATCH HTTP method by default.

More information : https://github.com/eclipse-ee4j/jersey/issues/4825

This comment is interesting in particular:
"Since Jersey 3.1, we do have the JNH connector (based on java.net.http package classes, that allow the HTTP Patch). Unfortunately, these classes prohibit some other features the default JDK HttpUrlConnector supports."
Ref. https://github.com/eclipse-ee4j/jersey/issues/4825#issuecomment-1983198605

This PR is a test if switching to the JNH connector is feasible for us.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -572,6 +572,11 @@
                 <artifactId>jersey-hk2</artifactId>
                 <version>${jersey.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.glassfish.jersey.connectors</groupId>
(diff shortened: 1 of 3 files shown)


eshryane:
Test successful, closing this PR (changes will be used in API key client).

PR #1597, opened 2024-11-29
Apikey authentication query

maggarwal13:


whois-client/src/main/java/net/ripe/db/whois/common/apiKey/OAuthSession.java
```diff
@@ -63,4 +63,31 @@ public String toString() {
                 .add("scopes", scopes)
```
MiguelAHM, line 63: why this class is not a record?

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/transform/FilterAuthFunction.java
```diff
@@ -106,34 +112,40 @@ public RpslObject apply(final RpslObject rpslObject) {
...
-        if (CollectionUtils.isEmpty(passwords) && StringUtils.isBlank(token) && (certificates == null || certificates.isEmpty())) {
+        if (CollectionUtils.isEmpty(passwords) && StringUtils.isBlank(token) && (certificates == null || certificates.isEmpty()) && (oAuthSession == null || oAuthSession.getUuid() == null)) {
             return false;
         }
 
+        final List<RpslObject> maintainers = getMaintainers(rpslObject);
+
         final List<RpslAttribute> extendedAuthAttributes = Lists.newArrayList(rpslObject.findAttributes(AttributeType.AUTH));
-        extendedAuthAttributes.addAll(getMntByAuthAttributes(rpslObject));
+        extendedAuthAttributes.addAll(getMntByAuthAttributes(maintainers));
```
MiguelAHM, line 122: Shouldn't be better to generate another list with mntners? I mean you are affecting rest of authentication methods adding mnt-by attributes
Shouldn't be any side effect because mnt-by format doesn't match the regex in each of them...however I think that it is safer to manage a different list ad-hoc for apikeyAuthenticated
something like:
`
final List<RpslAttribute> extendedAuthAttributes = Lists.newArrayList(rpslObject.findAttributes(AttributeType.AUTH));
`
`
final List<RpslAttribute> apikeyExtendedAttributes =  extendedAuthAttributes.addAll(getMntByAuthAttributes(maintainers));
`
`
apiKeyAuthenticated(rpslObject, maintainers, apikeyExtendedAttributes)
`

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/transform/FilterAuthFunction.java
```diff
@@ -172,4 +184,43 @@ private boolean passwordAuthentication(final List<RpslAttribute> authAttributes)
...
+                    return false;
+                }
+            }
+        }
+
+        return false;
+    }
+
+    private static boolean validateScope(final List<RpslObject> maintainers, final OAuthSession.ScopeFormatter scopeFormatter) {
+        return scopeFormatter.getAppName().equalsIgnoreCase("whois")
```
MiguelAHM, line 221: are we ignoring the fourth part of the scope?
It will always be "write" right now.
Maybe we should check it...just for being consistent?
I don't know, I leave it to you

whois-client/src/main/java/net/ripe/db/whois/common/apiKey/OAuthSession.java
maggarwal13: Why you want to make it a record? This will be JSON parsed  once proxy layer is implemented.

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/transform/FilterAuthFunction.java
maggarwal13: Yes , it will not work as I need to add the dependency whois-update on whois-query which I don't want and this does not belong to whois-common.I am not sure, I can move the whole APIKeyValidator to whois common but that will also not work as there are lot of implementation of that interface so need to do lot of refactoring. We are doing the same thing ClientAuthCertificateValidator so I just did the same for consitency.

maggarwal13: May be I can do other way around. Let me try

whois-client/src/main/java/net/ripe/db/whois/common/apiKey/OAuthSession.java
MiguelAHM: no, I don't care if it is a record or not.
Just wander why it is no a record...

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/transform/FilterAuthFunction.java
MiguelAHM: Thanks for the change

MiguelAHM (approved):

PR #1601, opened 2024-12-02
Use query param apikey

maggarwal13:


MiguelAHM:
I don't understand the tittle.
It is not a query parameter or at least I can see @QueryParam in the controller.
It seems that the controller is getting that OauthSession from header

MiguelAHM:
> I don't understand the tittle. It is not a query parameter or at least I can't see @QueryParam in the controller. It seems that the controller is getting that OAuthSession from header and putting it in the UpdateContext which is OK I think

discussed via chat

MiguelAHM (approved):

maggarwal13:
Use bearer token header

PR #1603, opened 2024-12-03
RDAP Relation Searches Endpoint

MiguelAHM:
This code is just for doing the controller changes.
Link relation and Extensions are not contemplated in this PR

Bottom relation pseudocode in the Jira

The status query parameter is not implemented until we have implemented administrative status task:
When status is there and inactive we should return the in region administrative resources. 
In case status is active we should return non administrative in region resources.
For now just always return active resource. Ignore status query parameter. (added a TODO)

We follow the points addressed in https://datatracker.ietf.org/doc/draft-ietf-regext-rdap-rir-search/ RFC


whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,199 @@
...
+        return ipEntries
+                .stream()
+                .map(ipEntry -> rpslObjectDao.getById(ipEntry.getObjectId()).getKey().toString())
+                .toList();
+    }
+
+    public List<String> getInetnumRelationPkeys(final String pkey, final RelationType relationType){
+        final IpInterval ip = IpInterval.parse(pkey);
+        final List<IpEntry> ipEntries = getIpEntries(getIpTree(ip), relationType, ip);
+        return ipEntries.stream().map(ipEntry -> ipEntry.getKey().toString()).toList();
```
MiguelAHM, line 68: format this consistently

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    private List<IpEntry> getIpEntries(final IpTree ipTree, final RelationType relationType,
+                                       final IpInterval reverseIp) {
+        return switch (relationType) {
+            case UP -> List.of(searchFirstLessSpecificCoMntner(ipTree, reverseIp));
+            case TOP -> searchCoMntnerTopLevel(ipTree, reverseIp);
```
MiguelAHM, line 75: why coMntner

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,199 @@
...
+
+    public List<String> getInetnumRelationPkeys(final String pkey, final RelationType relationType){
+        final IpInterval ip = IpInterval.parse(pkey);
+        final List<IpEntry> ipEntries = getIpEntries(getIpTree(ip), relationType, ip);
+        return ipEntries.stream().map(ipEntry -> ipEntry.getKey().toString()).toList();
+    }
+
+    private List<IpEntry> getIpEntries(final IpTree ipTree, final RelationType relationType,
+                                       final IpInterval reverseIp) {
+        return switch (relationType) {
```
MiguelAHM, line 73: should we have a default? or not...maybe create a default returning a not implemented?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,199 @@
...
+            case BOTTOM -> searchMostSpecificFillingOverlaps(ipTree, reverseIp);
+        };
+    }
+
+    private List<IpEntry> searchMostSpecificFillingOverlaps(final IpTree ipTree, final IpInterval reverseIp){
+        final List<IpEntry> mostSpecificValues = ipTree.findMostSpecific(reverseIp);
+        final List<? extends IpInterval<?>> ipResources = mostSpecificValues.stream().map(ip -> IpInterval.parse(ip.getKey().toString())).toList();
+        final Set<IpEntry> mostSpecificFillingOverlaps = Sets.newConcurrentHashSet();
+
+        for (int countIps = 0; countIps < mostSpecificValues.size(); countIps++){
```
MiguelAHM, line 86: review this logic, why are we using a for loop with a counter...we could just iterate the values

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapServiceTestIntegration.java
```diff
@@ -3127,6 +3128,422 @@ public void get_help_response() {
...
+        assertThat(domainResults.getFirst().getHandle(), is("1.2.0.192.in-addr.arpa"));
+        assertThat(domainResults.get(1).getHandle(), is("2.0.192.in-addr.arpa"));
+    }
+
+    @Test
+    public void get_non_existing_domain_bottom_then_empty(){
+        loadIpv4RelationTreeExample();
+        loadIpv4RelationDomainExample();
+
+        final SearchResult searchResult = createResource("domains/rirSearch1/bottom/1.2.0.192.in-addr.arpa")
```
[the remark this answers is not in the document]
MiguelAHM: the answer is to return an empty result
from the rfc 

``` 
 If there are no more-specific objects for a given INR object value,
   then the set of bottom objects for that INR object value will be
   empty.  192.0.2.0/32 is an example of such an INR object value.
```

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapServiceTestIntegration.java
```diff
@@ -3127,6 +3128,422 @@ public void get_help_response() {
...
+
+        final List<Ip> ipResults = searchResult.getIpSearchResults();
+        assertThat(ipResults, is(nullValue()));
+    }
+
+    @Test
+    public void down_with_status_then_501(){
+
+        final ServerErrorException notImplementedException = assertThrows(ServerErrorException.class, () -> {
+            createResource("ip/rirSearch1/down/192.0.2.0/24?status=inactive")
```
MiguelAHM, line 3536: this should be returning error for the type (invalid type)
use ips in the example

whois-rpsl/src/main/java/net/ripe/db/whois/common/rpsl/attrs/Inet6numStatus.java
```diff
@@ -80,6 +80,11 @@ public boolean worksWithParentInHierarchy(final InetStatus parentInHierarchyMain
         return true;
     }
 
+    @Override
+    public boolean isAdministrativeResource() {
```
MiguelAHM, line 84: move the logic

whois-rpsl/src/main/java/net/ripe/db/whois/common/rpsl/attrs/InetnumStatus.java
```diff
@@ -107,6 +107,11 @@ public boolean worksWithParentInHierarchy(final InetStatus parentInHierarchyMain
         return true;
     }
 
+    @Override
+    public boolean isAdministrativeResource() {
```
MiguelAHM, line 111: move the logic to the Relation service

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
MiguelAHM: this was a change that I did at the beginning. I forgot to rename this method

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationTypeConverter.java
```diff
@@ -0,0 +1,34 @@
...
+
+    @Override
+    public <T> ParamConverter<T> getConverter(Class<T> rawType, Type type, Annotation[] annotations) {
+        if(!rawType.equals(RelationType.class)) {
+            return null;
+        }
+
+        return (ParamConverter<T>) new ParamConverter<RelationType>() {
+            @Override
+            public RelationType fromString(String relationType){
```
maggarwal13, line 24: is it possible to add a toString method in enum for this ?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapService.java
```diff
@@ -245,6 +250,38 @@ public Response help(@Context final HttpServletRequest request) {
...
+            @Context final HttpServletRequest request,
+            @PathParam("objectType") RdapRequestType requestType,
+            @PathParam("relation") RelationType relationType,
+            @PathParam("key") final String key,
+            @QueryParam("status") String status) {
+
+        //TODO: [MH] Status is being ignored until administrative resources are included in RDAP. If status is not
+        // given or status is inactive...include administrative resources in the output. However, if status is active
+        // return just non administrative resources, as we are doing now.
+        if (status != null && (relationType.equals(RelationType.DOWN) || relationType.equals(RelationType.BOTTOM))){
```
maggarwal13, line 266: use stringUtils isEmpty method to check for empty string as well

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapService.java
```diff
@@ -245,6 +250,38 @@ public Response help(@Context final HttpServletRequest request) {
...
+            throw new RdapException("501 Not Implemented", "Status is not implement in down and bottom relation", HttpStatus.NOT_IMPLEMENTED_501);
+        }
```
maggarwal13, line 268: should we just check if status is inactive and UP and DOWN? as by default we do return active status , also add a notification status is ignored rather than sending error?

whois-commons/src/main/java/net/ripe/db/whois/common/etree/IntervalMap.java
```diff
@@ -161,4 +161,20 @@ public interface IntervalMap<K extends Interval<?>, V> {
      *         intervals.
      */
     List<V> findExactAndAllMoreSpecific(K key);
+
+    /**
+     * Finds all values associated with intervals that are contained within (more specific than)
+     * and whose prefix is higher, more specific
```
maggarwal13, line 167: Isn't all prefix in more specific will be higher than interval?  Should it be whose prefix is most higher ?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationTypeConverter.java
MiguelAHM: check RdapRequestTypeConverter
I am following the same logic.

so I suggest leaving in this way for consistency. Maybe we can change it all together?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapService.java
MiguelAHM: true, I always forget about the stringUtils...

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationTypeConverter.java
maggarwal13: No, we are creating a new one so we should do it correctly.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapService.java
MiguelAHM: I added in the README. So I think that it should be enough. Community should know that for now we only consider active status

Relation Searches doesn't consider the "status" query parameter
-----------------------------------------------------------------
This is related to the previous point. An object with "administrative" status is never returned.
Therefore, only "active" (non-administrative) objects are taking into account when using relation searches.

maggarwal13: Yes but if a user is sending an active status and we are supporting it , we should return the object. It is inactive we are not supporting. No?

MiguelAHM: In addition
The task says: Return "501" if status is combined with "down" or "bottom" relation

maggarwal13: yes we should return 501 only if status is inactive

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,202 @@
...
+            return rpslObjectDao.getByKeyOrNull(ObjectType.INETNUM, key);
+        }
+        return rpslObject;
+    }
+
+    private IpTree getIpTree(final IpInterval reverseIp) {
+        if (reverseIp instanceof Ipv4Resource) {
+            return ip4Tree;
+        } else if (reverseIp instanceof Ipv6Resource) {
+            return ip6Tree;
```
maggarwal13, line 187: No need for else if ? use ternary operator. There are only two impelmentations of IpInterval

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,202 @@
...
+        }
+
+        throw new IllegalArgumentException("Unexpected reverse ip: " + reverseIp);
+    }
+
+    private IpTree getIpDomainTree(final IpInterval reverseIp) {
+        if (reverseIp instanceof Ipv4Resource) {
+            return ipv4DomainTree;
+        } else if (reverseIp instanceof Ipv6Resource) {
+            return ipv6DomainTree;
```
maggarwal13, line 197: same as above comment

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationTypeConverter.java
MiguelAHM: OK, I moved into the enum

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapService.java
```diff
@@ -486,6 +529,34 @@ private String objectTypesToString(final Collection<ObjectType> objectTypes) {
         return COMMA_JOINER.join(objectTypes.stream().map(ObjectType::getName).toList());
     }
 
+    private List<RpslObject> handleRelationQuery(final HttpServletRequest request, final RdapRequestType requestType,
+                                                 final RelationType relationType, final String key) {
```
maggarwal13, line 533: Do you think it makes more sense to have a separate  controller for rirsearch ?

MiguelAHM: I don't think so because there is always one endpoint and it is part of RDAP

whois-commons/src/main/java/net/ripe/db/whois/common/etree/IntervalMap.java
```diff
@@ -161,4 +161,20 @@ public interface IntervalMap<K extends Interval<?>, V> {
...
+     * <p/>
+     * The resulting values are the most specific values associated with the
+     * <code>key</code>.
+     *
+     * @param key the key to find the exact and all levels more specific values
+     *            for
+     * @return the (possibly empty) list of values associated with the matching
+     *         intervals.
+     */
+    List<V> findMostSpecific(K key);
```
maggarwal13, line 179: Should it be list of just one element as it is most specific?

whois-commons/src/main/java/net/ripe/db/whois/common/etree/IntervalMap.java
```diff
@@ -161,4 +161,20 @@ public interface IntervalMap<K extends Interval<?>, V> {
      *         intervals.
      */
     List<V> findExactAndAllMoreSpecific(K key);
+
+    /**
+     * Finds all values associated with intervals that are contained within (more specific than)
+     * and whose prefix is most high
+     * <code>key</code>.
```
maggarwal13, line 168: Is it possible to use findFirstMoreSpecific ? It says " The resulting values are ordered from least specific interval to most specific interval"

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,202 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(RdapRelationService.class);
+
+    private final Ipv4Tree ip4Tree;
+    private final Ipv6Tree ip6Tree;
+    private final Ipv4DomainTree ipv4DomainTree;
+    private final Ipv6DomainTree ipv6DomainTree;
+    private final RpslObjectDao rpslObjectDao;
+
+    public RdapRelationService(final Ipv4Tree ip4Tree, final Ipv6Tree ip6Tree,
```
maggarwal13, line 47: use autowired for consistency?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapService.java
maggarwal13: Thanks it's just that this class is getting so big, infact we pu so many logic in controller level but I agree it is been like this forever

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,202 @@
...
+
+    private List<IpEntry> searchMostSpecificFillingOverlaps(final IpTree ipTree, final IpInterval reverseIp){
+        final List<IpEntry> mostSpecificValues = ipTree.findMostSpecific(reverseIp);
+        final List<? extends IpInterval<?>> ipResources = mostSpecificValues.stream().map(ip -> IpInterval.parse(ip.getKey().toString())).toList();
+        final Set<IpEntry> mostSpecificFillingOverlaps = Sets.newConcurrentHashSet();
+        ipResources.forEach(ipResource -> processChildren(ipTree, reverseIp, ipResource, mostSpecificFillingOverlaps));
+        return mostSpecificFillingOverlaps.stream().toList();
+    }
+
+    private static void processChildren(final IpTree ipTree, final IpInterval reverseIp,
```
maggarwal13, line 95: does this needs to be static?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,202 @@
...
+    private IpEntry searchFirstLessSpecific(final IpTree ipTree, final IpInterval reverseIp){
+        final List<IpEntry> parentList = ipTree.findFirstLessSpecific(reverseIp);
+        if (parentList.isEmpty() || !resourceExist(parentList.getFirst())){
+            throw new RdapException("404 Not Found", "No up level object has been found for " + reverseIp.toString(), HttpStatus.NOT_FOUND_404);
+        }
+        return parentList.getFirst();
+    }
+
+    private List<IpEntry> searchTopLevel(final IpTree ipTree, final IpInterval reverseIp) {
+        for (final Object parentEntry : ipTree.findAllLessSpecific(reverseIp)) {
```
maggarwal13, line 149: Can you use stream and filter findAny ?

whois-commons/src/main/java/net/ripe/db/whois/common/etree/IntervalMap.java
MiguelAHM: List
there can be a lot of most specific. It is the last leaf of the tree but the tree could have more than one branch

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,202 @@
...
+                return List.of(ipEntry);
+            }
+        }
+        throw new RdapException("404 Not Found", "No top level object has been found for " + reverseIp.toString(), HttpStatus.NOT_FOUND_404);
+    }
+
+    private boolean resourceExist(final IpEntry firstLessSpecific){
+        final RpslObject rpslObject = getResourceByKey(firstLessSpecific.getKey().toString());
+        if (rpslObject == null || isAdministrativeResource(rpslObject)) {
+            LOGGER.error("INET(6)NUM {} does not exist in RIPE Database ", firstLessSpecific.getKey().toString());
```
maggarwal13, line 161: do we need to log as error?

whois-commons/src/main/java/net/ripe/db/whois/common/etree/IntervalMap.java
maggarwal13: Ok got it we are getting the node with no children. Thanks

whois-commons/src/main/java/net/ripe/db/whois/common/etree/IntervalMap.java
maggarwal13: Discussed in previous comment

MiguelAHM: noup
We need to hit the last leafs of the tree. `findFirstMoreSpecific` will give you just the first more specific
We need to get the last more specific lets say...we need to get the children without children (last leafs of the tree)

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
MiguelAHM: nop, it turned into debug

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
MiguelAHM: no needed

whois-commons/src/main/java/net/ripe/db/whois/common/etree/IntervalMap.java
MiguelAHM: I gave you an example in slack

whois-commons/src/main/java/net/ripe/db/whois/common/etree/IntervalMap.java
MiguelAHM: I gave you an example in slack

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
MiguelAHM: I am filling mostSpecificFillingOverlaps list processing the leafs
It is not an easy name

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
MiguelAHM: done

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
MiguelAHM: done

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
MiguelAHM: Done

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,202 @@
...
+    private static boolean isAdministrativeResource(final RpslObject rpslObject) {
+        final CIString statusAttributeValue = rpslObject.getValueForAttribute(AttributeType.STATUS);
+        return (rpslObject.getType().equals(ObjectType.INETNUM) && InetnumStatus.getStatusFor(statusAttributeValue).equals(ALLOCATED_UNSPECIFIED))
+                || (rpslObject.getType().equals(ObjectType.INET6NUM) && Inet6numStatus.getStatusFor(statusAttributeValue).equals(ALLOCATED_BY_RIR));
+    }
+
+
+    @Nullable
+    private RpslObject getResourceByKey(final String key){
+        final RpslObject rpslObject = rpslObjectDao.getByKeyOrNull(ObjectType.INET6NUM, key);
```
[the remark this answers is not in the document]
MiguelAHM: Done

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
MiguelAHM: Done

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapService.java
MiguelAHM: done

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationTypeConverter.java
MiguelAHM: done

whois-commons/src/main/java/net/ripe/db/whois/common/etree/IntervalMap.java
MiguelAHM: changed

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,213 @@
...
+
+    private List<IpEntry> searchTopLevelResource(final IpTree ipTree, final IpInterval reverseIp){
+        IpEntry ipEntry;
+        try {
+            ipEntry = searchFirstLessSpecific(ipTree, reverseIp);
+        } catch (RdapException ex){
+            throw new RdapException("404 Not Found", "No top-level object has been found for " + reverseIp.toString(), HttpStatus.NOT_FOUND_404);
+        }
+
+        return List.of(loopUpLevels(ipTree, ipEntry));
```
maggarwal13, line 158: put the return statement inside try block

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,213 @@
...
+                || (rpslObject.getType().equals(ObjectType.INET6NUM) && !Inet6numStatus.getStatusFor(childrenStatus).isAssignment() &&
+                Inet6numStatus.getStatusFor(statusAttributeValue).equals(ALLOCATED_BY_RIR));
+    }
+
+
+    @Nullable
+    private RpslObject getResourceByKey(final String key){
+        if (IpInterval.parse(key) instanceof Ipv4Resource){
+            return rpslObjectDao.getByKeyOrNull(ObjectType.INETNUM, key);
+        }
```
maggarwal13, line 196: parse the key and then hit the database

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,213 @@
...
+    private IpEntry searchFirstLessSpecific(final IpTree ipTree, final IpInterval reverseIp){
+        final List<IpEntry> parentList = ipTree.findFirstLessSpecific(reverseIp);
+        if (parentList.isEmpty() || !resourceExist(reverseIp, parentList.getFirst())){
+            throw new RdapException("404 Not Found", "No up level object has been found for " + reverseIp.toString(), HttpStatus.NOT_FOUND_404);
+        }
+        return parentList.getFirst();
+    }
+
+    private List<IpEntry> searchTopLevelResource(final IpTree ipTree, final IpInterval reverseIp){
+        IpEntry ipEntry;
```
maggarwal13, line 151: the loopUpLevel and searchTopLevelResource doing same thing ?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,213 @@
...
+        } catch (RdapException ex){
+            /*
+            * Do Nothing, end of loop
+            * */
+        }
+        return reverseIp;
+    }
+
+    private boolean resourceExist(final IpInterval ip, final IpEntry firstLessSpecific){
+        final RpslObject children = getResourceByKey(ip.toString());
```
maggarwal13, line 174: you are sending toString and again parsing it back to ipinterval in the calling method?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,213 @@
...
+        final RpslObject rpslObject = getResourceByKey(firstLessSpecific.getKey().toString());
+        if (rpslObject == null || isAdministrativeResource(children, rpslObject)) {
+            LOGGER.debug("INET(6)NUM {} does not exist in RIPE Database ", firstLessSpecific.getKey().toString());
```
maggarwal13, line 177: can children be null to ?

whois-rpsl/src/main/java/net/ripe/db/whois/common/rpsl/attrs/Inet6numStatus.java
```diff
@@ -99,4 +99,8 @@ public boolean isValidOrgType(final OrgType orgType) {
     public String toString() {
         return literalStatus.toString();
     }
+
+    public boolean isAssignment(){
```
maggarwal13, line 103: If you are adding this function in Inet6num , it should also be added in Inetnum ?

whois-rpsl/src/main/java/net/ripe/db/whois/common/rpsl/attrs/Inet6numStatus.java
```diff
@@ -99,4 +99,8 @@ public boolean isValidOrgType(final OrgType orgType) {
...
+        return this.equals(ASSIGNED) || this.equals(ASSIGNED_ANYCAST) || this.equals(ASSIGNED_PI);
+    }
```
maggarwal13, line 105: There is a attribute in ManagedAttributeSearch class : INET6NUM_ASSIGNMENT_STATUSES. We are not including  status ASSIGNED there but here we are which one is correct?

whois-commons/src/main/java/net/ripe/db/whois/common/etree/NestedIntervalMap.java
```diff
@@ -226,6 +232,17 @@ private List<InternalNode<K, V>> internalFindFirstMoreSpecific(K range) {
         return result;
     }
 
+    private List<InternalNode<K, V>> internalFindMostSpecific(K range){
+        final List<InternalNode<K, V>> result = internalFindAllMoreSpecific(range);
+        if (result.isEmpty()){
+            return Lists.newArrayList();
```
maggarwal13, line 238: No need for this check if you are using stream

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,213 @@
...
+            return ipv4Resource.begin() == parentIpv4Resource.begin() && lastIpv4Resource.end() == parentIpv4Resource.end();
+        }
+
+        final Ipv6Resource ipv6Resource = (Ipv6Resource)firstResource;
+        final Ipv6Resource lastIpv6Resource = (Ipv6Resource)lastResource;
+        final Ipv6Resource parentIpv6Resource = (Ipv6Resource)parent;
+        return Objects.equals(ipv6Resource.begin(), parentIpv6Resource.begin()) && Objects.equals(lastIpv6Resource.end(), parentIpv6Resource.end());
+    }
+
+    private static List<IpEntry> findSiblingsAndExact(final IpTree ipTree, final IpInterval parentResource) {
```
maggarwal13, line 134: You are passing the mostSpecific resource right? Variable name cannot be parentResource?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,213 @@
...
+        final List<IpEntry> parentList = ipTree.findFirstLessSpecific(parentResource);
+        if (parentList.isEmpty()){
+            return ipTree.findExact(parentResource);
+        }
+        return ipTree.findFirstMoreSpecific(IpInterval.parse(parentList.getFirst().getKey().toString()));
```
maggarwal13, line 139: Ok this is findExactorFirstMoreSpecific ? If that's the case can you add a method in IntervalMap? As there is already a method for findExactOrFirstLessSpecific there

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,213 @@
...
+                .stream()
+                .map(ipEntry -> ipEntry.getKey().toString())
+                .toList();
+    }
+
+    private List<IpEntry> getIpEntries(final IpTree ipTree, final RelationType relationType,
+                                       final IpInterval reverseIp) {
+        return switch (relationType) {
+            case UP -> List.of(searchFirstLessSpecific(ipTree, reverseIp));
+            case TOP -> searchTopLevelResource(ipTree, reverseIp);
```
maggarwal13, line 83: TOP should be single resource like UP, not a list? 
" If the server receives a search containing the relation value "top",
   it will return the top object for the specified INR object value as
   though that object had been requested directly. "

maggarwal13: Also TOP should be getLast from findAllLessSpecific?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,213 @@
...
+            case DOWN -> ipTree.findFirstMoreSpecific(reverseIp);
+            case BOTTOM -> searchMostSpecificFillingOverlaps(ipTree, reverseIp);
```
maggarwal13, line 85: Correct me if I am wrong. For BOTTOM objects : FindAllMoreSpecific and then for each entry in it findFirstMoreSpecific. If the size of findFirstMoreSpecific is less than 2 you add it in the list else ignore

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    public List<String> getDomainRelationPkeys(final String pkey, final RelationType relationType){
+        final Domain domain = Domain.parse(pkey);
+        final IpInterval reverseIp = domain.getReverseIp();
+        final List<IpEntry> ipEntries = getIpEntries(getIpDomainTree(reverseIp), relationType, reverseIp);
+
+        return ipEntries
+                .stream()
+                .map(ipEntry -> rpslObjectDao.getById(ipEntry.getObjectId()).getKey().toString())
```
maggarwal13, line 64: why you need to hit the database? Can this throw exception if object is deleted? I think even if deleted pkey will be available

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
MiguelAHM: talked over a call.
The amount of the children doesn't matter. We need to check if childrens covers the father.
If we go from up to bottom...we need to iterate all of them until we get the last leaf. We don't know if we can add the node or not until we process all the leafs.
If we go from bottom to up...we are already iterating the leaf, so we just need to add it

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
MiguelAHM: our top level is always administrative resource
for IPv4 is 0.0.0.0 - 255.255.255.255 always
for IPv6 is ::/0 always

For IPv4 is easy, we get the top level removing that one which has "ALLOCATED UNSPECIFIED" status
For IPv6 to detect the administrative resource is not that easy. An administrative resource is the one with "ALLOCATED-BY-RIR" status when the children is "ALLOCATED-BY-RIR" as well.

If we hit a resource with "ALLOCATED-BY-RIR" we cannot directly find the firstMoreSpecific, because they can be more than 1.
So we need to iterate and search for UP. If the Xresource is "ALLOCATED-BY-RIR" and his parent is "ALLOCATED-BY-RIR" as well...then the parent is an administrative resource and we return the Xresource.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
MiguelAHM: Discussed over call
Here I have the IP and the Domain's object ID
I am using that object ID to fetch the reverse IP...We are not going to implement a logic to get the reverse IP out of the IP now so I added a TODO

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
MiguelAHM: nop, I am passing the lessSpecific -> Parent

whois-commons/src/main/java/net/ripe/db/whois/common/etree/NestedIntervalMap.java
MiguelAHM: True

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
MiguelAHM: the children was the parent of the previous iteration.
So if anyone delete that entry from database in the same ms that someone is doing this call could happen yes.
I will add a validation just to cover all the possibilities

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
MiguelAHM: I cannot understand what you mean. Maybe I have already refactored this?
I am just doing searchIp.toString to search in database.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
MiguelAHM: There are some differences.

searchTopLevelResource -> Is doing first iteration and checking that at least there is one TOP element
loopUpLevels -> Is trying to get the TOP levels.

SearchTopLevelResource is getting the first UP entry and checking if no top-level object has found.
if lookUpLevels is not able to search anything then first UP is returned

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
MiguelAHM: I don't understand what you mean
I am already parsing and then hitting the database

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
MiguelAHM: Done

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,200 @@
...
+            throw new RdapException("404 Not Found", "No up level object has been found for " + searchIp.toString(), HttpStatus.NOT_FOUND_404);
+        }
+        return parentList.getFirst();
+    }
+
+    private IpEntry searchTopLevelResource(final IpTree ipTree, final IpInterval searchIp){
+        try {
+            IpEntry ipEntry = searchUpResource(ipTree, searchIp);
+            return searchTopRemainingResource(ipTree, ipEntry);
+        } catch (RdapException ex){
```
maggarwal13, line 143: SearchUp already throws this exception, you don't need to catch and throw again? also searchTopRemainingResource seems to be doing same thing

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,200 @@
...
+            throw new RdapException("404 Not Found", "No top-level object has been found for " + searchIp.toString(), HttpStatus.NOT_FOUND_404);
+        }
+    }
+
+    private IpEntry searchTopRemainingResource(final IpTree ipTree, IpEntry searchIp) {
+        try {
+            searchIp = searchUpResource(ipTree, (IpInterval) searchIp.getKey());
+            searchTopRemainingResource(ipTree, searchIp);
+        } catch (RdapException ex){
+            /*
```
maggarwal13, line 153: RdapException is very generic , I think you meant to check for  404 exception. Do you need to check status code as well?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
maggarwal13: Sorry I must have looked at it wrongly

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,200 @@
...
+        final RpslObject rpslObject = getResourceByKey(firstLessSpecific.getKey().toString());
+        if (children == null || rpslObject == null || isAdministrativeResource(children, rpslObject)) {
+            LOGGER.debug("INET(6)NUM {} does not exist in RIPE Database ", firstLessSpecific.getKey().toString());
+            return false;
+        }
+        return true;
+    }
+
+    private static boolean isAdministrativeResource(final RpslObject children, final RpslObject rpslObject) {
+        final CIString childrenStatus = children.getValueForAttribute(AttributeType.STATUS);
```
maggarwal13, line 171: rename to childStatus as it is singular

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,200 @@
...
+        } catch (RdapException ex){
+            /*
+            * Do Nothing, end of loop
+            * */
+        }
+        return searchIp;
+    }
+
+    private boolean existAndNoAdministrative(final IpInterval searchIp, final IpEntry firstLessSpecific){
+        final RpslObject children = getResourceByKey(searchIp.toString());
```
maggarwal13, line 161: You can pass IPInterval directly rather then converting it to string and then again converting to IPInterval?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,200 @@
...
+
+    private IpEntry searchUpResource(final IpTree ipTree, final IpInterval searchIp){
+        final List<IpEntry> parentList = ipTree.findFirstLessSpecific(searchIp);
+        if (parentList.isEmpty() || !existAndNoAdministrative(searchIp, parentList.getFirst())){
+            throw new RdapException("404 Not Found", "No up level object has been found for " + searchIp.toString(), HttpStatus.NOT_FOUND_404);
+        }
+        return parentList.getFirst();
+    }
+
+    private IpEntry searchTopLevelResource(final IpTree ipTree, final IpInterval searchIp){
```
maggarwal13, line 139: I think if findAllLessSpecific getLast and its firstLessSpecific is already non administrative we can retun directly, instead of traversing whole tree because it can be really huge and its higher probability that you will  find within 2 levels

maggarwal13 (approved):

PR #1605, opened 2024-12-10
Fix Rollback NRTMv4 Client

MiguelAHM:
- When using "parallel()" each record will use one transaction, so rollback is not possible because each record is being committed.
So I have to refactor to use ExecutorService and manually manage the transactions. I am creating a transaction commit per batch.
So there could be an issue with the third batch, when two previous batch have been already committed. So in this case the third batch will rollback and there will be a truncate tables. So the database will be clean.

- For rest of the situations I relay in the transaction. If deltas fails the transaction will take place
- Refactor to persist the version at the end of the import. I am using a different database, so a different datasource for version. So I cannot rollback if I add the snapshot before processing the records and there is a problem with processing the records (with the current setup, transaction chains are no set). So I order the code for using first one datasource (doing the import) and the another (add version) and not in the other way around


whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/importer/SnapshotMirrorImporter.java
```diff
@@ -145,31 +145,59 @@ public void run() {
...
+                    final Map.Entry<RpslObject, RpslObjectUpdateInfo> persistedRecord = nrtm4ClientRepository.processSnapshotRecord(mirrorRpslObject);
+                    nrtm4ClientRepository.createIndexes(persistedRecord.getKey(), persistedRecord.getValue());
+                    processedCount.incrementAndGet();
+                } catch (Exception e) {
+                    nrtmClientUpdateTransaction.rollback(transactionStatus);
+                    LOGGER.error("Unable to process record {}", record, e);
+                    throw new IllegalStateException(e);
+                }
+            });
+        } catch (Exception e) {
```
eshryane, line 196: I don't think the outside try/catch is useful, unless you expect `submit()` to fail. The record is processed asynchronously. You could return the `Future` and check in the loop whether the job failed, and then stop submitting jobs if any failed.

eshryane: Or use a global `AtomicBoolean` to flag if any job failed and if so, stop all jobs immediately.

MiguelAHM: I go with future for handling errors. So I check for the errors

whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/importer/SnapshotMirrorImporter.java
```diff
@@ -62,32 +86,51 @@ public void doImport(final String source,
...
         }
 
+        final AtomicInteger snapshotVersion = new AtomicInteger(0);
         final AtomicInteger processedCount = new AtomicInteger(0);
+
+        persisSnapshot(source, payload, sessionId, snapshotVersion, processedCount);
+        persistSnapshotVersion(source, snapshotVersion.get(), sessionId);
+
+        stopwatch.stop();
+        LOGGER.info("Loading snapshot file took {} for source {} and added {} records", stopwatch.elapsed().toMillis(), source, processedCount.get());
```
eshryane, line 96: No need to convert snapshot to milliseconds, the `toString()` will always print something readable.

whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/importer/SnapshotMirrorImporter.java
```diff
@@ -33,16 +50,23 @@ public class SnapshotMirrorImporter extends AbstractMirrorImporter {
...
                                   final Nrtm4ClientInfoRepository nrtm4ClientInfoMirrorDao,
-                                  final Nrtm4ClientRepository nrtm4ClientRepository) {
+                                  final Nrtm4ClientRepository nrtm4ClientRepository,
+                                  @Qualifier(NRTM_CLIENT_UPDATE_TRANSACTION) final PlatformTransactionManager transactionManagerNrtmClientUpdate) {
         super(nrtm4ClientInfoMirrorDao, nrtm4ClientRepository);
         this.nrtmRestClient = nrtmRestClient;
+        this.nrtmClientUpdateTransaction =  transactionManagerNrtmClientUpdate;
 
+        final int numThreads = Runtime.getRuntime().availableProcessors();
+        this.numberOfThreads = (int)Math.ceil(numThreads/4.0);
```
eshryane, line 67: No need for a float, you can also use `Math.min(1, numThreads/4)`

whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/config/NrtmClientTransactionConfiguration.java
```diff
@@ -0,0 +1,33 @@
...
+
+import javax.sql.DataSource;
+
+
+@Configuration
+@Conditional(Nrtm4ClientCondition.class)
+@EnableTransactionManagement(mode = AdviceMode.ASPECTJ)
+public class NrtmClientTransactionConfiguration {
+
+    public static final String NRTM_CLIENT_UPDATE_TRANSACTION = "nrtm-client-update-transaction-manager";
```
eshryane, line 21: No need to define fields if value is only used once

whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/dao/Nrtm4ClientInfoRepository.java
```diff
@@ -64,7 +68,7 @@ SELECT id, source, MAX(version), session_id, type, hostname, created
     @Nullable
     public NrtmClientVersionInfo getNrtmLastVersionInfoForDeltasPerSource(final String source){
         try {
-            return jdbcMasterTemplate.queryForObject("""
+            return jdbcSlaveTemplate.queryForObject("""
```
eshryane, line 71: Why are you using the *slave* datasource here when elsewhere you're using the master datasource during import ? Is it possible that the slave datasource will be out-of-sync ?

whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/importer/SnapshotMirrorImporter.java
```diff
@@ -105,20 +148,72 @@ public void run() {
...
+                    processedCount.incrementAndGet();
+                } catch (Exception e) {
+                    nrtmClientUpdateTransaction.rollback(transactionStatus);
+                    LOGGER.error("Unable to process record {}", record, e);
+                    throw new IllegalStateException(e);
+                }
+            });
     }
 
     public static RpslObject getPlaceholderPersonObject() {
```
eshryane, line 219: This method has been deleted in a separate PR correct?

whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/dao/Nrtm4ClientInfoRepository.java
MiguelAHM: yes, I will use master

whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/importer/SnapshotMirrorImporter.java
```diff
@@ -1,29 +1,46 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import static net.ripe.db.nrtm4.client.config.NrtmClientTransactionConfiguration.NRTM_CLIENT_UPDATE_TRANSACTION;
+
 
 @Service
 @Conditional(Nrtm4ClientCondition.class)
```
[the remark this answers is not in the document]
MiguelAHM: yes, deltas and snapshot
and there is common test between them

whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/config/NrtmClientTransactionConfiguration.java
MiguelAHM: It is public, it is being in use by repository layer...
Is how TransactionConfiguration works as well

whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/importer/SnapshotMirrorImporter.java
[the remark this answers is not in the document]
eshryane: OK

whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/dao/Nrtm4ClientInfoRepository.java
MiguelAHM: Done

eshryane (approved):

PR #1606, opened 2024-12-12
RDAP Relation Search Links

MiguelAHM:
https://datatracker.ietf.org/doc/draft-ietf-regext-rdap-rir-search/
section 3.4

Each of the relations defined in section 3.2.2 has a corresponding
link relation that can be used for a link object contained within
another RDAP object. When constructing these link objects, the
server MUST use the corresponding search URL for the link target, or
a URL that yields the same response as for the corresponding search
as at the time of the request. The following is an elided example of
a response that makes use of those link relations

Take into account that some Extensions needs to be added in case the link is included. So it is necessary to add next extensions in following situations
- If response includes IP basic search links -> "rirSearch1" and "ips"
- If response includes ASN basic search links -> "rirSearch1" and "autnums"
- If response that includes DOMAIN relation search link -> "rirSearch1"


MiguelAHM:
Code scanning results / CodeQL is throwing an issue not related with this PR. It is scanning the class and complaining about a regex that exist already in master. I created a task in the backlog to look into it.
It is out of scope of this PR

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -75,6 +85,39 @@ public List<String> getInetnumRelationPkeys(final String pkey, final RelationTyp
                 .toList();
     }
 
+    public void mapRelationLinks(final RdapObject rdapResponse, final String requestUrl){
+        if (requestUrl == null || requestUrl.contains("rirSearch1")){
```
maggarwal13, line 89: change this to "rirSearch1".contains(requestUrl). Better to use enum RdapConformance

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -155,16 +159,19 @@ public Object mapSearch(final String requestUrl, final List<RpslObject> objects,
                 case DOMAIN -> {
                     final Domain domain = (Domain) getRdapObject(requestUrl, object, null);
                     mapRedactions(domain);
+                    rdapRelationService.mapRelationLinks(domain, requestUrl);
```
maggarwal13, line 162: If this needs to be  added for every case can it be outside switch ?

whois-api/src/test/java/net/ripe/db/whois/api/AbstractIntegrationTest.java
```diff
@@ -96,4 +104,23 @@ protected void removeLog4jAppender() {
...
+        return links.stream()
+                .filter(link -> {
+                    if (link.getRel() == null){
+                        return false;
+                    }
+                    try {
+                        RelationType.valueOf(link.getRel().toUpperCase());
+                        return true;
+                    } catch (IllegalArgumentException ex) {
+                        return link.getRel().equals("up-active") || link.getRel().equals("top-active");
```
maggarwal13, line 118: That may not be necessary? Also if you are expecting a relationType then up-active and top-active should also be part of enum ?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -489,7 +498,7 @@ private static Remark createRemark(final CIString key, final AbuseContact abuseC
     }
 
     private static boolean hasDescriptionsOrRemarks(final RpslObject rpslObject) {
-        return !rpslObject.getValuesForAttribute(AttributeType.DESCR).isEmpty() || !rpslObject.getValuesForAttribute(AttributeType.REMARKS).isEmpty();
+        return !rpslObject.getValuesForAttribute(DESCR).isEmpty() || !rpslObject.getValuesForAttribute(REMARKS).isEmpty();
```
maggarwal13, line 501: This takes multiple attributes so you can pass both descr and remarks in single call

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
MiguelAHM: good catch, but now I did refactor this.
This not longer applies

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
MiguelAHM: yes, I have done this already as part of my last review/refactor. Thanks

MiguelAHM: anyways we are already checking the Response type inside mapCommonRelationLinks so it should be safe.
I am refactoring it

MiguelAHM: I did refactor it, this can go inside getRdapObject

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
MiguelAHM: I remove that method, it is only called from one place and It is clear reading the code what is doing...it is no needed to create a new method for it

eshryane:
Tested rdap conformance OK using commit 3fe377c

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -342,7 +357,7 @@ private void mapEntityLinks(final Entity entity, final String requestUrl, final
                 throw new IllegalStateException("Malformed Url");
             }
             entity.getLinks().add(new Link(requestUrl, "self",
-                    url.getProtocol() + "://" + url.getHost() + "/" + entity.getObjectClassName() + "/" + attributeValue,
+                    url.getProtocol() + "://" + url.getHost() + PATH_DELIMITER + entity.getObjectClassName() + PATH_DELIMITER + attributeValue,
```
eshryane, line 360: I suggest either removing the `PATH_DELIMITER` constant (it doesn't help readability) and use String.format() instead OR (better I think) switch to `jakarta.ws.rs.core.UriBuilder` which is intended exactly for this purpose.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -471,11 +486,11 @@ private IpEntry lookupParentIpEntry(final IpInterval ipInterval) {
     private static Remark createRemark(final RpslObject rpslObject) {
         final List<String> descriptions = Lists.newArrayList();
 
-        for (final CIString description : rpslObject.getValuesForAttribute(AttributeType.DESCR)) {
+        for (final CIString description : rpslObject.getValuesForAttribute(DESCR)) {
```
eshryane, line 489: I suggest not removing the `AttributeType` as it makes the code less readable (i.e. I have no idea what type `DESCR` is by just reading the code).

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -133,13 +138,15 @@ public RdapObjectMapper(
...
-            @Value("${rdap.port43:}") final String port43) {
+            @Value("${rdap.port43:}") final String port43,
+            @Value("${rdap.public.baseUrl:}") final String baseUrl) {
         this.noticeFactory = noticeFactory;
         this.rpslObjectDao = rpslObjectDao;
         this.ipv4Tree = ipv4Tree;
         this.ipv6Tree = ipv6Tree;
         this.port43 = port43;
         this.reservedResources = reservedResources;
+        this.rdapRirSearchSkeleton = baseUrl + "/%s/rirSearch1/%s/%s";
```
eshryane, line 149: Define this structure in a static method using UriBuilder

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapService.java
```diff
@@ -237,7 +237,7 @@ public Response relationSearch(
             @PathParam("key") final String key,
             @QueryParam("status") String status) {
 
-        final RelationType relation = RelationType.fromString(relationType);
+        final RelationType relation = RelationType.fromName(relationType);
```
eshryane, line 240: I suggest renaming `fromName` to `fromValue` to be consistent with the naming in the `RelationType` class.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/RelationType.java
```diff
@@ -19,12 +21,21 @@ public String getValue() {
         return value;
     }
 
-    public static RelationType fromString(final String value) {
+    public static RelationType fromName(final String value) {
         for (RelationType type : RelationType.values()) {
             if (type.name().equalsIgnoreCase(value)) {
```
eshryane, line 26: Is `type.name()` going to match all values properly? For example, `RelationType.UP_ACTIVE.name()` is `UP_ACTIVE` not `up-active`. Do you mean to use `getValue()` instead?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
MiguelAHM: baseUrl come from the properties, it is not static.
I cannot make an static method using baseUrl

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/RelationType.java
MiguelAHM: this is why I called fromName.
"up-active" doesn't match because the call should be .../top/...?status=active

getValue() is used in isValidValue for Integration Tests.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapService.java
MiguelAHM: explained below, it is fromName because it is using the type.name.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
eshryane: OK. Then use a private method (not static) to generate the URL using UriBuilder (i.e. don't use String concatenation).

eshryane: Agreed to use UriBuilder

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/RelationType.java
eshryane: Agreed to use `getValue()` and match value not enum name.

eshryane (approved):

PR #1608, opened 2024-12-13
Add RDAP Relation extension

MiguelAHM:
Implementing section 3, "Relation Searches" of RIR Search RFC:
https://www.ietf.org/id/draft-ietf-regext-rdap-rir-search-13.html

- If response includes IP basic search links -> "rirSearch1" and "ips"
- If response includes IP basic search request -> "rirSearch1", "ips", and "ipSearchResults"
- If response includes ASN basic search links -> "rirSearch1" and "autnums"
- If response includes ASN basic search request -> "rirSearch1", "autnums", and "autnumSearchResults"
- If response that includes DOMAIN relation search link -> "rirSearch1"
- If response that includes DOMAIN relation search request -> "rirSearch1"
- /help -> "rirSearch1", "ips", "ipSearchResults", "autnums" and "autnumSearchResults"

For up or top we are returning the output as lookup json structure. However, It is a basic search request.
So for lookups when the call is a rirSearch we should add the compliance

This PR covers:
- If response includes IP basic search request -> "rirSearch1", "ips", and "ipSearchResults"
- If response includes ASN basic search request -> "rirSearch1", "autnums", and "autnumSearchResults"
- If response that includes DOMAIN relation search request -> "rirSearch1"
- /help -> "rirSearch1", "ips", "ipSearchResults", "autnums" and "autnumSearchResults"

Links with the respective extensions are added in the next PR:
https://github.com/RIPE-NCC/whois/pull/1606/files


whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -362,6 +367,10 @@ private RdapObject mapCommonConformances(final RdapObject rdapResponse) {
         return rdapResponse;
     }
 
+    private void mapRelationConformances(final RdapObject rdapResponse, final String requestUrl) {
```
eshryane, line 370: rename to `includeRirSearchConformance` as `RelationConformance` is too generic

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -0,0 +1,219 @@
...
+                .toList();
+    }
+
+    public List<String> getInetnumRelationPkeys(final String pkey, final RelationType relationType, final String status){
+        final IpInterval ip = IpInterval.parse(pkey);
+        final List<IpEntry> ipEntries = getIpEntries(getIpTree(ip), relationType, ip, status);
+        return ipEntries.stream().map(ipEntry -> ipEntry.getKey().toString()).toList();
+    }
+
+    public void mapRelationConformance(final RdapObject rdapObject, final String requestUrl){
```
eshryane, line 73: rename to `includeRirSearchConformance`

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/RelationType.java
```diff
@@ -0,0 +1,18 @@
...
+public enum RelationType {
+    UP("up"),
+    TOP("top"),
+    DOWN("down"),
+    BOTTOM("bottom");
+
+    private final String value;
+
+    RelationType(String value) {
+        this.value = value;
```
eshryane, line 12: How are invalid relation types handled?

eshryane: Add an integration test for an "invalid" relation type. Expect 400 Bad Request and "unknown relation" in the response.

whois-rpsl/src/main/java/net/ripe/db/whois/common/rpsl/attrs/InetnumStatus.java
```diff
@@ -107,6 +107,11 @@ public boolean worksWithParentInHierarchy(final InetStatus parentInHierarchyMain
         return true;
     }
 
+    @Override
+    public boolean isAdministrativeResource() {
+        return this.equals(ALLOCATED_UNSPECIFIED);
```
eshryane, line 112: This is not precise enough. This status may be used on out-of-region placeholders or the root 0/0 object.

According to https://www.iana.org/assignments/rdap-json-values/rdap-json-values.xhtml
```
The object instance has been allocated administratively (i.e., not for use by the recipient in their own right in operational networks).
```

We need to implement this separately.

whois-rpsl/src/main/java/net/ripe/db/whois/common/rpsl/attrs/Inet6numStatus.java
```diff
@@ -80,6 +80,11 @@ public boolean worksWithParentInHierarchy(final InetStatus parentInHierarchyMain
         return true;
     }
 
+    @Override
+    public boolean isAdministrativeResource() {
+        return this.equals(ALLOCATED_BY_RIR);
```
eshryane, line 85: This is not precise enough. This status may be used on out-of-region placeholders or the root `::/0` object.

According to https://www.iana.org/assignments/rdap-json-values/rdap-json-values.xhtml the "administrative" status means:
"The object instance has been allocated administratively (i.e., not for use by the recipient in their own right in operational networks)."
We need to implement this separately.

whois-commons/src/main/java/net/ripe/db/whois/common/etree/NestedIntervalMap.java
```diff
@@ -226,6 +232,17 @@ private List<InternalNode<K, V>> internalFindFirstMoreSpecific(K range) {
         return result;
     }
 
+    private List<InternalNode<K, V>> internalFindMostSpecific(K range){
```
eshryane, line 235: Add tests for this code

whois-rpsl/src/main/java/net/ripe/db/whois/common/rpsl/attrs/Inet6numStatus.java
MiguelAHM: I removed this, we are not handling status query parameter in this PR
So I don't need to check if it is an administrative Resource or not

whois-rpsl/src/main/java/net/ripe/db/whois/common/rpsl/attrs/InetnumStatus.java
MiguelAHM: I removed this, we are not handling status query parameter in this PR
So I don't need to check if it is an administrative Resource or not

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/RelationType.java
MiguelAHM: Thanks, I missed this test I will update https://github.com/RIPE-NCC/whois/pull/1603 with it

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
MiguelAHM: Done

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -183,6 +185,7 @@ public Object mapSearch(final String requestUrl, final List<RpslObject> objects,
 
         final RdapObject rdapObject = mapCommonNoticesAndPort(searchResult, requestUrl);
         mapCommonLinks(rdapObject, requestUrl);
+        includeRirSearchConformance(rdapObject, requestUrl);
         return mapCommonConformances(rdapObject);
```
maggarwal13, line 189: This conformance should be included if it is a RIR search ?

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapService.java
```diff
@@ -245,6 +250,35 @@ public Response help(@Context final HttpServletRequest request) {
                 .build();
     }
 
+    @GET
+    @Produces({MediaType.APPLICATION_JSON, CONTENT_TYPE_RDAP_JSON})
+    @Path("/{objectType}/rirSearch1/{relation}/{key:.*}")
+    public Response relationSearch(
```
maggarwal13, line 256: Can we create a different controller for rirsearch1 or these two paths are two entangled to seperate ? I believe there would be future changes or addition to it

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
MiguelAHM: The lira says 
"Always return the "rirSearch1" extension on all responses."

maggarwal13: If that is the case  we should add all conformance to every  rdap response but we do not. For ex: we add geofeed  Conformance only if it is Ip request . I think we need to be consistent about it.  Let's discuss it with Ed

MiguelAHM: I have been talking to Ed
"rirSearch1" will only appear in the /help
so I've removed it from "includeRirSearchConformance" the rest of the conformance should be there depending if the search is IP or Autnums

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -362,6 +367,10 @@ private RdapObject mapCommonConformances(final RdapObject rdapResponse) {
         return rdapResponse;
     }
 
+    private void includeRirSearchConformance(final RdapObject rdapResponse, final String requestUrl) {
```
maggarwal13, line 370: Why this is in rdapObjectMapper? This is specific to relation RDAP search so the way it should happen is rdaprelationService should have dependency on rdapObjectMapper and that RdapRelationService should call rdapObjectMapper get the object and then add RirSearchConformance , this way we are not mixing normal rdapObjectMapper code with relationObjectMapper code.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -66,6 +68,21 @@ public List<String> getDomainsByRelationType(final String pkey, final RelationTy
                 .toList();
     }
 
+
+    public void includeRirSearchConformance(final RdapObject rdapObject, final String requestUrl){
+        rdapObject.getRdapConformance().add(RdapConformance.RIR_SEARCH_1.getValue());
+        if (requestUrl == null) {
```
maggarwal13, line 74: use stringUtils.isEmpty

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -66,6 +68,21 @@ public List<String> getDomainsByRelationType(final String pkey, final RelationTy
...
+            return;
+        }
+        if (requestUrl.contains(RdapRequestType.IPS.name().toLowerCase())){
+            rdapObject.getRdapConformance().addAll(List.of(RdapConformance.IPS.getValue(), RdapConformance.IP_SEARCH_RESULTS.getValue()));
+            return;
```
maggarwal13, line 79: I do not think you can add IP_SEARCH_RESULTS and AUTNUM_SEARCH_RESULTS by default. It should be added only if it is an array.

maggarwal13: Discussed over huddle, section 6 points out this scenario

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
MiguelAHM: we are only doing for basic search as the RFC points
so it will not be any conformance for UP or TOP because they are lookup

MiguelAHM: updated the description and commented via chat

MiguelAHM:
merged as part of 
https://github.com/RIPE-NCC/whois/pull/1606
adding the links into the conformance

PR #1611, opened 2024-12-17
Fixes to RDAP RIR-Search (Basic Searches) Feature

MiguelAHM:
This PR contains some fixes for the RDAP RIR-Search (Basic Searches) feature, including:

* The `ips?handle` search input should match against the handle field and not the name field of the object.
  * The handle in IPs should match the handle in the RDAP output.
* Do not add an implicit trailing `*` (i.e. match all) to the `ips?handle` search.
  * Perform an exact match search in IP searches.
  * The user can add a trailing `*` themselves if they wish.
* Do not add an implicit trailing `*` (i.e. match all) to the `autnums?handle` search.
  * Perform an exact match search in autnum searches.
* Return an empty value in the search response if no match is found. Do not return a 404.
* Make searches case insensitive.


whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapElasticFullTextSearchService.java
```diff
@@ -90,19 +91,34 @@ protected List<RpslObject> doSearch() throws IOException {
...
+                }
+
+                private boolean hasNotWildCard(){
+                    return term.indexOf('*') == -1 && term.indexOf('?') == -1;
+                }
+
+                private BoolQueryBuilder createExactMatchQuery(){
+                    final BoolQueryBuilder rawSearchBuilder = QueryBuilders.boolQuery();
+                    for (String field : fields) {
+                        rawSearchBuilder.should(QueryBuilders.termQuery(String.format("%s.raw", field), term));
```
MiguelAHM, line 111: termQuery: 
looks for the exact term 
It does not analyze the input term or the field's value. Instead, it checks for an exact match between the provided value (term) and the value stored in the field.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapElasticFullTextSearchService.java
```diff
@@ -90,19 +91,34 @@ protected List<RpslObject> doSearch() throws IOException {
...
                 }
 
-                private QueryBuilder getQueryBuilder(final String[] fields, final String term) {
-                    if (term.indexOf('*') == -1 && term.indexOf('?') == -1) {
-                        final MultiMatchQueryBuilder multiMatchQuery = new MultiMatchQueryBuilder(term, fields)
-                                .type(MultiMatchQueryBuilder.Type.PHRASE_PREFIX)
-                                .operator(Operator.AND);
-                        return multiMatchQuery;
+                private QueryBuilder getQueryBuilder(final String[] fields, final String term, final boolean matchExact) {
+                    if (hasNotWildCard()) {
```
MiguelAHM, line 95: invert it

eshryane:
FYI I tested wildcard search with a term that will match a lot of objects (i.e. every inet6num primary key) and the response was returned quickly and  limited correctly to 100 results.
```
$ curl -o file -v "http://localhost:1080/rdap/ips?handle=2001*"
$ fgrep '"handle"' file | grep "2001" | wc -l
100
```

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapElasticFullTextSearchService.java
MiguelAHM: This is done

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapElasticFullTextSearchService.java
```diff
@@ -90,13 +91,30 @@ protected List<RpslObject> doSearch() throws IOException {
                     return results;
                 }
 
-                private QueryBuilder getQueryBuilder(final String[] fields, final String term) {
-                    if (term.indexOf('*') == -1 && term.indexOf('?') == -1) {
-                        return new MultiMatchQueryBuilder(term, fields)
-                                .type(MultiMatchQueryBuilder.Type.PHRASE_PREFIX)
-                                .operator(Operator.AND);
+                private QueryBuilder getQueryBuilder(final String[] fields, final String term, final boolean matchExact) {
+                    if (hasWildCard()) {
```
maggarwal13, line 95: MatcExact  and hasWildcard seems to be contradictory. if  hasWildCard() then   createWildCardQuery() else exact match?

MiguelAHM: I am already doing this, right?

                private QueryBuilder getQueryBuilder(final String[] fields, final String term, final boolean matchExact) {
                    if (hasWildCard()) {
                        return createWildCardQuery();
                    }

                    return matchExact ? createExactMatchQuery() :
                            new MultiMatchQueryBuilder(term, fields)
                                    .type(MultiMatchQueryBuilder.Type.PHRASE_PREFIX)
                                    .operator(Operator.AND);
                }

maggarwal13: Yes , That's what I am asking, why you need matchExact boolean value ? If it has wildcard , it goes to createWildCardQuery else it should be always createExactMatchQuery(). or it is for some other functionality ?

maggarwal13: I think if we say user needs to put wildcard, we should be consistent and get rid of matchQuery flag all together. Does RFC points that only for ips and autnums we need to do exact match  and rest can match multiple ?

MiguelAHM: I think that the RFC only talks about ips and autnums.
Maybe discuss with Ed, I think that we discussed about it and the outcome was just to modify ips and autnums

maggarwal13: In that case, is it possible to check the fields here to return exactMatch or MultiMatch rather passing the boolean flag from controller level? That would make it more clearer

maggarwal13 (approved):

PR #1616, opened 2024-12-30
Fix Regular Expression to Match Sonar

MiguelAHM:
- Old Regex:
old_regex

- New Regex:
new_regex

Differences with new and old regex:

Test String: 12-34-56
  Old Regex: false
  New Regex: true

Test String: 0-034
  Old Regex: true
  New Regex: false

Test String: 12-34-56-78
  Old Regex: false
  New Regex:: true

However, these situations can never happen because this code is already refusing this use cases

```
       if (cleanAddress.contains("-")) {
            Validate.isTrue(reverseParts.size() == 4 && reverseParts.get(0).contains("-"), "Dash notation not in 4th octet: ", address);
            Validate.isTrue(cleanAddress.indexOf('-') == cleanAddress.lastIndexOf('-'), "Only one dash allowed: ", address);
            hasDash = true;
        }
```

This code (which is throwing an early error) is taking care of situations like x-x-x or x-x.x-x. etc
this only allows like xx-xx.xx.xx.xx for example and in those situations both regex do the same


MiguelAHM:
TEST OK

MiguelAHM:
we will just ignore it in sonar

PR #1617, opened 2024-12-31
Fix Flaky Client Certificate Authentication Test

MiguelAHM:
This PR attempts to fix the flaky (flapping) integration test:
```
[ERROR] Failures: 
[ERROR]   ClientCertificateServiceTestIntegration.fail_on_no_client_certificate:38 
Expected: a string containing "javax.net.ssl.SSLHandshakeException: Received fatal alert: bad_certificate"
     but: was "java.io.IOException: Error writing to server"
```

It seems that the retryFor is not enough. Most of the times it works well after performing an OK request.
There are some time that even performing an OK request is not enough, for those situation I think that we should keep the retryFor. Following this, the test are OK

I think that the root issue is with the setup of the IT for Client Certificate. This requires a deeper look at the integration test set up

### DIFF
--- a/whois-api/src/test/java/net/ripe/db/whois/api/httpserver/ClientCertificateServiceTestIntegration.java
+++ b/whois-api/src/test/java/net/ripe/db/whois/api/httpserver/ClientCertificateServiceTestIntegration.java
@@ -6,6 +6,7 @@ import net.ripe.db.whois.api.SecureRestTest;
 import net.ripe.db.whois.common.aspects.RetryFor;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
+import org.springframework.test.annotation.DirtiesContext;
 
 import java.io.IOException;
 
@@ -14,6 +15,7 @@ import static org.hamcrest.Matchers.containsString;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
 @Tag("IntegrationTest")
+@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_EACH_TEST_METHOD)
 public class ClientCertificateServiceTestIntegration extends AbstractClientCertificateIntegrationTest {
 
     @Test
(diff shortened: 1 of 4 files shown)


whois-api/src/test/java/net/ripe/db/whois/api/httpserver/ClientCertificateServiceTestIntegration.java
```diff
@@ -3,40 +3,44 @@
...
-        final String response = SecureRestTest.target(getClientSSLContext(), getClientCertificatePort(), "whois/client")
-            .request()
-            .get(String.class);
-
+        final String response = getClientCertificateWithCertificate();
         assertThat(response, containsString("Found 1 certificate(s)."));
         assertThat(response, containsString("-----BEGIN CERTIFICATE-----"));
         assertThat(response, containsString("-----END CERTIFICATE-----"));
+
+        final ProcessingException processingException = assertThrows(ProcessingException.class,
```
eshryane, line 29: Why are you expecting a `ProcessingException` ? The test should succeed

eshryane: Should the test retry on `ProcessingException` ? For example
```
@RetryFor(value = {IOException.class, ProcessingException.class}, attempts = 10, intervalMs = 10000)
```

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceClientCertificateTestIntegration.java
```diff
@@ -223,6 +225,8 @@ public void update_person_missing_private_key_unauthorised() throws Exception {
              "mnt-by: OWNER-MNT\n" +
              "source: TEST");
 
+        sendUpdateRequest(updatedPerson);
```
eshryane, line 228: Why was an extra update added to the test?

eshryane:
Alternative : #1625

whois-api/src/test/java/net/ripe/db/whois/api/httpserver/ClientCertificateServiceTestIntegration.java
MiguelAHM: This is no fixing 100% the test. It is a draft, please do not review it

PR #1631, opened 2025-01-15
Update key-cert lookup URL in update notification

pkern:
The old lookup.html page no longer works and there is currently no redirect to the new query interface, so update the links in the notification to point to the new one.


eshryane:
@pkern Thanks very much for your contribution!

PR #1633, opened 2025-01-21
Remove RDAP business logic from controller into a server layer

MiguelAHM:
- Move all logic of RDAP to a service layer
- Keep the redirects in controller
- Just turn the output into response in controller


maggarwal13 (approved):

MiguelAHM:
test OK

PR #1634, opened 2025-01-21
Single-Result Searches should Lookup

MiguelAHM:
https://datatracker.ietf.org/doc/draft-ietf-regext-rdap-rir-search/

4.1.  Single-Result Searches

   The "up" and "top" relations are single-result searches.  When
   processing these searches, if there is a result for the search, the
   server returns that object as though it were requested directly via a
   lookup URL (see section 3.1 of [RFC9082]).  If there is no result for
   the search, the server returns an HTTP 404 (Not Found) [RFC9110]
   response code.

### DIFF
--- a/whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
@@ -33,9 +33,12 @@ import javax.annotation.Nullable;
 import java.util.Collection;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Stream;
 
 import static net.ripe.db.whois.api.rdap.RdapService.COMMA_JOINER;
 import static net.ripe.db.whois.common.rpsl.ObjectType.DOMAIN;
+import static net.ripe.db.whois.common.rpsl.ObjectType.INET6NUM;
+import static net.ripe.db.whois.common.rpsl.ObjectType.INETNUM;
 import static net.ripe.db.whois.common.rpsl.attrs.Inet6numStatus.ALLOCATED_BY_RIR;
 import static net.ripe.db.whois.common.rpsl.attrs.InetnumStatus.ALLOCATED_UNSPECIFIED;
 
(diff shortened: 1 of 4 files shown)


whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapServiceTestIntegration.java
```diff
@@ -3336,26 +3324,25 @@ public void get_ipv6_top_not_found(){
     public void get_ipv6_top_found_if_assignment(){
         loadIpv6RelationTreeExample();
 
-        databaseHelper.updateObject("" +
-                "inet6num:       2000::/3\n" +
+        databaseHelper.addObject("" +
```
MiguelAHM, line 3327: I checked

rdapLookupService.lookupObject

and I see that we are validating RdapObjectMapper.isIANABlock(resultObject)
We never return ::/0 or 0/0 in lookups 
so I created a parent for 2000::/3 to avoid to hit ::/0

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -62,54 +62,77 @@ public RdapRelationService(final Ipv4Tree ip4Tree,
...
             case AUTNUMS -> throw new RdapException("501 Not Implemented", "Relation queries not allowed for autnum", HttpStatus.NOT_IMPLEMENTED_501);
             case DOMAINS -> {
                 rdapRequestValidator.validateDomain(key);
-                final List<String> relatedPkeys = getDomainsByRelationType(key, relationType);
+                final List<IpEntry> domainEntries = getDomainsEntriesByRelationType(key, relationType);
+
+                if (shouldReturnLookup){
+                    final IpEntry ipEntry = domainEntries.getFirst();
+                    final RpslObject domainObject = rpslObjectDao.getById(ipEntry.getObjectId());
+                    return rdapLookupService.lookupForDomain(request, domainObject.getKey().toString(), ipEntry.getKey().toString());
```
maggarwal13, line 95: LookupForDomain is again hitting  database to fetch domain Object ? Can you avoid multiple calls ? May eb overload the function?

maggarwal13 (approved):

MiguelAHM:
test OK
https://gitlab.ripe.net/swe-database-team/whois/-/jobs/2224786

PR #1638, opened 2025-01-29
Return empty array in RDAP search response when no match is found

MiguelAHM:
The [RDAP RIR Search](https://datatracker.ietf.org/doc/draft-ietf-regext-rdap-rir-search/) specification for Multiple-Result searches states:
```
If the search can be processed by the server, but there are no
results for the search, then the server returns an HTTP 200 (OK)
[RFC9110] response code, with the body of the response containing an
empty results array.
``` 

Our interpretation of the specification is as follows : 
* For /ips/rirSearch1/ searches, return an empty "ipSearchResults" array if there are no results
* For /autnums/rirSearch1/ searches, return an empty "autnumSearchResults" if there are no results.
* For /domains/rirSearch1/ searches, return an empty "domainSearchResults" if there are no results

For consistency, we extend this convention to the existing [JSON Responses](https://datatracker.ietf.org/doc/html/rfc9083) endpoints:
* For /domains searches, return an empty "domainSearchResults" if there are no results
* For /entities searches, return an empty "entitySearchResults" if there are no results
* For /nameservers searches, 501 Not Implemented is always returned

If there *are* results then the matching object(s) are returned in the specific array.

### DIFF
--- a/whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
@@ -202,7 +202,7 @@ public class RdapObjectMapper {
             case DOMAINS -> searchResult.setDomainResults(Lists.newArrayList());
             case IPS -> searchResult.setIpResults(Lists.newArrayList());
             case AUTNUMS -> searchResult.setAutnumResults(Lists.newArrayList());
-            default -> searchResult.setEntityResults(Lists.newArrayList());
+            case ENTITY -> searchResult.setEntityResults(Lists.newArrayList());
         }
         return searchResult;
     }
--- a/whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
@@ -155,10 +155,9 @@ public class RdapObjectMapper {
         return mapCommons(getRdapObject(requestUrl, rpslObject, abuseContact), requestUrl);
     }
 
-    public Object mapSearch(final String requestUrl, final List<RpslObject> objects, final int maxResultSize) {
-        final SearchResult searchResult = objects.isEmpty() ?
-                new SearchResult().initialiseEmpty() :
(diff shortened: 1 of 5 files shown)


whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -156,7 +156,10 @@ public Object map(final String requestUrl,
     }
 
     public Object mapSearch(final String requestUrl, final List<RpslObject> objects, final int maxResultSize) {
-        final SearchResult searchResult = new SearchResult();
+        final SearchResult searchResult = objects.isEmpty() ?
```
eshryane, line 159: Will this change return all 4 arrays if *any* search result is populated? We have to be careful only to return an empty array if the user has searched for the related type.

MiguelAHM: I misunderstood the requirement. Last push should do it correctly

eshryane: I updated the PR description to match our interpretation of the RFC.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapObjectMapper.java
```diff
@@ -194,6 +196,17 @@ public Object mapSearch(final String requestUrl, final List<RpslObject> objects,
         return mapCommonConformances(rdapObject);
     }
 
+    private SearchResult initialiseSearchResultByType(final RdapRequestType requestType){
+        final SearchResult searchResult = new SearchResult();
+        switch (requestType){
+            case DOMAINS -> searchResult.setDomainResults(Lists.newArrayList());
+            case IPS -> searchResult.setIpResults(Lists.newArrayList());
+            case AUTNUMS -> searchResult.setAutnumResults(Lists.newArrayList());
+            default -> searchResult.setEntityResults(Lists.newArrayList());
```
eshryane, line 205: Can you set entity results only for ENTITY requestType ?

eshryane (approved):

PR #1644, opened 2025-01-31
Use environment property and Refactor Dummifier JMX

MiguelAHM:
- Refactor DatabaseDummifierJmx to use the properties from the service.
- - Check from were are we calling this JMX -> It is not called. We are relying on a oneshot script for dummification
- - Tested on PREPDEV
- Use whois.environment property
- - TODO: Modify the properties files to use whois.environment instead of sentry.environment (once merged)


whois-commons/src/main/java/net/ripe/db/whois/common/DatabaseDummifierJmx.java
```diff
@@ -43,33 +42,34 @@
...
+    private final JdbcTemplate jdbcTemplate;
 
     private static final DummifierRC dummifier = new DummifierRC();
-
     private static final AtomicInteger jobsAdded = new AtomicInteger();
     private static final AtomicInteger jobsDone = new AtomicInteger();
 
-    public DatabaseDummifierJmx() {
+    @Autowired
+    public DatabaseDummifierJmx(@Value("${whois.environment:}") final String environment,
```
eshryane, line 63: Should the environment be mandatory? Or default to e.g. localhost ?

whois-commons/src/main/java/net/ripe/db/whois/common/DatabaseDummifierJmx.java
```diff
@@ -43,33 +42,34 @@
...
+                                @Qualifier("whoisMasterDataSource") final DataSource writeDataSource) {
         super(LOGGER);
+        this.environment = Environment.valueOf(environment.toUpperCase());
```
eshryane, line 66: Catch `IllegalArgumentException` in case of an invalid value? Do we want to list valid values?

whois-commons/src/test/java/net/ripe/db/whois/common/dao/jdbc/AbstractDatabaseHelperIntegrationTest.java
```diff
@@ -51,6 +51,7 @@ public static synchronized void setupAbstractDatabaseHelperTest() throws Excepti
         System.setProperty("grs.sources", "TEST-GRS");
         System.setProperty("grs.sources.dummify", "TEST-GRS");
         System.setProperty("api.rest.baseurl", "http://rest-test.db.ripe.net");
+        System.setProperty("whois.environment", "dev");
```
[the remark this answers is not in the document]
MiguelAHM: I think that for testing purposes we shouldn't modify the Environment enum to add a local environment. 
we can just use dev.

whois-commons/src/main/java/net/ripe/db/whois/common/DatabaseDummifierJmx.java
MiguelAHM: I think that it should be mandatory, rather than adding a localhost

PR #1649, opened 2025-02-03
API KEYs must be valid just within the environment

MiguelAHM:
i.e.: In case an API KEY belong to training environment it shouldn't be valid for PROD environment


whois-api/src/test/java/net/ripe/db/whois/api/ApiKeyAuthServerDummy.java
```diff
@@ -57,6 +60,9 @@ public class ApiKeyAuthServerDummy implements Stub {
         APIKEY_TO_OAUTHSESSION.put(BASIC_AUTH_PERSON_OWNER_MNT, new OAuthSession(AUD, "l6lRZgvOFIphjiGwtCGuLwqw","person@net.net", "906635c2-0405-429a-800b-0602bd716124", "profile email whois.mntner:OWNER-MNT"));
         APIKEY_TO_OAUTHSESSION.put(BASIC_AUTH_TEST_TEST_MNT, new OAuthSession(AUD, "hHZjAbXPtxGxUJCgdwv2ufhY","test@ripe.net", "8ffe29be-89ef-41c8-ba7f-0e1553a623e5", "whois.mntner:TEST-MNT profile email"));
         APIKEY_TO_OAUTHSESSION.put(BASIC_AUTH_INVALID_SIGNATURE_API_KEY, new OAuthSession(AUD, "hHZjAbXPtxGxUJCgdwv2ufhY","invalid@ripe.net", "8ffe29be-89ef-41c8-ba7f-0e1553a623e5", "profile email whois.mntner:TEST-MNT"));
+        APIKEY_TO_OAUTHSESSION.put(BASIC_AUTH_FULL_ENV_SIGNATURE_API_KEY, new OAuthSession(AUD, "3TE0KYX35SZQUKIABM5VAOGW","test@ripe.net", "906635c2-0405-429a-800b-0602bd716124", "whois.environment:LOCALHOST.mntner:OWNER-MNT.write"));
```
maggarwal13, line 63: In this way in future when we support multiple scope it has to be in every scope . I think correct way to implement is having a second scope just for environment and keep the existing scope structure same.

whois-api/src/test/java/net/ripe/db/whois/api/ApiKeyAuthServerDummy.java
```diff
@@ -57,6 +60,9 @@ public class ApiKeyAuthServerDummy implements Stub {
...
+        APIKEY_TO_OAUTHSESSION.put(BASIC_AUTH_NOT_MNT_ENV_SIGNATURE_API_KEY, new OAuthSession(AUD, "SJ9691NXUKZFVZMQXLQYBVRH","test@ripe.net", "906635c2-0405-429a-800b-0602bd716124", "whois.environment:LOCALHOST.write"));
+        APIKEY_TO_OAUTHSESSION.put(BASIC_AUTH_FULL_WRONG_ENV_SIGNATURE_API_KEY, new OAuthSession(AUD, "RQMO5ASNRUN8ZJNP6TCS1Q2Q","test@ripe.net", "906635c2-0405-429a-800b-0602bd716124", "whois.environment:PREPDEV.write"));
```
maggarwal13, line 65: Please add the profile and email scope as well in the list.

whois-commons/src/main/java/net/ripe/db/whois/common/apiKey/ApiKeyUtils.java
```diff
@@ -35,7 +37,11 @@ public static boolean validateScope(final OAuthSession oAuthSession, final List<
 
         final OAuthSession.ScopeFormatter scopeFormatter = new OAuthSession.ScopeFormatter(whoisScope.get());
 
-        if(StringUtils.isEmpty(scopeFormatter.getScopeKey()) || StringUtils.isEmpty(scopeFormatter.getScopeType()) || StringUtils.isEmpty(scopeFormatter.getAppName())) {
+        if (!StringUtil.isNullOrEmpty(scopeFormatter.getScopeEnv()) && !environment.equalsIgnoreCase(scopeFormatter.getScopeEnv())){
```
maggarwal13, line 40: No need to check for null or empty

whois-commons/src/main/java/net/ripe/db/whois/common/apiKey/ApiKeyUtils.java
```diff
@@ -44,21 +50,22 @@ public static boolean validateScope(final OAuthSession oAuthSession, final List<
...
 
+
     private static Optional<String> getWhoisScope(OAuthSession oAuthSession) {
         final List<String> scopes = Arrays.asList(StringUtils.split(oAuthSession.getScope(), " "));
         return scopes.stream().filter(scope -> scope.startsWith("whois")).findFirst();
     }
 
-    public static boolean hasValidApiKey(final OAuthSession oAuthSession, final List<RpslObject> maintainers, final List<RpslAttribute> authAttributes) {
-        if(oAuthSession == null || oAuthSession.getUuid() == null) {
+    public static boolean hasValidApiKey(final OAuthSession oAuthSession, final List<RpslObject> maintainers, final List<RpslAttribute> authAttributes, final String environment) {
```
maggarwal13, line 59: If you are adding in the scope validation should be at scope level in method validateScope not here

whois-commons/src/main/java/net/ripe/db/whois/common/apiKey/ApiKeyUtils.java
```diff
@@ -84,7 +91,7 @@ public static boolean isAPIKeyRequest(final String authHeader) {
     }
 
     public static String getApiKeyId(final String authHeader) {
-        if(authHeader == null || !authHeader.startsWith("Basic ")) {
+        if (authHeader == null || !authHeader.startsWith("Basic ")) {
```
maggarwal13, line 94: May be "Basic ".startsWith(authHeader) , it will do null check as well

whois-commons/src/main/java/net/ripe/db/whois/common/apiKey/OAuthSession.java
```diff
@@ -87,19 +87,33 @@ public static class ScopeFormatter {
...
+        final String scopeEnv;
         final String scopeKey;
 
         public ScopeFormatter(final String scope) {
-            final String[] parts = scope.split(":|\\.");
+            final String[] parts = scope.split("[:.]");
 
-            if(parts.length == 0 || parts.length < 2) {
+            if (parts.length <= 4){
+                this.appName = parts[0];
```
maggarwal13, line 97: I think we should use the seperate environment scope rather than having in environemnt in middle of one scope. This makes it very clear which scope is for environment and which scope is for mntnr

whois-query/src/test/java/net/ripe/db/whois/query/planner/RpslResponseDecoratorTest.java
```diff
@@ -85,6 +86,7 @@ public void setup() {
                 ssoTokenTranslator,
                 authServiceClient,
                 clientAuthCertificateValidator,
+                Environment.DEV.name(),
```
maggarwal13, line 89: use TEST for tests?

whois-commons/src/main/java/net/ripe/db/whois/common/apiKey/OAuthSession.java
MiguelAHM: yes, done

whois-api/src/test/java/net/ripe/db/whois/api/ApiKeyAuthServerDummy.java
MiguelAHM: yes, done. I misunderstood the task

whois-api/src/test/java/net/ripe/db/whois/api/ApiKeyAuthServerDummy.java
MiguelAHM: Done

whois-commons/src/main/java/net/ripe/db/whois/common/apiKey/ApiKeyUtils.java
MiguelAHM: This doesn't apply now, the code has changed

whois-commons/src/main/java/net/ripe/db/whois/common/apiKey/ApiKeyUtils.java
MiguelAHM: Done

whois-query/src/main/java/net/ripe/db/whois/query/planner/RpslResponseDecorator.java
```diff
@@ -168,7 +172,7 @@ private Iterable<? extends ResponseObject> filterAuth(Query query, final Iterabl
                 (CollectionUtils.isEmpty(passwords) && StringUtils.isBlank(ssoToken) && hasNotCertificates(certificates) && oAuthSession == null)?
                         FILTER_AUTH_FUNCTION :
                         new FilterAuthFunction(passwords, oAuthSession, ssoToken, ssoTokenTranslator, authServiceClient,
-                                rpslObjectDao, certificates, clientAuthCertificateValidator);
+                                rpslObjectDao, certificates, clientAuthCertificateValidator, environment);
```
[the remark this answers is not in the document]
MiguelAHM: for getting the environment value we need to use one component.
ApiKeyUtils is an static class and if turn that into a component we still need to inject that into the RpslResponseDecorator.

RpslResponseDecorator (component) -> FilterAuthFunction (non a component) -> ApiKeyUtils (non a component)

another option would be to modify Query object to include the environment, for that we need to add the environment in SearchQueryExecutor...which makes no sense for me

whois-api/src/test/java/net/ripe/db/whois/api/ApiKeyAuthServerDummy.java
```diff
@@ -48,15 +48,23 @@ public class ApiKeyAuthServerDummy implements Stub {
...
 
     public static final Map<String, OAuthSession> APIKEY_TO_OAUTHSESSION =  Maps.newHashMap();
 
     {
-        APIKEY_TO_OAUTHSESSION.put(BASIC_AUTH_TEST_NO_MNT, new OAuthSession(AUD, "hHZjAbXPtxGxUJCgdwv2ufhY", "test@ripe.net", "8ffe29be-89ef-41c8-ba7f-0e1553a623e5", "profile email"));
+        APIKEY_TO_OAUTHSESSION.put(BASIC_AUTH_TEST_NO_MNT, new OAuthSession(AUD, "hHZjAbXPtxGxUJCgdwv2ufhY", "test@ripe.net", "8ffe29be-89ef-41c8-ba7f-0e1553a623e5", "profile email whois.environment:LOCALHOST.write"));
         APIKEY_TO_OAUTHSESSION.put(BASIC_AUTH_PERSON_NO_MNT, new OAuthSession(AUD, "l6lRZgvOFIphjiGwtCGuLwqw","person@net.net", "906635c2-0405-429a-800b-0602bd716124", null));
-        APIKEY_TO_OAUTHSESSION.put(BASIC_AUTH_PERSON_OWNER_MNT, new OAuthSession(AUD, "l6lRZgvOFIphjiGwtCGuLwqw","person@net.net", "906635c2-0405-429a-800b-0602bd716124", "profile email whois.mntner:OWNER-MNT"));
+        APIKEY_TO_OAUTHSESSION.put(BASIC_AUTH_PERSON_OWNER_MNT, new OAuthSession(AUD, "l6lRZgvOFIphjiGwtCGuLwqw", "person@net.net", "906635c2-0405-429a-800b-0602bd716124", "profile email whois.mntner:OWNER-MNT"));
```
maggarwal13, line 61: Shouldn't environment scope be mandatory or user already had existing keys without environment and this one tests that?

whois-commons/src/main/java/net/ripe/db/whois/common/apiKey/OAuthSession.java
```diff
@@ -79,27 +79,36 @@ public String toString() {
...
-                .add("scopes", scope)
+                .add("scope", scope)
                 .toString();
     }
 
    public static class ScopeFormatter {
 
         final String appName;
         final String scopeType;
+        final String scopeEnv;
```
maggarwal13, line 90: This is not needed right ? Can you rename this to whoisScopeFormatter and for environment check  just check hardcoded value as you don't need it anywhere else

whois-api/src/test/java/net/ripe/db/whois/api/ApiKeyAuthServerDummy.java
MiguelAHM: At this point we have keys without environment yes. We are just creating keys with environment since we deploy these changes. This is testing the situation when one used has an existing key without env yes.

whois-commons/src/main/java/net/ripe/db/whois/common/apiKey/OAuthSession.java
MiguelAHM: I need this to extract the environment from the key that the user is passing. I don't understand what you mean here.
The key could have been created in RC environment while this is deployed in PROD environment. We need to discard that key because we are not in the correct environment (there is a IT for that already)

MiguelAHM: In addition, I am using this parser from whois-internal as well. So I need it there as well, and the logic here is the same for whois/whois-internal so we should keep it in one place for maintainability (we just need to change the parser in one place)

maggarwal13: In that case, can you use same scopeKey as envName and scopeType as ENVIRONMENT? You can also make it scopeType as enum with MNTNR and ENVIRONMENT instead of adding scopeEnv as new variable

maggarwal13: This will make it easy to extend multiple scope in future with scopeType for ex: for IP address level scope scopeType will be IP and scopeKey as ip address

MiguelAHM: yes, I was thinking about that already. I think that is a good idea

MiguelAHM: This is done now. Thanks for suggesting this change

whois-api/src/main/java/net/ripe/db/whois/api/apiKey/BearerTokenExtractor.java
```diff
@@ -46,25 +49,25 @@ public OAuthSession extractBearerToken(final HttpServletRequest request, final S
...
 
             if(!verifyJWTSignature(signedJWT)) {
               LOGGER.debug("JWT signature verification failed for {}", apiKeyId);
-              return new OAuthSession(apiKeyId);
+              return new OAuthSession(apiKeyId, this.environment);
             }
             
             //TODO[MA]: remove when apiKeyId is available from api registry call
-            return OAuthSession.from(new ObjectMapper().readValue(signedJWT.getPayload().toString(), OAuthSession.class), apiKeyId);
+            return OAuthSession.from(new ObjectMapper().readValue(signedJWT.getPayload().toString(), OAuthSession.class), apiKeyId, this.environment);
```
MiguelAHM, line 64: @maggarwal13 I am unsure if this is OK
My way of thinking is that we are now identifying a session by its environment as well. 
keycloak is using same env for our PROD, Training, TEST, RC so for them the session is the same for all of those environment...but not for us.
So for me make sense to add environment into the OAuthSession for identifying our env from the session that Keycloak is providing. But maybe it is overthinking and no clear?

whois-query/src/main/java/net/ripe/db/whois/query/planner/RpslResponseDecorator.java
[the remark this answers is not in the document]
MiguelAHM: found a solution. Adding env into OauthSession...I put an explanation in another comment

whois-api/src/main/java/net/ripe/db/whois/api/apiKey/BearerTokenExtractor.java
```diff
@@ -46,25 +49,25 @@ public OAuthSession extractBearerToken(final HttpServletRequest request, final S
...
+            return new OAuthSession(apiKeyId, this.environment);
         }
 
         try {
             final SignedJWT signedJWT = SignedJWT.parse(StringUtils.substringAfter(bearerToken, "Bearer "));
 
             if(!verifyJWTSignature(signedJWT)) {
               LOGGER.debug("JWT signature verification failed for {}", apiKeyId);
-              return new OAuthSession(apiKeyId);
+              return new OAuthSession(apiKeyId, this.environment);
```
maggarwal13, line 60: If API key is invalid no need to add environment. Also make sure the audit logs also have this variable.

whois-commons/src/main/java/net/ripe/db/whois/common/apiKey/ApiKeyUtils.java
```diff
@@ -23,42 +24,55 @@ public class ApiKeyUtils {
...
+    private static boolean isValidEnvironment(final Environment environment, final List<OAuthSession.ScopeFormatter> whoisScope) {
+        final Optional<OAuthSession.ScopeFormatter> environmentScope = whoisScope.stream()
+                .filter(scopeFormatter -> OAuthSession.ScopeType.ENVIRONMENT.equals(scopeFormatter.getScopeType()))
+                .findFirst();
+
+        return environmentScope.map(scopeFormatter -> environment.name().equals(scopeFormatter.getScopeKey())).orElse(true);
+    }
+
+
+    private static List<String> getWhoisScope(final OAuthSession oAuthSession) {
```
maggarwal13, line 61: Can you change it to getWhoisMntnrScope ? and create new method to getWhoisEnvScope? Or use stream and groupBy scopeType?

whois-commons/src/main/java/net/ripe/db/whois/common/apiKey/OAuthSession.java
```diff
@@ -72,38 +79,42 @@ public String getScope() {
...
+
     @Override
     public String toString() {
         return MoreObjects.toStringHelper(this)
                 .add("aud", aud)
                 .add("keyId", keyId)
                 .add("email", email)
                 .add("uuid", uuid)
-                .add("scopes", scope)
+                .add("scope", scope)
```
maggarwal13, line 93: can you add env also in toString?

whois-commons/src/main/java/net/ripe/db/whois/common/apiKey/ApiKeyUtils.java
MiguelAHM: no, this method is for getting all the scopes starting by whois.
This is ment for removing unwanted scopes. The name is also OK

whois-api/src/main/java/net/ripe/db/whois/api/apiKey/BearerTokenExtractor.java
MiguelAHM: Done

MiguelAHM:
This kind of scopes will not be supported by renovate
They will provide a different clients per PROD environment...so just property changes will be required

PR #1653, opened 2025-02-06
Use PEM format NRTMv4 client

maggarwal13:


MiguelAHM (approved):

whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/processor/UpdateNotificationFileProcessor.java
```diff
@@ -175,11 +176,11 @@ private UpdateNotificationFileResponse getUpdateNotificationFileResponse(final J
 
     private boolean isCorrectSignature(final JWSObject jwsObjectParsed) {
         try {
-            final ECKey parsedPublicKey =  ECKey.parse(readPublicKey());
+            final ECKey parsedPublicKey = (ECKey) JWK.parseFromPEMEncodedObjects(readPublicKey());
```
MiguelAHM, line 179: readPublicKey() 
this is reading the public key which is in public.key.
This file is in src.main.java.resources as well, so in a NON local environment it will hit the public.key of src

maggarwal13: I will delete thefile from main resources package public key. We should not commit env specific files in codebase. It will be very diffiult to enable it in other environment as we need code changes for that.

MiguelAHM: fine, just take into consideration that test will start breaking because they expect the .key in the current path

maggarwal13: I changed it in pem_format for now and we will fix this in separate task.

MiguelAHM (approved):

PR #1655, opened 2025-02-07
Do not show unexpected highlighted keys in FullTextSearch

MiguelAHM:
Reported that in FulltextSearch lowercase highlights are showing up. 
We should avoid that


MiguelAHM:
test OK

PR #1656, opened 2025-02-11
Use property for whois keycloak Id

maggarwal13:
Create property for whois keycloakId instead of hardcoding to whois

### DIFF
--- a/whois-api/src/main/java/net/ripe/db/whois/api/apiKey/BearerTokenExtractor.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/apiKey/BearerTokenExtractor.java
@@ -48,6 +48,12 @@ public class BearerTokenExtractor   {
         return getOAuthSession(bearerToken, apiKeyId);
     }
 
+    @Nullable
+    public OAuthSession extractAndValidateAudience(final HttpServletRequest request, final String apiKeyId) {
+      final OAuthSession oAuthSession = extractBearerToken(request, apiKeyId);
+      return ApiKeyUtils.validateAudience(oAuthSession, whoisKeycloakId) ? oAuthSession : new OAuthSession(apiKeyId);
+    }
+
     private OAuthSession getOAuthSession(final String bearerToken, final String apiKeyId) {
         if(StringUtils.isEmpty(bearerToken)) {
             return new OAuthSession(apiKeyId);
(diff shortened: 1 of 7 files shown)


whois-api/src/main/java/net/ripe/db/whois/api/apiKey/BearerTokenExtractor.java
```diff
@@ -26,12 +27,15 @@ public class BearerTokenExtractor   {
 
     private final ApiPublicKeyLoader apiPublicKeyLoader;
     private final boolean enabled;
+    private final String whoisKeycloakId;
 
     @Autowired
     public BearerTokenExtractor(final ApiPublicKeyLoader apiPublicKeyLoader,
-                                @Value("${apikey.authenticate.enabled:false}") final boolean enabled) {
+                                @Value("${apikey.authenticate.enabled:false}") final boolean enabled,
+                                @Value("${keycloak.idp.client:whois}") final String whoisKeycloakId) {
```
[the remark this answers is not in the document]
maggarwal13: Miguel changed it to use whoisKeycloakId. It is keycloak Id for whois

whois-api/src/main/java/net/ripe/db/whois/api/apiKey/BearerTokenExtractor.java
```diff
@@ -56,9 +60,12 @@ private OAuthSession getOAuthSession(final String bearerToken, final String apiK
               LOGGER.debug("JWT signature verification failed for {}", apiKeyId);
               return new OAuthSession(apiKeyId);
             }
-            
+
             //TODO[MA]: remove when apiKeyId is available from api registry call
-            return OAuthSession.from(new ObjectMapper().readValue(signedJWT.getPayload().toString(), OAuthSession.class), apiKeyId);
+            final OAuthSession oAuthSession = OAuthSession.from(new ObjectMapper().readValue(signedJWT.getPayload().toString(), OAuthSession.class), apiKeyId);
+
+            return ApiKeyUtils.validateAudience(oAuthSession, whoisKeycloakId) ? oAuthSession : new OAuthSession(apiKeyId);
```
eshryane, line 67: Will authentication silently fail if a `new OAuthSession()` is returned? If so can you add logging, e.g. `client id X didn't match Y` for example

maggarwal13: As INFO? or WARN?

whois-api/src/main/java/net/ripe/db/whois/api/apiKey/BearerTokenExtractor.java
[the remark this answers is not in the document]
eshryane: OK clear then thanks

whois-api/src/main/java/net/ripe/db/whois/api/apiKey/BearerTokenExtractor.java
maggarwal13: I was thinking that but we do allow user to use multiple type of authentication and also I would prefer it goes through our update flow

eshryane: OK. Is it possible to add a warning to the update and return to the user in the ack if the api key is invalid ?

maggarwal13: I think so, It should be possible. I will check

eshryane: Agreed to use message 
```The API key cannot be used because it was created for a different application or environment```

eshryane (approved):

PR #1658, opened 2025-02-12
Move NRTMv4 Servlet Code out of API Module

eshryane:
Put all NRTMv4 code in the same whois-nrtmv4 module, including servlet code and integration tests.

* Moved code from whois-api to whois-nrtmv4
* Renamed NrtmClientService.java to NrtmService for consistency also as it's a server-side controller class, not related to the nrtmv4 client code.
* Use JSR 305 Nonnull annotation


whois-api/src/main/resources/applicationContext-api.xml
```diff
@@ -12,7 +12,7 @@
 
     <bean class="net.ripe.db.whois.query.WhoisQueryConfig" />
     <import resource="applicationContext-update.xml"/>
-    <bean class="net.ripe.db.nrtm4.Nrtmv4Config" />
+<!--    <bean class="net.ripe.db.nrtm4.Nrtmv4Config" />-->
```
MiguelAHM, line 15: if this is no needed I suggest removing it

whois-api/src/test/java/net/ripe/db/whois/api/ElasticSearchHelper.java
```diff
@@ -70,7 +65,7 @@ public void resetElasticIndexes(final String indexName, final String metaDetaInd
         }
     }
 
-    @NotNull
+    @Nonnull
```
MiguelAHM, line 68: why this change?
Is NotNull wrong used? why we prioritise @Nonnull

whois-nrtm4/src/test/java/net/ripe/db/nrtm4/AbstractNrtmIntegrationTest.java
```diff
@@ -287,10 +290,10 @@ protected static List<DeltaFileRecord> getDeltaChanges(final String[] records) t
 
     protected static void assertNrtmFileInfo(final String nrtmInfo, final String type, final int version, final String source) throws JSONException {
         final JSONObject jsonObject = new JSONObject(nrtmInfo);
-        assertThat(jsonObject.getInt("nrtm_version"), is(4));
-        assertThat(jsonObject.getString("type"), is(type));
-        assertThat(jsonObject.getString("source"), is(source));
-        assertThat(jsonObject.getInt("version"), is(version));
+        MatcherAssert.assertThat(jsonObject.getInt("nrtm_version"), is(4));
```
MiguelAHM, line 293: why this change? isn't better to load just the static method that we want instead whole MatcherAssert?

whois-nrtm4/src/test/java/net/ripe/db/nrtm4/DeltaFileGenerationTestIntegration.java
```diff
@@ -321,7 +321,7 @@ public void snapshot_should_match_last_delta_version() throws JSONException {
         final UpdateNotificationFile testUpdateNotification = getNotificationFileBySource("TEST");
         final String[] testDelta = getDeltasFromUpdateNotificationBySource("TEST", 2);
 
-        assertThat(testUpdateNotification.getSnapshot().getVersion(), is(new JSONObject(testDelta[0]).getLong("version")));
+        assertThat(testUpdateNotification.getSnapshot().getVersion(), Matchers.is(new JSONObject(testDelta[0]).getLong("version")));
```
MiguelAHM, line 324: same as before I think that it is better just to load the static method instead whole class...
This is like using "*" and load more that you really need

whois-nrtm4/src/test/java/net/ripe/db/nrtm4/AbstractNrtmIntegrationTest.java
eshryane: Agree, I fixed it

eshryane: Not always the case though, sometimes it's clearer to use the class name with the static method

whois-nrtm4/src/test/java/net/ripe/db/nrtm4/DeltaFileGenerationTestIntegration.java
eshryane: Fixed

whois-api/src/test/java/net/ripe/db/whois/api/ElasticSearchHelper.java
eshryane: We are being inconsistent in using *two* types of annotation :
* JSR 305 javax.annotation.Nonnull
* Jetbrains IDEA import org.jetbrains.annotations.NotNull
I prefer to use *one* annotation (Nonnull) everywhere

eshryane: We could (maybe should) migrate away from Nonnull e.g. see 
https://github.com/google/guava/issues/2960
But that's a separate issue

whois-api/src/main/resources/applicationContext-api.xml
eshryane: Fixed

MiguelAHM (approved):

eshryane:
Integration tests passed

PR #1661, opened 2025-02-13
Removing AUT-NUM Mntner from AS-SET Will Cause AUT-NUM Update to Fail

eshryane:
If the aut-num contains a member-of: reference to a set, then that reference must be authenticated. If the aut-num mntner is removed from the as-set, then updating the aut-num object will fail.


eshryane:
Moved to #1662

PR #1666, opened 2025-02-14
TEST case demonstrating wrong hierarchy situation allowed for Override

MiguelAHM:
This happened in 2023 the child and parent were LEGACY and turned into wrong status using override user

### DIFF
--- a/whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/update/InetnumSpec.groovy
+++ b/whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/update/InetnumSpec.groovy
@@ -6159,4 +6159,48 @@ class InetnumSpec extends BaseQueryUpdateSpec {
         then:
         created =~ /Create FAILED: \[inetnum] 192.168.0.0 - 192.168.0.255/
     }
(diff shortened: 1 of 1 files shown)


whois-endtoend/src/test/groovy/net/ripe/db/whois/spec/update/InetnumSpec.groovy
```diff
@@ -6159,4 +6159,48 @@ class InetnumSpec extends BaseQueryUpdateSpec {
...
+                    tech-c:     TP1-TEST
+                    mnt-by:     LIR-MNT
+                    source:     TEST
+                    """.stripIndent(true))
+        whoisFixture.reloadTrees()
+        when:
+        def update = syncUpdate(new SyncUpdate(data: """\
+                        inetnum:    192.115.107.0 - 192.115.107.255
+                        netname:    RIPE-NCC
+                        status:     ASSIGNED PI
```
eshryane, line 6193: Change status to "ASSIGNED PA"

MiguelAHM:
Closed because it is testing a situation that should not happen

PR #1674, opened 2025-02-17
Use JWS Claimset for API Keys

maggarwal13:
** Make JWS verification mandatory
** Use JWS jose lib to load Public key
** Allow caching of public key , defaults 5 mins
** Add check to validate claimset has uuid, email, Aud and expiration time 
** Check for expired token
** improve tests


whois-api/src/main/java/net/ripe/db/whois/api/apiKey/BearerTokenExtractor.java
```diff
@@ -60,40 +67,40 @@ private OAuthSession getOAuthSession(final String bearerToken, final String apiK
...
-        try {
-            final List<RSAKey> rsaKeys = apiPublicKeyLoader.loadPublicKey();
+            final ConfigurableJWTProcessor<SecurityContext> jwtProcessor = new DefaultJWTProcessor<>();
 
-            if(rsaKeys.isEmpty()) {
-              LOGGER.debug("RSA public key is not available");
-              return true;
-            }
+            final JWKSource<SecurityContext> keySource = JWKSourceBuilder
+                    .create(new URL(jwksSetUrl))
```
MiguelAHM, line 76: I think that URL is deprecated. I remember to have this issue with NRTM Client
You can use`URI.create(jwksSetUrl).toURL()`

MiguelAHM (approved):

PR #1676, opened 2025-02-17
Log if RIPE NCC Access API calls take longer than expected

eshryane:


whois-client/pom.xml
```diff
@@ -16,6 +16,10 @@
             <groupId>net.ripe.db</groupId>
             <artifactId>whois-rpsl</artifactId>
         </dependency>
+        <dependency>
+            <groupId>net.ripe.db</groupId>
+            <artifactId>whois-commons</artifactId>
+        </dependency>
```
maggarwal13, line 22: Can we move the client to whois-api rather in adding this dependency?

eshryane: Yes! `whois-client` should only be used for connecting clients to Whois, not to other APIs !

maggarwal13: true

maggarwal13 (approved):

PR #1680, opened 2025-02-21
Generate table for email_status_history

MiguelAHM:
Move the email from email_status to email_status_history if the email is no longer referenced.
For that we need to create a new table in INTERNALS schema

### DIFF
--- a/whois-commons/src/main/resources/internals_schema.sql
+++ b/whois-commons/src/main/resources/internals_schema.sql
@@ -160,6 +160,18 @@ CREATE TABLE `email_status` (
    PRIMARY KEY (`email`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
 
+
+DROP TABLE IF EXISTS `email_status_archived`;
+CREATE TABLE `email_status_archived` (
+     `email` varchar(320) NOT NULL,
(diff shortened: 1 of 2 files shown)


maggarwal13 (approved):

whois-commons/src/main/resources/internals_schema.sql
```diff
@@ -160,6 +160,18 @@ CREATE TABLE `email_status` (
...
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
 
+
+DROP TABLE IF EXISTS `email_status_history`;
+CREATE TABLE `email_status_history` (
+     `email` varchar(320) NOT NULL,
+     `status` varchar(120) NOT NULL,
+     `message` longblob,
+     `archived_date` datetime DEFAULT now(),
+     `reported_date` datetime NOT NULL,
```
eshryane, line 170: Use `last_updated` to match email_status column name (i.e. so you can match all columns from the email_status table) and maybe `history_timestamp` or `history_last_updated` for when it was added to the history table ?

eshryane: Discussed and agreed on `email_status_update` (for the copied timestamp from email_status) and `history_update` (for the timestamp when added to history table).

PR #1681, opened 2025-02-21
refactor scopes

maggarwal13:


whois-api/src/main/java/net/ripe/db/whois/api/apiKey/BearerTokenExtractor.java
```diff
@@ -97,11 +98,21 @@ private OAuthSession getOAuthSession(final String bearerToken, final String apiK
...
-                    claimSet.getStringClaim("scope"));
+                    getWhoisScopes(claimSet.getStringClaim("scope")));
 
         } catch (Exception e) {
             LOGGER.info("Failed to read OAuthSession from BearerToken ", e);
             return new OAuthSession(apiKeyId);
         }
     }
+
+    private List<OAuthSession.ScopeFormatter> getWhoisScopes(final String scopes) {
```
MiguelAHM, line 109: This will return an empty list in next both situations:
- if the scope starts by something different than whois -> NOT ALLOWED
- if the scope is empty -> ALLOWED
Is this working fine for each of the situations?
Should we directly return an error here if the scope is not empty and doesn't start by WHOIS?

whois-api/src/test/java/net/ripe/db/whois/api/ApiKeyAuthServerDummy.java
```diff
@@ -55,15 +54,24 @@ public class ApiKeyAuthServerDummy implements Stub {
...
     public static final String BASIC_AUTH_INVALID_SIGNATURE_API_KEY = "TXp1ZzRxRVlpSTVET1dqOXI1Qkp1Y2k4OnZBdzgyRTFCMkZ2dFVyYjB0MDF0Ykt2cg==";
 
-    public static final Map<String, OAuthSession> APIKEY_TO_OAUTHSESSION =  Maps.newHashMap();
+    public static final Map<String, JWTClaimsSet> APIKEY_TO_OAUTHSESSION =  Maps.newHashMap();
 
     {
-        APIKEY_TO_OAUTHSESSION.put(BASIC_AUTH_TEST_NO_MNT, new OAuthSession(AUD, "hHZjAbXPtxGxUJCgdwv2ufhY", "test@ripe.net", "8ffe29be-89ef-41c8-ba7f-0e1553a623e5", "profile email"));
-        APIKEY_TO_OAUTHSESSION.put(BASIC_AUTH_PERSON_NO_MNT, new OAuthSession(AUD, "l6lRZgvOFIphjiGwtCGuLwqw","person@net.net", "906635c2-0405-429a-800b-0602bd716124", null));
-        APIKEY_TO_OAUTHSESSION.put(BASIC_AUTH_PERSON_OWNER_MNT, new OAuthSession(AUD, "l6lRZgvOFIphjiGwtCGuLwqw","person@net.net", "906635c2-0405-429a-800b-0602bd716124", "profile email whois.mntner:OWNER-MNT"));
-        APIKEY_TO_OAUTHSESSION.put(BASIC_AUTH_TEST_TEST_MNT, new OAuthSession(AUD, "hHZjAbXPtxGxUJCgdwv2ufhY","test@ripe.net", "8ffe29be-89ef-41c8-ba7f-0e1553a623e5", "whois.mntner:TEST-MNT profile email"));
```
MiguelAHM, line 64: mmm...I though that ApiKey service are sending us an OAuthSession

whois-commons/src/main/java/net/ripe/db/whois/common/apiKey/ApiKeyUtils.java
```diff
@@ -24,38 +22,20 @@ public class ApiKeyUtils {
...
-            return true;
-        }
-
-        final OAuthSession.ScopeFormatter scopeFormatter = new OAuthSession.ScopeFormatter(whoisScope.get());
-
-        if(StringUtils.isEmpty(scopeFormatter.getScopeKey()) || StringUtils.isEmpty(scopeFormatter.getScopeType()) || StringUtils.isEmpty(scopeFormatter.getAppName())) {
-            return true;
-        }
+    public static boolean validateMntnrScope(final OAuthSession oAuthSession, final List<RpslObject> maintainers) {
+        final List<String> mntnrScopes =  oAuthSession.getScopes().stream()
```
MiguelAHM, line 26: I think that in a previous step we are sending empty scopes in case no scope starts by "whois"
so this will be empty at this point so this method will return true?

whois-api/src/main/java/net/ripe/db/whois/api/apiKey/BearerTokenExtractor.java
maggarwal13: Any custom whois scope will always starts from whois and JWT already by default has scope " profile and email " which we should ignore. There are test cases for these situations. Scope can be or cannot be empty .

whois-commons/src/main/java/net/ripe/db/whois/common/apiKey/ApiKeyUtils.java
maggarwal13: Yes, mntnr scope is not mandatory

whois-api/src/test/java/net/ripe/db/whois/api/ApiKeyAuthServerDummy.java
maggarwal13: No, JWT token and hence its a needed improvement in test cases

PR #1686, opened 2025-02-28
Upgrade Mariadb to LTS 11.4 in CI Build Pipeline

eshryane:
MariaDB 11.4 is the current LTS (Long Term Support) release, since April 2024:
https://mariadb.org/11-4-lts/

Make sure to use this version in our CI build pipeline, as well as in all deployed environments.


eshryane:
Integration tests passed

PR #1688, opened 2025-03-03
Support oauth2

maggarwal13:


whois-api/src/main/java/net/ripe/db/whois/api/rest/HttpsAuthHeaderFiler.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @Override
+    public void doFilter(final ServletRequest request, final ServletResponse response, final FilterChain chain) throws IOException, ServletException {
+
+        if(!(request instanceof HttpServletRequest httpRequest)) {
+            chain.doFilter(request, response);
+            return;
+        }
+
+        if (StringUtils.isNotEmpty(httpRequest.getHeader(HttpHeaders.AUTHORIZATION)) && RestServiceHelper.isHttpProtocol(httpRequest)){
```
MiguelAHM, line 34: Question:
We will apply this filter always when hitting `/whois/` 
There are multiple authenticate method that doesn't require the usage of `Authorization` header like PGP key
https://docs.db.ripe.net/Authorisation/Using-the-Authorisation-Methods/

whois-api/src/main/java/net/ripe/db/whois/api/rest/HttpsBasicAuthFiler.java
```diff
@@ -31,14 +31,6 @@ public void doFilter(final ServletRequest request, final ServletResponse respons
             return;
         }
 
-        final HttpServletRequest httpRequest = (HttpServletRequest) request;
```
MiguelAHM, line 34: Is this class needed?
you are already creating a filter called `HttpsAuthHeaderFiler`

whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisServletDeployer.java
```diff
@@ -72,6 +73,7 @@ public WhoisServletDeployer(final WhoisRestService whoisRestService,
                                 final BatchUpdatesService batchUpdatesService,
                                 final HealthCheckService healthCheckService,
                                 final HttpsBasicAuthFiler httpsBasicAuthFiler,
```
MiguelAHM, line 75: should we remove this? `HttpsBasicAuthFiler`

whois-api/src/main/java/net/ripe/db/whois/api/rest/HttpsBasicAuthFiler.java
maggarwal13: It is needed till we support passwords in basic auth header. Check the HttpBasicAuthRequestWrapper

whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisServletDeployer.java
maggarwal13: see above comment

whois-api/src/main/java/net/ripe/db/whois/api/rest/HttpsAuthHeaderFiler.java
maggarwal13: Yes it will always hit 

Document says :  

PGP : When using the HTTP POST and GET requests for Syncupdates, the signed update must be supplied in the message body and PGP-signed updates cannot be submitted using the RESTful API.

MD5 Password: With the RESTful API, any required passwords must be included in the URI query parameters. 

X.509 signed updates:  When using the HTTP POST and GET requests for Syncupdates, the signed update must be supplied in the message body.

X.509 signed updates cannot be submitted via the RESTful API.

MiguelAHM (approved):

PR #1693, opened 2025-03-12
Fix RDAP Relation For Cross-Prefix Inetnum Range

MiguelAHM:
Issue caused by https://github.com/RIPE-NCC/whois/pull/1603/files
The current tree is transforming X.X.X.255 - X.X.X.0 into /31 when this should be two different /32

so using IpInterval.parse is not safe here. I am using a workaround that we are already using in other parts of the code


whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -270,4 +272,12 @@ private String objectTypesToString(final Collection<ObjectType> objectTypes) {
         return COMMA_JOINER.join(objectTypes.stream().map(ObjectType::getName).toList());
     }
 
+    private IpInterval transformToIpInterval(final Interval interval) {
```
eshryane, line 275: This method is unnecessary, please remove it

whois-rpsl/src/main/java/net/ripe/db/whois/common/ip/IpInterval.java
```diff
@@ -36,6 +36,7 @@ public static IpInterval<?> parse(final CIString prefix) {
 
     public static IpInterval<?> parse(final String prefix) {
         if (prefix.indexOf(':') == -1) {
+            // It could fail when prefix is taken from IpInterval.getKey().toString() for X.X.X.255 - X.X.X.0
```
eshryane, line 39: Remove unnecessary comment, it is too specific to this issue

MiguelAHM: I will remove it but if in the future we use the tree together with this parse we will get an error again.
And the tree only accepts IpInterval
- tree.findFirstLessSpecific(IpInterval ipInterval)
- IpInterval.parse(Interval.toString) -> Could throw an error because of this issue

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
eshryane: The method is not doing anything. You are passing an `IpInterval` and returning an `IpInterval`

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapControllerTestIntegration.java
```diff
@@ -3685,6 +3685,136 @@ public void get_bottom_wrong_type_then_400(){
...
+                        "source:         TEST");
+
+        ipTreeUpdater.rebuild();
+
+        final SearchResult searchResult = createResource("ips/rirSearch1/bottom/37.210.0.0%20-%2037.211.255.255")
+                    .request(MediaType.APPLICATION_JSON_TYPE)
+                    .get(SearchResult.class);
+
+        final List<Ip> ipResults = searchResult.getIpSearchResults();
+        assertThat(ipResults.size(), is(8));
```
eshryane, line 3815: Should you test that the cross-prefix range is included? Not just the size of the response. Maybe also remove some of the unnecessary inetnums if they are not being tested.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
MiguelAHM: no, I am passing `Interval` and returning `IpInterval`

whois-api/src/test/java/net/ripe/db/whois/api/rdap/RdapControllerTestIntegration.java
MiguelAHM: Yes, good idea

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
eshryane: All you are doing is casting the parameter and returning it, you're not transforming it. I suggest merging it into the other method.

MiguelAHM: OK, I merged that into the calling methods

eshryane: OK I see what you're doing, the method *is* necessary but please rename it from `transformToIpRangeString` to `intervalToString` for consistency (i.e. see the preceding method). Also restore the `transformToIpInterval` method as it's more readable but rename to `internalToIpInterval`.

eshryane: It's possible to "fix" this so it's unnecessary, by making `Interval` and `IpInterval` `sealed` and only allowing `Ipv4Resource` and `Ipv6Resource` as implemented classes.

whois-rpsl/src/main/java/net/ripe/db/whois/common/ip/Interval.java
```diff
@@ -10,7 +10,7 @@
  *
  * @param <K> the interval type
  */
-public interface Interval<K extends Interval<K>> {
+public sealed interface Interval<K extends Interval<K>> permits IpInterval {
```
eshryane, line 13: I moved these changes to a separate PR #1695

whois-api/src/main/java/net/ripe/db/whois/api/httpserver/dos/WhoisDoSFilter.java
```diff
@@ -64,7 +64,6 @@ protected boolean checkWhitelist(final String candidate) {
                 }
                 yield false;
             }
-            default -> false;
```
eshryane, line 67: Moved to #1695

whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
MiguelAHM: Done by your PR and updated this one which those changes

eshryane (approved):

PR #1697, opened 2025-03-14
Netty SMTP Server

eshryane:
Implement a simple SMTP server using Netty. Incoming messages are written into mailupdates.

Implemented:
* HELO EHLO MAIL RCPT DATA NOOP HELP RSET SIZE QUIT

Not implemented:
* LIMITS MAILMAX RCPTMAX
* PIPELINING 
* PIPECONNECT
* CHUNKING
* STARTTLS
* AUTH
* BDAT (binary data)
* No IP restriction (allow any client to connect)

If we need something more complex, consider integrating the Apache James project:
https://github.com/apache/james-project/tree/master


whois-smtp/src/main/java/net/ripe/db/whois/smtp/SmtpServer.java
```diff
@@ -0,0 +1,99 @@
...
+    private NioEventLoopGroup bossGroup;
+    private NioEventLoopGroup workerGroup;
+    int port;
+
+    @Autowired
+    public SmtpServer(
+            final SmtpServerChannelsRegistry smtpServerChannelsRegistry,
+            final SmtpServerChannelInitializer smtpServerChannelInitializer,
+            final MaintenanceMode maintenanceMode,
+            @Value("${mail.smtp.server.enabled:false}") final boolean smtpEnabled,
```
MiguelAHM, line 40: Feel free to ignore.
We are creating a bunch of beans that maybe are not required.
How about using @Conditional annotation, as we do in elsewhere? and put that @Condition together with @Component, @Service...etc related to this module? so any bean will be created if mail.smtp.server.enabled is false

MiguelAHM: maybe if you do this, you have an issue in the .xml when injecting dependencies...
because those dependencies are calculated at runtime, so when the .xml is read maybe are not yet there...and if the property is false will never be there.
So just ignore this

MiguelAHM (approved):

whois-smtp/src/main/java/net/ripe/db/whois/smtp/SmtpServer.java
eshryane: Thanks Miguel, making the beans conditional indeed creates dependency problems. I prefer to leave the beans initialised but enable or disable the feature with a flag as it's simpler.

PR #1698, opened 2025-03-19
NRTMv4-Client: Reject in case ahead instead of Reinitialise

MiguelAHM:
MUST verify that the UNF version is the same or higher than the client's current most recent version. If not, the UNF MUST be rejected. It is RECOMMENDED
for the client to distinguish between an UNF that is a single version older, and a much older version, in any status messages (4.3)


whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/processor/UpdateNotificationFileProcessor.java
```diff
@@ -113,7 +113,6 @@ public void processFile(){
 
             if (nrtmClientLastVersionInfo != null && nrtmClientLastVersionInfo.version() > updateNotificationFile.getVersion()) {
                 LOGGER.info("The local version cannot be higher than the update notification version {}", source);
```
eshryane, line 115: Should this log statement be an ERROR ?

MiguelAHM: Done

eshryane (approved):

PR #1700, opened 2025-03-20
RDAP - Add "-rdap" prefix to Relations And Change links according to the RFC

MiguelAHM:
last changes:
https://datatracker.ietf.org/doc/draft-ietf-regext-rdap-rir-search/16/

- `up` needs to be changed to `rdap-up`...etc (same with the rest of Relations type)
- `up-active` Link relation doesn't exist. Now it is called `rdap-up rdap-active`


whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/LinkRelationType.java
```diff
@@ -0,0 +1,32 @@
...
+
+    LinkRelationType(String value) {
+        this.value = value;
+    }
+
+    public String getValue() {
+        return value;
+    }
+
+    public static String concatActiveLinkRelation(final LinkRelationType type) {
```
eshryane, line 20: Is it more readable to make it generic such as `public static String concat(final LinkRelationType ... types)` ? Then you can specify `LinkRelationType.concat(UP, ACTIVE)` etc.

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/LinkRelationType.java
```diff
@@ -0,0 +1,32 @@
...
+        return type.value.concat(" ").concat(ACTIVE.getValue());
+    }
+
+    public static boolean containsValidValue(final String value) {
+        for (RelationType type : RelationType.values()) {
+            if (value.contains(type.getValue())) {
```
eshryane, line 26: Value must be case-sensitive? Does formatting matter? Can value contain multiple values?

MiguelAHM: This is only used for Its and it is always lowercase. So doesn't really matter

eshryane: OK

whois-api/src/main/java/net/ripe/db/whois/api/rdap/domain/LinkRelationType.java
MiguelAHM: Done

eshryane (approved):

PR #1703, opened 2025-03-21
RDAP- Fix issue with Non Existing Search IP Terms in TOP and UP

MiguelAHM:
The issue is that we are considering that search ip term always exist.
However, that's not the case...
ExactLessSpecific turns into lessSpecific if the search ip doesn't exist...dealing to weird results.

We should considered that maybe the child doesn't exist as well.


whois-api/src/main/java/net/ripe/db/whois/api/rdap/RdapRelationService.java
```diff
@@ -211,35 +211,47 @@ private IpEntry searchUpResource(final IpTree ipTree, final IpInterval searchIp)
     }
 
     private IpEntry searchTopLevelResource(final IpTree ipTree, final IpInterval searchIp){
-        final List<IpEntry> lessAndExact = ipTree.findExactAndAllLessSpecific(searchIp);;
+        final List<IpEntry> lessAndExact = ipTree.findExactAndAllLessSpecific(searchIp); //Exact only if exists
 
-        for (int countLessSpecific = 0; countLessSpecific < lessAndExact.size() - 1; countLessSpecific++){
```
MiguelAHM, line 216: lessAndExact.size() - 1;
Only works when Ip search term exist...in case the search resource doesn't exist...this just contains less
so you need to loop over all of them

eshryane: OK

eshryane (approved):

MiguelAHM:
test ok
https://gitlab.ripe.net/swe-database-team/whois/-/pipelines/847534

PR #1706, opened 2025-03-27
NRTMv4-Client: Refactor test for consistency and fix row mapper

MiguelAHM:
- IT change: Update Notification File version should match last snap/delta versions
- IT change: First delta in the Update Notification File should match the Snapshot version + 1
- IT change: For same snap (version 1) it cannot happen that just deltas 3,4 are returned. It should be 2,3,4.
- Use master datasource. We are not considering slave/master architecture.
- Fix rowmapper. New object with new values set to 0, null were returned, should be a null instead an object. Issue because MAX(version) is turned into 0 when null value.

### DIFF
--- a/whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/dao/Nrtm4ClientRepository.java
+++ b/whois-nrtm4-client/src/main/java/net/ripe/db/nrtm4/client/dao/Nrtm4ClientRepository.java
@@ -34,23 +34,20 @@ import static net.ripe.db.whois.common.dao.jdbc.JdbcRpslObjectOperations.updateL
 public class Nrtm4ClientRepository {
 
     private final JdbcTemplate jdbcMasterTemplate;
-    private final JdbcTemplate jdbcSlaveTemplate;
     private final DateTimeProvider dateTimeProvider;
 
     public Nrtm4ClientRepository(@Qualifier("nrtmClientMasterDataSource") final DataSource masterDataSource,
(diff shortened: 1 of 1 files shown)


whois-nrtm4-client/src/test/java/net/ripe/db/nrtm4/client/processor/UpdateNotificationFileProcessorTestIntegration.java
```diff
@@ -37,7 +37,7 @@ public void process_UNF_when_already_created_same_version_then_version_not_added
         updateNotificationFileProcessor.processFile();
 
         final List<NrtmClientVersionInfo> versionInfosPerSource = nrtm4ClientInfoRepository.getNrtmLastVersionInfoForUpdateNotificationFile();
-        assertThat(versionInfosPerSource.getFirst().version(), is(1L));
+        assertThat(versionInfosPerSource.getFirst().version(), is(2L));
```
eshryane, line 40: Why has the version changed?

MiguelAHM: Update Notification File version is always the latest version of snap/delta
Here we are inserting one snap with version 1 and one delta with version 2
so the update notification file version will be 2...because the last version is the one of the delta (v2)

MiguelAHM: first point described in the description above

eshryane (approved):

PR #1708, opened 2025-04-01
Set mailupdates changed timestamp to now by default

eshryane:


whois-commons/src/main/resources/patch/mailupdates-1.117.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE mailupdates MODIFY COLUMN changed int(10) unsigned DEFAULT (UNIX_TIMESTAMP());
```
MiguelAHM, line 1: I think that you need to apply this to mail updates_schema.sql as well.
I think the ITs read the schema from those files

eshryane: Yes that's done now, I made another commit

MiguelAHM (approved):

PR #1712, opened 2025-04-07
Wait maximum 10 minutes for a Zonemaster test to complete

eshryane:
The backed Zonemaster service waits up to 10 minutes for a test to complete. Whois currently waits 5 minutes. We should consider extending the Whois timeout to match Zonemaster. However this is only for outlier cases, the vast majority of tests complete in less than 5 minutes. We should also check what the impact is - e.g. are shorter tests timing out if we accept (multiple) long-running tests?

### DIFF
--- a/whois-update/src/main/java/net/ripe/db/whois/update/dns/zonemaster/ZonemasterDnsGateway.java
+++ b/whois-update/src/main/java/net/ripe/db/whois/update/dns/zonemaster/ZonemasterDnsGateway.java
@@ -37,7 +37,7 @@ public class ZonemasterDnsGateway implements DnsGateway {
     private static final Logger LOGGER = LoggerFactory.getLogger(ZonemasterDnsGateway.class);
 
     private static final int TEST_PROGRESS_SLEEP_SECONDS = 5;
-    private static final int TEST_PROGRESS_MAXIMUM_RETRIES = ((60 * 5) / TEST_PROGRESS_SLEEP_SECONDS);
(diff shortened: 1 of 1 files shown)


whois-update/src/main/java/net/ripe/db/whois/update/dns/zonemaster/ZonemasterDnsGateway.java
```diff
@@ -37,7 +37,7 @@ public class ZonemasterDnsGateway implements DnsGateway {
     private static final Logger LOGGER = LoggerFactory.getLogger(ZonemasterDnsGateway.class);
 
     private static final int TEST_PROGRESS_SLEEP_SECONDS = 5;
-    private static final int TEST_PROGRESS_MAXIMUM_RETRIES = ((60 * 5) / TEST_PROGRESS_SLEEP_SECONDS);
+    private static final int TEST_PROGRESS_MAXIMUM_RETRIES = ((60 * 10) / TEST_PROGRESS_SLEEP_SECONDS);
```
maggarwal13, line 40: We have timeout of 5 mins for requests from db-web-ui. Will this help if we wait 10 mins? User still gets 500 Internal server error

eshryane: We can also extend the read timeout in db-web-ui for the domain object wizard *only* to 10 minutes. I tested this works OK.

PR #1715, opened 2025-04-10
NRTMv4 Client Do Not Throw Error When Non Parseable Delta

MiguelAHM:
If a delta is not correctly formed we just skip that delta


eshryane:
This change seems to conflict with the spec NOT to skip rejected deltas?

```
   If the Update Notification File or one of the Delta Files is
   rejected, the mirror client MUST NOT process any newer Deltas than
   those that are valid and have been successfully verified.  If some
   Delta Files are rejected, it MAY process the valid Delta Files, but
   MUST NOT skip over any rejected Delta Files while doing so.
   Additionally, the changes in a specific Delta File MUST be processed
   either completely, or not at all, i.e. a Delta File must never be
   partially processed.
```

Ref. https://htmlpreview.github.io/?https://github.com/mxsasha/nrtmv4/blob/main/draft-ietf-grow-nrtm-v4.html#section-4.3

eshryane (approved):

PR #1716, opened 2025-04-15
Configure Velocity classpath options for increased performance

eshryane:
Set the following Velocity classpath options : 
* classpath.resource.loader.cache = true (cache templates)
* classpath.resource.loader.modificationCheckInterval = 0 (don't reload templates)

Ref. https://velocity.apache.org/engine/2.0/configuration.html

Profiled Whois update using Whois REST API integration tests : 
* org.apache.velocity.app.VelocityEngine.getTemplate(String) went from 120ms to 10ms.
* org.apache.velocity.Template.merge(Context, Writer) went from 80ms to 10ms.


maggarwal13 (approved):

eshryane:
Integration tests passed

PR #1723, opened 2025-04-23
Do not apply ACL accounting for an Authenticated Maintainer's own Objects

MiguelAHM:
- We decided to not ACL account personal objects that are owned by the user making the request.
- This should stop blocking people that is just querying for their own personal information.
- We only can identify a person if the SSO token is present in the request. This meant that this only can be applied to: SSO authentication, OAUTH2 authentication and APIKEYs authentication.

### DIFF
--- a/whois-api/src/test/java/net/ripe/db/whois/api/ApiKeyAuthServerDummy.java
+++ b/whois-api/src/test/java/net/ripe/db/whois/api/ApiKeyAuthServerDummy.java
@@ -18,8 +18,8 @@ import jakarta.servlet.http.HttpServletResponse;
 import jakarta.ws.rs.NotAuthorizedException;
 import jakarta.ws.rs.NotFoundException;
 import net.ripe.db.whois.common.Stub;
-import net.ripe.db.whois.common.oauth.ApiKeyAuthServiceClient;
 import net.ripe.db.whois.common.aspects.RetryFor;
+import net.ripe.db.whois.common.oauth.ApiKeyAuthServiceClient;
 import net.ripe.db.whois.common.profiles.WhoisProfile;
 import org.apache.commons.lang3.StringUtils;
 import org.eclipse.jetty.server.NetworkConnector;
@@ -57,6 +57,7 @@ public class ApiKeyAuthServerDummy implements Stub {
     public static final String BASIC_AUTH_PERSON_OWNER_MNT_WRONG_AUDIENCE = "aFR0cm9lZUpWYWlmSWNQR1BZUW5kSmhnOmp5akhYR2g4WDFXRWZyc2M5SVJZcUVYbw==";
     public static final String BASIC_AUTH_INVALID_API_KEY = "aDZsUlpndk9GSXBoamlHd3RDR3VMd3F3OjJDVEdQeDVhbFVFVzRwa1Rrd2FRdGRPNg==";
     public static final String BASIC_AUTH_INVALID_SIGNATURE_API_KEY = "TXp1ZzRxRVlpSTVET1dqOXI1Qkp1Y2k4OnZBdzgyRTFCMkZ2dFVyYjB0MDF0Ykt2cg==";
+    public static final String BASIC_AUTH_PERSON_OWNING_MNT = "aAAsUlpndk9GSXBoamlHd3RDR3VMd3F3OjJDVEdQeDVhbFVFVzRwa1Rrd2FRdGRPNg==";
 
     public static final Map<String, JWTClaimsSet> APIKEY_TO_OAUTHSESSION =  Maps.newHashMap();
 
@@ -68,6 +69,7 @@ public class ApiKeyAuthServerDummy implements Stub {
         APIKEY_TO_OAUTHSESSION.put(BASIC_AUTH_TEST_TEST_MNT,  getJWT(AUD, "test@ripe.net", "8ffe29be-89ef-41c8-ba7f-0e1553a623e5", "whois.mntner:TEST-MNT profile email"));
         APIKEY_TO_OAUTHSESSION.put(BASIC_AUTH_INVALID_SIGNATURE_API_KEY,  getJWT(AUD, "invalid@ripe.net", "8ffe29be-89ef-41c8-ba7f-0e1553a623e5", "profile email whois.mntner:TEST-MNT"));
         APIKEY_TO_OAUTHSESSION.put(BASIC_AUTH_PERSON_OWNER_MNT_WRONG_AUDIENCE, getJWT(Arrays.asList("account", "whois-invalid"), "person@net.net", "906635c2-0405-429a-800b-0602bd716124", "profile email whois.mntner:TEST-MNT"));
+        APIKEY_TO_OAUTHSESSION.put(BASIC_AUTH_PERSON_OWNING_MNT,  getJWT(AUD, "person@net.net", "906635c2-0405-429a-800b-0602bd716124", "profile email whois.mntner:USER-OWNING-MNT"));
     }
 
     private Server server;
--- a/whois-api/src/test/java/net/ripe/db/whois/api/ApiKeyAuthServerDummy.java
+++ b/whois-api/src/test/java/net/ripe/db/whois/api/ApiKeyAuthServerDummy.java
@@ -57,7 +57,6 @@ public class ApiKeyAuthServerDummy implements Stub {
     public static final String BASIC_AUTH_PERSON_OWNER_MNT_WRONG_AUDIENCE = "aFR0cm9lZUpWYWlmSWNQR1BZUW5kSmhnOmp5akhYR2g4WDFXRWZyc2M5SVJZcUVYbw==";
     public static final String BASIC_AUTH_INVALID_API_KEY = "aDZsUlpndk9GSXBoamlHd3RDR3VMd3F3OjJDVEdQeDVhbFVFVzRwa1Rrd2FRdGRPNg==";
     public static final String BASIC_AUTH_INVALID_SIGNATURE_API_KEY = "TXp1ZzRxRVlpSTVET1dqOXI1Qkp1Y2k4OnZBdzgyRTFCMkZ2dFVyYjB0MDF0Ykt2cg==";
-    public static final String BASIC_AUTH_PERSON_OWNING_MNT = "aAAsUlpndk9GSXBoamlHd3RDR3VMd3F3OjJDVEdQeDVhbFVFVzRwa1Rrd2FRdGRPNg==";
 
     public static final Map<String, JWTClaimsSet> APIKEY_TO_OAUTHSESSION =  Maps.newHashMap();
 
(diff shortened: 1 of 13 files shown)


whois-query/src/main/java/net/ripe/db/whois/query/acl/AccountingIdentifier.java
```diff
@@ -2,21 +2,5 @@
...
-    }
-
-    public String getUserName() {
-        return userName;
-    }
-
-    public InetAddress getRemoteAddress() {
-        return remoteAddress;
-    }
+public record AccountingIdentifier(InetAddress remoteAddress, String userName, String ssoToken) {
```
MiguelAHM, line 5: I decided to add the ssoToken coming from the request header in here, as an extra user identification step
There is another option to just keep the userName (as it is) and call the auth service to get the related ssoToken (it is 1 to 1 relation). However, this would allow the user to bypass the counting, putting a wrong email-uuid relation...(we would be ignoring the UUID from the request).
This is needed because in the Auth Index table (which I need to use because of task requirement) we are storing "SSO <UUID>" as value for `Auth:` attribute

whois-query/src/main/java/net/ripe/db/whois/query/acl/AccountingIdentifier.java
```diff
@@ -6,10 +6,12 @@ public class AccountingIdentifier {
 
     final private InetAddress remoteAddress;
     final private String userName;
+    private final String ssoToken;
```
MiguelAHM, line 9: I decided to add the ssoToken coming from the request header in here, as an extra user identification step

There is another option to just keep the userName (as it is) and call the auth service to get the related ssoToken (it is 1 to 1 relation). However, this would allow the user to bypass the counting, putting a wrong email-uuid relation...(we would be ignoring the UUID from the request).

This is needed because in the Auth Index table (which I need to use because of task requirement) we are storing "SSO " as value for Auth: attribute. So for fetching the maintainters belonging to that user we need the UUID and not the email.

whois-query/src/main/java/net/ripe/db/whois/query/acl/AccessControlListManager.java
```diff
@@ -65,13 +72,18 @@ public AccessControlListManager(final DateTimeProvider dateTimeProvider,
...
+        this.jdbcRpslObjectSlaveDao = jdbcRpslObjectSlaveDao;
     }
 
-    public boolean requiresAcl(final RpslObject rpslObject, final Source source) {
+    public boolean requiresAcl(final RpslObject rpslObject, final Source source, final AccountingIdentifier accountingIdentifier) {
         if (source.isGrs()) {
             return false;
         }
 
+        if (!StringUtil.isNullOrEmpty(accountingIdentifier.getSsoToken()) && isUserOwnedObject(rpslObject, accountingIdentifier.getSsoToken())){
```
maggarwal13, line 83: Can you use username ? Why you need crowdToken? If you start using crowdToken , it will not be universal as oauth2 and apikey will not have sso token

whois-query/src/main/java/net/ripe/db/whois/query/acl/AccessControlListManager.java
```diff
@@ -148,6 +160,11 @@ private String getUserName(final String ssoToken, final OAuthSession oAuthSessio
         return null;
     }
 
+    @Nullable
+    private String getSsoTokenFromOauth(final OAuthSession oAuthSession){
```
maggarwal13, line 164: Why qury happens it always accounts for username why you need oauthsession here ? We should not add that differentiation while counting for personal objects

whois-query/src/main/java/net/ripe/db/whois/query/acl/AccessControlListManager.java
MiguelAHM: please read my comment below

``` 
I decided to add the ssoToken coming from the request header in here, as an extra user identification step

There is another option to just keep the userName (as it is) and call the auth service to get the related ssoToken (it is 1 to 1 relation). However, this would allow the user to bypass the counting, putting a wrong email-uuid relation...(we would be ignoring the UUID from the request).

This is needed because in the Auth Index table (which I need to use because of task requirement) we are storing "SSO " as value for Auth: attribute. So for fetching the maintainters belonging to that user we need the UUID and not the email.
```

For reverse query I need the UUID

maggarwal13: SSOToken is different then uuid , SSOToken gets translated into uuid and username via authservice client, I think you meant to uuid only?

whois-query/src/main/java/net/ripe/db/whois/query/acl/AccessControlListManager.java
```diff
@@ -163,6 +180,11 @@ public void accountPersonalObjects(final AccountingIdentifier accountingIdentifi
         accountingManager.accountPersonalObjects(amount);
     }
 
+    private boolean isUserOwnedObject(final RpslObject rpslObject, final String ssoToken){
+        final List<RpslObjectInfo> mntnerInfoList = jdbcRpslObjectSlaveDao.findByAttribute(AttributeType.AUTH, "SSO " + ssoToken);
+        return mntnerInfoList.stream().anyMatch(rpslObjectInfo -> rpslObject.getValueForAttribute(AttributeType.MNT_BY).contains(rpslObjectInfo.getKey()));
+    }
```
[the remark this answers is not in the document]
MiguelAHM: yes, I totally forgot to put valueS
I am using the correct method now `getValuesForAttribute` instead of `getValueForAttribute`

whois-query/src/main/java/net/ripe/db/whois/query/acl/AccountingIdentifier.java
MiguelAHM: makes sense. Here we only put the credentials that we are accounting for.
In this way looks like that we are accounting for an extra auth method called ssoToken...

In my last change I merge UserName with the uuid...so there are still just 2 accounting methods but the userName is now ssoUser which contains both the uuid and the userName. Hope it is clear now

whois-query/src/main/java/net/ripe/db/whois/query/acl/AccountingIdentifier.java
```diff
@@ -5,18 +5,20 @@
...
-        return userName;
+    public UserInfo getSsoUser() {
+        return ssoUser;
     }
 
     public InetAddress getRemoteAddress() {
         return remoteAddress;
     }
+
+    public record UserInfo(String userName, String uuid) {}
```
maggarwal13, line 23: Is it possible to send the list of maintainers for a uuid as parameter to requiresAcl? AccountingIdentifier class  should have only those fields for which we do accounting. This function  requiresAcl is being called from two places only from a queryHandler and ElasticSearchCallback. This also helps to keep the accessListManager class granular too i.e it just do counting for personal objects and related operations  rather than hitting the database and fetching the maintainers too.

whois-query/src/main/java/net/ripe/db/whois/query/acl/AccessControlListManager.java
```diff
@@ -65,13 +72,18 @@ public AccessControlListManager(final DateTimeProvider dateTimeProvider,
...
+        this.jdbcRpslObjectSlaveDao = jdbcRpslObjectSlaveDao;
     }
 
-    public boolean requiresAcl(final RpslObject rpslObject, final Source source) {
+    public boolean requiresAcl(final RpslObject rpslObject, final Source source, final AccountingIdentifier accountingIdentifier) {
         if (source.isGrs()) {
             return false;
         }
 
+        if (accountingIdentifier.getSsoUser() != null && isUserOwnedObject(rpslObject, accountingIdentifier.getSsoUser().uuid())){
```
maggarwal13: on a second thought , there is no need to check for isOwnedObject if objectType is not PERSON or Role. this check should go below the person or role check. In this case it makes sense to fetch the maintainers here only.

MiguelAHM: I agree with this. I just refactored that method to check if it not a person or role beforehand

whois-api/src/test/java/net/ripe/db/whois/api/ApiKeyAuthServerDummy.java
```diff
@@ -57,6 +57,7 @@ public class ApiKeyAuthServerDummy implements Stub {
     public static final String BASIC_AUTH_PERSON_OWNER_MNT_WRONG_AUDIENCE = "aFR0cm9lZUpWYWlmSWNQR1BZUW5kSmhnOmp5akhYR2g4WDFXRWZyc2M5SVJZcUVYbw==";
     public static final String BASIC_AUTH_INVALID_API_KEY = "aDZsUlpndk9GSXBoamlHd3RDR3VMd3F3OjJDVEdQeDVhbFVFVzRwa1Rrd2FRdGRPNg==";
     public static final String BASIC_AUTH_INVALID_SIGNATURE_API_KEY = "TXp1ZzRxRVlpSTVET1dqOXI1Qkp1Y2k4OnZBdzgyRTFCMkZ2dFVyYjB0MDF0Ykt2cg==";
+    public static final String BASIC_AUTH_PERSON_OWNING_MNT = "aAAsUlpndk9GSXBoamlHd3RDR3VMd3F3OjJDVEdQeDVhbFVFVzRwa1Rrd2FRdGRPNg==";
```
maggarwal13, line 61: Is it possible to use the existing apikeys? We are anyways checking with uuid and email so should be possible to use existing one, else this list will get too long if we add token for every ussecase?

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceSSOAclTestIntegration.java
```diff
@@ -36,8 +32,10 @@ public class WhoisRestServiceSSOAclTestIntegration extends AbstractIntegrationTe
     private static final String LOCALHOST = "127.0.0.1";
     private static final String LOCALHOST_WITH_PREFIX = "127.0.0.1/32";
     public static final String VALID_TOKEN_USER_NAME = "person@net.net";
+    public static final String DB_E2E_1_USER_NAME = "db_e2e_1@ripe.net";
     public static final UserSession VALID_TOKEN_USER_SESSION = new UserSession("aff2b59f-7bd0-413b-a16f-5bc1c5c3c3ef","person@net.net", "Test User", true, "2033-01-30T16:38:27.369+11:00");
     public static final String VALID_TOKEN = "valid-token";
+    public static final String DB_E2E_1_TOKEN = "db_e2e_1";
```
maggarwal13, line 38: Can you re-use the VALID_TOKEN ? Mntner SSO uuid should be of test@ripe.net?

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceSSOAclTestIntegration.java
```diff
@@ -140,7 +138,7 @@ public void lookup_person_using_sso_acl_counted_no_ip_counted() throws Exception
 
         assertThat(whoisResources.getWhoisObjects().get(0).getAttributes()
                             .stream()
-                            .anyMatch( (attribute)-> attribute.getName().equals(AttributeType.E_MAIL)),
+                            .anyMatch( (attribute)-> attribute.getName().equals(AttributeType.E_MAIL.getName())),
```
maggarwal13, line 141: Good catch :)

whois-api/src/test/java/net/ripe/db/whois/api/ApiKeyAuthServerDummy.java
MiguelAHM: I removed this

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceSSOAclTestIntegration.java
MiguelAHM: Done

maggarwal13 (approved):

PR #1741, opened 2025-04-28
Close SMTP connection on read timeout

eshryane:


whois-smtp/src/main/java/net/ripe/db/whois/smtp/SmtpServerExceptionHandler.java
```diff
@@ -28,30 +27,25 @@ public void exceptionCaught(final ChannelHandlerContext ctx, final Throwable exc
 
         switch (exception) {
```
MiguelAHM, line 28: Do you mind explaining why the connection should be closed just when timeout?
In the task it doesn't say the reason

eshryane: Because we don't want a connection to stay idle and open, a misbehaving client can consume a lot of resources that way. If a client stops sending commands and we hit the idle timeout, then we should disconnect the client.

MiguelAHM (approved):

PR #1744, opened 2025-04-29
Fix regression in Log SMTP DATA message content

eshryane:
See #1740 failing test


eshryane:
Integration tests passed

PR #1748, opened 2025-05-07
Allow Override to Authenticate (non-filtered) Lookups

MiguelAHM:
Allow override to authenticate a Whois REST API lookup, in addition to passwords and RIPE NCC Access.


whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/transform/FilterAuthFunction.java
```diff
@@ -4,9 +4,12 @@
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
-import net.ripe.db.whois.common.oauth.OAuthSession;
+import io.netty.util.internal.StringUtil;
```
eshryane, line 7: I don't think you mean to use that class? 

You should add "internal" packages to the "Exclude from auto-import and completion" in IntelliJ.

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/transform/FilterAuthFunction.java
```diff
@@ -46,13 +47,18 @@ public class FilterAuthFunction implements FilterFunction {
...
+    private boolean isTrusted;
     private OAuthSession oAuthSession;
     private UserSession  userSession;
     private RpslObjectDao rpslObjectDao = null;
+    private UserDao userDao;
     private AuthServiceClient authServiceClient;
     private List<X509CertificateWrapper> certificates;
     private ClientAuthCertificateValidator clientAuthCertificateValidator;
 
+    private static final Splitter OVERRIDE_SPLITTER = Splitter.on(',').trimResults().limit(3);
```
eshryane, line 60: I guess you're copying from `OverrideCredential` ? Maybe move the Override class to the common package?

eshryane: You could leave the OverrideCredential in update but create an Override domain class in commons.

MiguelAHM: To be honest I didn't want to do any refactor here. I can go with a little one if needed yes to avoid duplicated logic

maggarwal13: I think we should use overrideOptiosn class and a method that takes string

whois-query/src/main/java/net/ripe/db/whois/query/planner/RpslResponseDecorator.java
```diff
@@ -2,27 +2,27 @@
...
+import net.ripe.db.whois.common.x509.ClientAuthCertificateValidator;
+import net.ripe.db.whois.common.x509.X509CertificateWrapper;
 import net.ripe.db.whois.query.QueryMessages;
 import net.ripe.db.whois.query.domain.MessageObject;
 import net.ripe.db.whois.query.executor.decorators.DummifyDecorator;
 import net.ripe.db.whois.query.executor.decorators.FilterPersonalDecorator;
 import net.ripe.db.whois.query.executor.decorators.FilterPlaceholdersDecorator;
 import net.ripe.db.whois.query.query.Query;
-import org.apache.commons.lang3.StringUtils;
+import org.elasticsearch.common.Strings;
```
maggarwal13, line 25: Wrong import? We should use stringUtils

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/transform/FilterAuthFunction.java
maggarwal13: If we support authenticated queries , the whole credential package should be in commons and this way it would be easy to keep the authentication method at one place and easy to maintain and consistent. But I guess that requires some refactor seperate from this PR.Lets have a discussion

whois-query/src/main/java/net/ripe/db/whois/query/planner/RpslResponseDecorator.java
eshryane: We should only use the Apache commons library for this, not any "internal" string handling classes from other libraries, that is not it's main function. I agree we should only use elasticsearch packages for full text search.

We should migrate from lang(2) to lang3, currently *both* are Whois dependencies, because some lang3 methods throw NullPointerException which is an unwelcome change and we need to either expect that or write our own. As discussed I'll add a new task to migrate off lang(2).

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/transform/FilterAuthFunction.java
eshryane: I'd prefer not to expose internal logic like the splitter used for handling override. Better to move or create an Override class in the commons module and use it both in query and override.

maggarwal13: Can you shift OverrideValues in OverrideCredential to commons, there is function that parses the string value into OverrideValues, this shoudl also shift

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/transform/FilterAuthFunction.java
MiguelAHM: Talked over huddle. I am adding the internal packages into that mentioned list?

whois-query/src/main/java/net/ripe/db/whois/query/planner/RpslResponseDecorator.java
MiguelAHM: I updated this using StringUtils

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/transform/FilterAuthFunction.java
MiguelAHM: this is done as part of 
https://github.com/RIPE-NCC/whois/pull/1751
and 
https://github.com/RIPE-NCC/whois/pull/1750

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/transform/FilterAuthFunction.java
```diff
@@ -72,12 +84,15 @@ public FilterAuthFunction() {
     @Override @Nonnull
     public RpslObject apply(final RpslObject rpslObject) {
         final List<RpslAttribute> authAttributes = rpslObject.findAttributes(AttributeType.AUTH);
-        if (authAttributes.isEmpty()) {
+        if (authAttributes.isEmpty() && override == null) {
```
eshryane, line 87: Why `override == null` ? There isn't a similar check when using password auth?

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/transform/FilterAuthFunction.java
```diff
@@ -72,12 +84,15 @@ public FilterAuthFunction() {
...
             return rpslObject;
         }
 
         final Map<RpslAttribute, RpslAttribute> replace = Maps.newHashMap();
-        final boolean authenticated = isMntnerAuthenticated(rpslObject);
+        final boolean authenticated = isMntnerAuthenticated(rpslObject) ||
```
eshryane, line 92: Can you move override check into `isMntnerAuthenticated` ?

MiguelAHM: they are different.
override doesn't have nothing to do with the mntner. isMntnerAUthenticated is checking the auth attribute of the mntner 
If I do that I should change the name of that method

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/transform/FilterAuthFunction.java
MiguelAHM: for the early return that we are doing there is necessary to check `override` as well.
If not we could be early returning in case `override` is there.
Override doesn't have nothing to do with authAttributes.

In the update we are checking that we are using override yes, `update.isOverride()` 
If this is true then we do an overrideAuthentication

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceTestIntegration.java
```diff
@@ -98,8 +99,12 @@
 
 // FIXME: make this into a suite that runs twice: once with XML, once with JSON
 @Tag("IntegrationTest")
+@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_EACH_TEST_METHOD)
```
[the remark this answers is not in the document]
MiguelAHM: The test started failing. DirtiesContext is just encapsulating each method, so this means for some reason with new changes on method is being affected by another one.
Anyways I don't see any issue here because we are calling a class with threadSafe annotation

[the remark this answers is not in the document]
MiguelAHM: adding 
`ipRanges.setTrusted("127.0.0.1");` 
in the setup() to run for each test is working as well.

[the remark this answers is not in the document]
eshryane: Can you remove the `@DirtiesContext` ? Tests shouldn't be failing because of your changes.

Also localhost is already in the properties so that is not necessary, see whois-commons/src/test/resources/whois.properties : 

```
ipranges.trusted=127.0.0.1,::1
```

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/transform/FilterAuthFunction.java
eshryane: It's a confusing change. Does authAttributes have to be empty *AND* override be null ? Can you add a separate `if` statement just for override ?

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/transform/FilterAuthFunction.java
eshryane: Can you create a separate method for override such as `isOverrideAuthenticated()` and move the logic into that? So it's clearer to read

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceTestIntegration.java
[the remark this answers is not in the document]
MiguelAHM: `@DirtiesContext` is no longer there
This property is overlapped for some tests 
that are doing `ipRanges.setTrusted("::0");` 
why do we have localhost always there, are we never testing non trusted?
I created some ITs for non trusted
Have a look at `lookup_with_override_non_trusted_ip_with_sso_succeeds`

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/transform/FilterAuthFunction.java
MiguelAHM: yes, authAttributes have to be empty AND override be null for the early return.
another way to read it...we need override OR at least one auth Attribute to make the authentication

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestServiceTestIntegration.java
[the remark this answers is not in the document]
MiguelAHM: Using `clientIp` as suggested by Mahesh.
I removed the ipRanges from the IT

whois-query/src/main/java/net/ripe/db/whois/query/query/Query.java
```diff
@@ -694,6 +700,10 @@ public String getCharsetName(){
         }
     }
 
+    public OverrideCredential getOverride() {
```
maggarwal13, line 703: may be use auto-generated getter and setter method names? Or is it like this for rest of getter methods too ?

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/transform/FilterAuthFunction.java
```diff
@@ -105,6 +118,15 @@ public RpslObject apply(final RpslObject rpslObject) {
         }
     }
 
+    private boolean canAuthenticate(final List<RpslAttribute> attributes){
+        return !attributes.isEmpty() || override != null;
+    }
```
maggarwal13, line 123: Why we just need to check for override, not for password or any other auth methods?

whois-query/src/main/java/net/ripe/db/whois/query/query/Query.java
```diff
@@ -136,12 +139,15 @@ public static Query parse(final String args, final UserSession userSession, fina
...
-    public static Query parse(final String args, final UserSession userSession, final List<String> passwords, final boolean trusted, final List<X509CertificateWrapper> certificates, final OAuthSession oAuthSession) {
+    public static Query parse(final String args, final UserSession userSession, final List<String> passwords,
+                              final String override, final boolean trusted,
+                              final List<X509CertificateWrapper> certificates, final OAuthSession oAuthSession) {
         final Query query = parse(args, Origin.REST, trusted);
         query.userSession = userSession;
         query.passwords = passwords;
         query.certificates = certificates;
         query.oAuthSession = oAuthSession;
+        query.override = StringUtils.isEmpty(override) ? null : OverrideCredential.parse(override);
```
maggarwal13, line 150: I think what you should be checking is OverrideValues in OverrideCredential is present or not , it might very well happen that override value is wrongly formatted in that case we put optional null in OverrideValues. Just checking override == null is not enough ?

whois-query/src/main/java/net/ripe/db/whois/query/planner/RpslResponseDecorator.java
```diff
@@ -157,16 +160,18 @@ private Iterable<? extends ResponseObject> filterChanged(final Iterable<? extend
...
+        final OverrideCredential override = query.getOverride();
 
         final FilterAuthFunction filterAuthFunction =
-                (CollectionUtils.isEmpty(passwords) && userSession == null && hasNotCertificates(certificates) && oAuthSession == null)?
+                (CollectionUtils.isEmpty(passwords) && override == null && userSession == null && hasNotCertificates(certificates) && oAuthSession == null)?
                         FILTER_AUTH_FUNCTION :
-                        new FilterAuthFunction(passwords, oAuthSession, userSession, authServiceClient,
-                                rpslObjectDao, certificates, clientAuthCertificateValidator);
+                        new FilterAuthFunction(passwords, override, oAuthSession, userSession, authServiceClient,
+                                rpslObjectDao, certificates, clientAuthCertificateValidator,
```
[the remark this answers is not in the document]
MiguelAHM: I believe that because we do not need all the info that we have in the Query
so we would not comply with coupling is we pass more info that is required
https://www.baeldung.com/cs/cohesion-vs-coupling

whois-query/src/main/java/net/ripe/db/whois/query/query/Query.java
MiguelAHM: I used auto-generated but the name of the attribute is `override` so this is why you see getOverride
I will rename it to `overrideCredential` so it will change to getOverrideCredential

whois-query/src/main/java/net/ripe/db/whois/query/query/Query.java
MiguelAHM: I move that check into OverrideCredential

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/transform/FilterAuthFunction.java
MiguelAHM: let say this
if canAuthenticate we check if isMntnerAuthenticated or isOverrideAuthenticated

So the credentials comes later, in canAuthenticate we only check if it has the requirements.

maggarwal13: I think it is confusing to check here.  

if authAttributes are empty but overrideCredentials are not , it will return true? 
In that case it should return false , no? 

Also just checking  overrideCredentials != null is not enough, you need to check overrideValues is present or not ?

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/transform/FilterAuthFunction.java
```diff
@@ -105,6 +118,15 @@ public RpslObject apply(final RpslObject rpslObject) {
         }
     }
 
+    private boolean canAuthenticate(final List<RpslAttribute> attributes){
+        return !attributes.isEmpty() || overrideCredential != null;
+    }
+
+    private boolean isOverrideAuthenticated(final ObjectType objectType){
+        return overrideCredentialValidator != null && overrideCredentialValidator.isAllowedAndValid(isTrusted,
```
maggarwal13, line 126: overrideCredentialValidator is a component, it can never be null?

MiguelAHM: it can be null
FilterAuthFunction is not a component
Please have a look at RpslResponseDecorator 
FILTER_AUTH_FUNCTION -> This setting everything to null
and the is calling apply in line 178

MiguelAHM: FilterAuthFunction should be a component...but I don't want to keep making refactors

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/transform/FilterAuthFunction.java
MiguelAHM: is it working as expected,
Override doesn't have any kind of relation with the auth attributes. This is not happening for the rest of authentications

maggarwal13: Only these four conditions can happen 

    // !true || true) == true (why ??) there are no auth attributes so why authenticate
    // !true )  == false
    // !false ) == true
    // !false ) == true

MiguelAHM: yes, the name of the class is FilterAuthAttributes...so if there is no one then we return doesn't matter what.
Sorry for the confusion, you are totally right I reverted the change

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/transform/FilterAuthFunction.java
MiguelAHM: Discussed with Mahesh, I changed this

whois-commons/src/main/java/net/ripe/db/whois/common/override/OverrideCredentialValidator.java
```diff
@@ -23,11 +23,26 @@ public OverrideCredentialValidator(final IpRanges ipRanges,
         this.userDao = userDao;
     }
 
+    public boolean isAllowedAndValid(final boolean isTrusted, final UserSession userSession,
+                                     final OverrideCredential overrideCredential,
+                                     final ObjectType objectType){
+        if (overrideCredential == null) {
+            return false;
+        }
+        return overrideCredential.getOverrideValues().filter(values -> (isTrusted || isAllowedBySSO(userSession, values.getUsername())) &&
```
maggarwal13, line 32: Can you return early if it is not trusted? will be much more readable than a check in streams

maggarwal13 (approved):
Approved with a minor change

whois-commons/src/main/java/net/ripe/db/whois/common/override/OverrideCredentialValidator.java
MiguelAHM: mmmm....I cannot do that if it is a list
I can just get the first value...in GET request you only can have one override

PR #1749, opened 2025-05-07
Run RIR Authoritative Resources Task Every Hour and Import Data Only When Updated

eshryane:
Currently the Whois Authoritative Resource Import Task runs once a day, at midnight UTC, to download the authoritative resources files from all RIRs simultaneously. However, other RIRs files are updated at different times of the day, so we often are (very) out of date.

We will instead run this job once an hour, so we can load an RIR's file with less delay. This change has the advantage of loading an RIRs authoritative resource file as soon as it's updated. Also if an RIR updates the file multiple times a day, we can keep our list of their resources more up-to-date.


eshryane:
Integration tests passed

maggarwal13 (approved):

PR #1752, opened 2025-05-08
Make GRS Import Steps Transactional

eshryane:
Make each step of the GRS (mirror databases) import task transactional, i.e.
- Import IRR Objects
- Import Objects
- Delete Objects Not Found In Import
- Update Indexes

So that the GRS database is more consistent. We do not want an inconsistent GRS database while the import is in progress (e.g. for handling user queries).

This is done by configuring a Transaction Manager for each GRS datasource, then programatically create a transaction whenever it's needed.

Do not create a single transaction for the entire import, as it's huge and not necessary. Each step is consistent.

Also make sure the GRS import runs *after* the Authoritative Resources Import job, so we know which resources belong in each GRS mirror.


eshryane:
Integration tests have passed

whois-scheduler/src/main/java/net/ripe/db/whois/scheduler/task/grs/GrsDao.java
```diff
@@ -54,9 +57,15 @@ private void ensureInitialized() {
             JdbcRpslObjectOperations.sanityCheck(slaveJdbcTemplate);
             this.masterJdbcTemplate = masterJdbcTemplate;
             this.slaveJdbcTemplate = slaveJdbcTemplate;
+            this.transactionTemplate = new TransactionTemplate(new DataSourceTransactionManager(masterJdbcTemplate.getDataSource()));
```
maggarwal13, line 60: We should be adding all transactionManager in TransactionConfiguration.

eshryane: I tried but GRS sources are dynamic. Is there a way to inject GRS sources and provide a TransactionTemplate?

Or, add a static method to TransactionConfiguration like : 
```
public static TransactionManager transactionManager(final DataSource datasource) {
    return new DataSourceTransactionManager(datasource);
}
```

PR #1754, opened 2025-05-12
support ANY scope in MNTR

maggarwal13:
use flag apikey.scope.mandatory to control scope is mandatory or not

### DIFF
--- a/whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestOAuthScopeOptionalTestIntegration.java
+++ b/whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestOAuthScopeOptionalTestIntegration.java
@@ -1,0 +1,210 @@
+package net.ripe.db.whois.api.rest;
+
+import jakarta.ws.rs.core.HttpHeaders;
+import net.ripe.db.whois.api.SecureRestTest;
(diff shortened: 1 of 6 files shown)


whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisRestOAuthScopeOptionalTestIntegration.java
```diff
@@ -0,0 +1,210 @@
...
+    public void reset() throws Exception {
+        databaseHelper.clearAclTables();
+
+        ipResourceConfiguration.reload();
+        ssoResourceConfiguration.reload();
+        testPersonalObjectAccounting.resetAccounting();
+    }
+
+    @Test
+    public void create_mntner_only_data_parameter_with_apiKey_fails_null_Scope() {
```
MiguelAHM, line 131: succeed instead of fails?

MiguelAHM: same for the rest of test here?
I expect that if it is 
` System.setProperty("apikey.scope.mandatory","false");`

The test will succeed if the scope is null

MiguelAHM (approved):

PR #1758, opened 2025-05-14
Send Filtered Data When Override is Unknown in Lookup

MiguelAHM:


whois-commons/src/main/java/net/ripe/db/whois/common/override/OverrideCredentialValidator.java
```diff
@@ -30,7 +30,12 @@ public boolean isAllowedAndValid(final boolean isTrusted, final UserSession user
             return false;
         }
         final OverrideCredential.OverrideValues overrideValues = overrideCredential.getOverrideValues().get();
-        return isAllowedToUseOverride(isTrusted, userSession, overrideValues.getUsername()) && isValidOverride(overrideValues, objectType);
+
+        try {
+            return isAllowedToUseOverride(isTrusted, userSession, overrideValues.getUsername()) && isValidOverride(overrideValues, objectType);
+        } catch (EmptyResultDataAccessException e){
```
MiguelAHM, line 36: the try catch for the update (which is directly calling `isValidOverride` ) is different
so we need to keep this catch here, for the lookups

whois-commons/src/main/java/net/ripe/db/whois/common/override/OverrideCredentialValidator.java
```diff
@@ -30,7 +30,12 @@ public boolean isAllowedAndValid(final boolean isTrusted, final UserSession user
             return false;
         }
         final OverrideCredential.OverrideValues overrideValues = overrideCredential.getOverrideValues().get();
-        return isAllowedToUseOverride(isTrusted, userSession, overrideValues.getUsername()) && isValidOverride(overrideValues, objectType);
+
+        try {
+            return isAllowedToUseOverride(isTrusted, userSession, overrideValues.getUsername()) && isValidOverride(overrideValues, objectType);
```
maggarwal13, line 35: Can you catch the exception in filterAuthFunction? As we do not catch exception for other public method and let the calling method handles it , it would be consistent

MiguelAHM: OK, Done

whois-commons/src/main/java/net/ripe/db/whois/common/rpsl/transform/FilterAuthFunction.java
```diff
@@ -119,8 +120,12 @@ public RpslObject apply(final RpslObject rpslObject) {
     }
 
     private boolean isOverrideAuthenticated(final ObjectType objectType){
-        return overrideCredentialValidator != null && overrideCredentialValidator.isAllowedAndValid(isTrusted,
-                userSession, overrideCredential, objectType);
+        try {
+            return overrideCredentialValidator != null && overrideCredentialValidator.isAllowedAndValid(isTrusted,
+                    userSession, overrideCredential, objectType);
+        } catch (EmptyResultDataAccessException e){
```
maggarwal13, line 126: Sorry I should have pointed put wth earlier. comment but can you  all exceptions? There can be incorrectrSizeesultExce[tion too

MiguelAHM: Updated

maggarwal13 (approved):

PR #1762, opened 2025-05-22
fix cross origin issues

maggarwal13:
* Only allow unauthenticated Cross-Origin (CORS) requests from anywhere using GET, HEAD, OPTIONS methods.
* If credentials are supplied then ignore  i.e send back unfiltered response but the request should still succeed.
* Do not allow cross-origin requests using POST, PUT, DELETE methods.


whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisCrossOriginFilter.java
```diff
@@ -0,0 +1,74 @@
...
+public class WhoisCrossOriginFilter extends CrossOriginFilter {
+
+    @Override
+    public void doFilter(final ServletRequest request, final ServletResponse response, final FilterChain chain) throws IOException, ServletException {
+
+        final HttpServletRequest httpRequest = (HttpServletRequest) request;
+
+        // GET request does not trigger a pre-flight request
+        if(httpRequest.getMethod().equals(HttpMethod.GET)
+                && httpRequest.getPathInfo().contains("syncupdates")
```
eshryane, line 28: Why hardcode "syncupdates" ?

maggarwal13: Because we allow updates using get only in case of syncupdates and the comment above mentiosn that since get does not trigger pre-flight request like POST and PUT , request in backend will still be successful but browser will show cross origin failed but till by then object has been updated using GET request

eshryane: OK, no CORS requests allowed to syncupdates API.

whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisCrossOriginFilter.java
```diff
@@ -0,0 +1,74 @@
...
+
+public class WhoisCrossOriginFilter extends CrossOriginFilter {
+
+    @Override
+    public void doFilter(final ServletRequest request, final ServletResponse response, final FilterChain chain) throws IOException, ServletException {
+
+        final HttpServletRequest httpRequest = (HttpServletRequest) request;
+
+        // GET request does not trigger a pre-flight request
+        if(httpRequest.getMethod().equals(HttpMethod.GET)
```
eshryane, line 27: move to a separate method, same as `addCORSHeaders`

whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisCrossOriginFilter.java
```diff
@@ -0,0 +1,74 @@
...
+            final HttpServletResponse httpResponse = (HttpServletResponse) response;
+            httpResponse.sendError(HttpStatus.UNAUTHORIZED_401, "Not Authorized");
+            return;
+        }
+
+        addCORSHeaders(httpRequest, (HttpServletResponse) response);
+        super.doFilter(new CrossOriginRequestWrapper(httpRequest), response, chain);
+    }
+
+    private void addCORSHeaders(final HttpServletRequest request, final HttpServletResponse response) {
```
eshryane, line 40: Make conditional on request also having an `Origin:` header? Otherwise not necessary.

eshryane (approved):

PR #1766, opened 2025-05-27
Cors add whitelist for hosts

maggarwal13:
Create a allow list of hosts for which CORS restrictions are a=ignored

### DIFF
--- a/whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisCrossOriginFilter.java
+++ b/whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisCrossOriginFilter.java
@@ -11,55 +11,67 @@ import jakarta.servlet.http.HttpServletResponse;
 import jakarta.ws.rs.HttpMethod;
 import jakarta.ws.rs.core.UriBuilder;
 import org.apache.commons.lang.StringUtils;
-import org.eclipse.jetty.http.HttpStatus;
 import org.eclipse.jetty.servlets.CrossOriginFilter;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
 
 import java.io.IOException;
+import java.util.Arrays;
 
+@Component
 public class WhoisCrossOriginFilter extends CrossOriginFilter {
+
+    final protected String[] allowedHostsforCrossOrigin;
+
+    @Autowired
+    public WhoisCrossOriginFilter(@Value("${whois.allow.cross.origin.hosts}") final String[] allowedHostsforCrossOrigin) {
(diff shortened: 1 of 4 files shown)


whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisCrossOriginFilter.java
```diff
@@ -69,7 +88,17 @@ public String getQueryString() {
...
+    }
+
+    private static String getHttpOrigin(HttpServletRequest request) {
+        return request.getHeader(HttpHeaders.ORIGIN);
+    }
+
+    private static boolean isHostsAllowedForCrossOrigin(final HttpServletRequest request, final String[] allowedHostsforCrossOrigin) {
+        if(!isOriginHeaderPresent(request)) return true;
+
+        return Arrays.stream(allowedHostsforCrossOrigin).anyMatch( host -> host.equalsIgnoreCase(getHttpOrigin(request)));
```
eshryane, line 102: So we will have to include the protocol in the `whois.allow.cross.origin.hosts` ? You don't strip it out of the Origin header

whois-api/src/test/java/net/ripe/db/whois/api/rest/WhoisCrossOriginRestServiceTestIntegration.java
```diff
@@ -271,14 +273,14 @@ public void cross_origin_get_request_from_ripe_net_is_allowed_no_sensitive_info_
...
-    public void cross_origin_post_request_from_apps_db_ripe_net_is_allowed_override_no_sensitive_information() {
+    public void cross_origin_get_request_from_apps_db_ripe_net_is_allowed_override_no_sensitive_information() {
         databaseHelper.insertUser(User.createWithPlainTextPassword("mherran", "zoh", ObjectType.MNTNER));
 
         final WhoisResources whoisResources = RestTest.target(getPort(), "whois/test/mntner/OWNER-MNT")
                 .queryParam("unfiltered", "")
                 .queryParam("override", encode("mherran,zoh,reason {notify=false}"))
                 .request(MediaType.APPLICATION_XML)
-                .header(com.google.common.net.HttpHeaders.ORIGIN, "https://apps.db.ripe.net")
+                .header(com.google.common.net.HttpHeaders.ORIGIN, "https://stats.ripe.net")
```
eshryane, line 283: Use `stat.ripe.net` as the example (i.e. the RIPEstat service)

whois-api/src/main/java/net/ripe/db/whois/api/rest/WhoisCrossOriginFilter.java
maggarwal13: that's a good idea. I will make that change

PR #1777, opened 2025-06-06
Remove HTTP Status code from RDAP response "title" element

eshryane:
Ref. RFC 9083, Section 6 "Error Response Body".


eshryane:
Integration tests passed

PR #1784, opened 2025-06-18
Don't log INFO stack trace if outgoing message is missing message-id

eshryane:


whois-api/src/main/java/net/ripe/db/whois/api/mail/dequeue/MessageDequeue.java
```diff
@@ -219,7 +219,10 @@ private void handleMessage(final String messageId) {
             }
 
         } catch (MailParsingException e){
-            LOGGER.info("Error detecting bounce detection or unsubscribing for messageId {}", messageId, e);
+            LOGGER.info("Error detecting bounce detection or unsubscribing for messageId {} due to {}: {} ({})",
```
MiguelAHM, line 222: why would you want to do this change?

MiguelAHM (approved):

Issue #1814, opened 2025-06-27
Inconsistent 3xEOL in --list-versions response


fvompe:
When whois protocol is used, 3xEOL is expected only between responses.
This doesn't work for old objects within --list-versions query, when this objects do not have version history, e.g.

Consider we have AS-ILOS as-set

```bash
$ echo '-r AS-ILOS' | nc whois.ripe.net 43
% This is the RIPE Database query service.
% The objects are in RPSL format.
%
% The RIPE Database is subject to Terms and Conditions.
% See https://docs.db.ripe.net/terms-conditions.html

% Note: this output has been filtered.
%       To receive output for a database update, use the "-B" flag.

% Information related to 'AS-ILOS'

as-set:         AS-ILOS
descr:          ILOS S.A. Luxembourg
members:        AS21179
admin-c:        HR9940-RIPE
tech-c:         HR9940-RIPE
mnt-by:         ILOS-MNT
created:        1970-01-01T00:00:00Z
last-modified:  2001-09-21T21:50:04Z
source:         RIPE

% This query was served by the RIPE Database Query Service version 1.117 (SHETLAND)


```

When --list-versions is used, we get 3xEOL in whois response :

```bash
$ echo '-k --list-versions AS-ILOS' | nc whois.ripe.net 43
% This is the RIPE Database query service.
% The objects are in RPSL format.
%
% The RIPE Database is subject to Terms and Conditions.
% See https://docs.db.ripe.net/terms-conditions.html

% Version history for AS-SET object "AS-ILOS"
% You can use "--show-version rev#" to get an exact version of the object.

rev#  Date              Op.


% This query was served by the RIPE Database Query Service version 1.117 (BUSA)


% This is the RIPE Database query service.
% The objects are in RPSL format.
%
% The RIPE Database is subject to Terms and Conditions.
% See https://docs.db.ripe.net/terms-conditions.html

^C
```

eshryane:
Hi @fvompe I agree this change is an improvement not to return 3 EOL's in a row.

The underlying cause is missing history, as presented at RIPE 84 (see slide 14) :
https://ripe84.ripe.net/wp-content/uploads/presentations/106-DB-WG-Operational-Update-RIPE84.pdf 

We have made progress to fix some missing history. The initial import on 21st September 2001 is now complete, including AS-ILOS. but we need to fix the index table containing all object versions.

PR #1815, opened 2025-06-27
fix(version-history): overcome 3xEOL in --list-versions response (#1814)

fvompe:
See #1814


eshryane:
Integration tests passed

PR #1817, opened 2025-06-27
Use local datasource in RDAP rirSearch1 requests

eshryane:
Moved to #1822


eshryane:
Tested in PREPDEV that this change causes the request lookups to be done on the local datasource.

whois-rdap/src/main/java/net/ripe/db/whois/rdap/RdapController.java
```diff
@@ -59,7 +59,7 @@ public class RdapController {
     private final String baseUrl;
     private final int maxResultSize;
     private final RdapRelationService rdapRelationService;
-
+    private final SourceContext sourceContext;
```
maggarwal13, line 62: Every Rdap query should go to slave datasource so I think we should fix it by changing the RdapRelationService to inject JdbcRpslObjectSlaveDao instaed of JdbcRpslObjectDao

PR #1848, opened 2025-08-07
Return clearer error in RDAP response when multiple objects match the entity primary key

MiguelAHM:
An RDAP entity primary key can match multiple objects, because different object types can match the same key.

If this happens, return clearer error.

This is a known issue:
https://github.com/RIPE-NCC/whois/blob/master/README.RDAP.md#entity-primary-key-can-match-multiple-objects

### DIFF
--- a/whois-rdap/src/main/java/net/ripe/db/whois/rdap/RdapLookupService.java
+++ b/whois-rdap/src/main/java/net/ripe/db/whois/rdap/RdapLookupService.java
@@ -2,7 +2,6 @@ package net.ripe.db.whois.rdap;
 
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Iterators;
 import jakarta.servlet.http.HttpServletRequest;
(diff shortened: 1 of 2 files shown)


whois-rdap/src/main/java/net/ripe/db/whois/rdap/RdapLookupService.java
```diff
@@ -219,8 +217,7 @@ private Object getRdapObject(final HttpServletRequest request, final Iterable<Rp
         RpslObject resultObject = rpslIterator.next();
 
         if (rpslIterator.hasNext()) {
-            throw new RdapException("Internal Error", "Unexpected result size: " + Iterators.size(rpslIterator),
```
eshryane, line 222: Any idea why the reported size was "1" and not the actual size ? Was it because the `rpslIterator` was already read once, and one remaining?

For example: https://rdap.db.ripe.net/entity/Nada
Should report 2 matches, not one.

eshryane (approved):

PR #1853, opened 2025-08-25
[Auth Service] Set Stopwatch to getOrgsContactsGroupsEmails

MiguelAHM:
This change is needed to know if the auth service gets stressed running multiple calls agains this endpoint


eshryane:
I suggest adding a description why you're making this change. Also is 100 ms a useful threshold (i.e. do you expect most calls to complete in less than that time?)

PR #1873, opened 2025-09-19
Auth Service: Implement LIRs Account call

MiguelAHM:
Related: #1878

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -103,7 +103,7 @@
         <shedlock.version>5.16.0</shedlock.version>
         <slf4j.version>2.0.17</slf4j.version>
         <solrj.version>1.4.1</solrj.version>
-        <spring.version>6.2.8</spring.version>
+        <spring.version>6.2.11</spring.version>
         <spock.version>2.3-groovy-4.0</spock.version>
         <stax-utils.version>20070216</stax-utils.version>
 


whois-commons/src/main/java/net/ripe/db/whois/common/sso/domain/MemberContactsResponse.java
```diff
@@ -32,12 +32,15 @@ public static class ContactDetails {
         @XmlElement(required = true)
         public String email;
         public String group;
+        public String membershipId;
```
maggarwal13, line 35: Can you map active and inactive too ? We do require it for filtering

whois-commons/src/main/java/net/ripe/db/whois/common/sso/AuthServiceClient.java
```diff
@@ -281,4 +282,33 @@ public MemberContactsResponse getOrgsContactsGroupsEmails(final Long membershipI
         }
     }
 
+    @Nullable
+    @Stopwatch(thresholdMs = 100L)
+    public MemberContactsResponse getLirsAccounts(final List<String> lirIds) {
```
maggarwal13, line 287: rename to getMemberDetailsForLirs?

whois-commons/src/main/java/net/ripe/db/whois/common/sso/AuthServiceClient.java
```diff
@@ -281,4 +282,33 @@ public MemberContactsResponse getOrgsContactsGroupsEmails(final Long membershipI
...
+        // Used by whois-internal. Do not remove
+        if (lirIds.isEmpty()){
+            return null;
+        }
+
+        final String membershipIds = String.join(",", lirIds);
+        try {
+            return client.target(restUrl)
+                    .path(ORGANISATION_MEMBERS_PATH)
```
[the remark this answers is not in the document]
MiguelAHM: I don't want to refactor Ticketia logic. 
Could this be done in a different PR?

whois-commons/src/main/java/net/ripe/db/whois/common/sso/AuthServiceClient.java
maggarwal13: Also do you think its better to return Map<Long, MemberContactsResponse> , As every call we need to filter out per memebershipId so why not return it from here only . Saves multiple times filtering it out

MiguelAHM: The map is not with the MemberCnntactResponse
There could be "batch_amount" memberships in the MemberContactResponse
The Map is with the ContactDetails there the relation is 1 to 1.

whois-commons/src/main/java/net/ripe/db/whois/common/sso/AuthServiceClient.java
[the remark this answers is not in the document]
MiguelAHM: yes, it is clear over huddle
I did the change

maggarwal13 (approved):

